PR #32, opened 2020-04-10
Ensure Nextstrain CLI installed

kairstenfay:
If the Nextstrain CLI is not installed, exit the `rebuild-staging`
script.


kairstenfay:
Rebased onto master.

PR #36, opened 2020-04-22
Replace blank division with country to prevent unexpected subsampling

emmahodcroft:
Adds a line to ingest so that if a sample has no `division`, the entry for `country` is copied over.

This is because in the pipeline we often filter by `division` per month per year. It turns out if `division` is missing, then `''` is counted as a division group. There are hundreds of these, so they do end up getting subsampled down by `''` - even though this isn't sensible. This produces unexpected output, as sequences seem to rather randomly go missing when they shouldn't.

The downside is that `country` isn't a division either, so we're likely to grab a few more sequences than we intend - however I feel this behaviour is less worrying & more transparent than sequences going missing. 

### Testing
I ran this locally. Seemed to work.

### Note
This is likely to cause a deluge of missing colours and lat-longs in the next run after this is pushed - due to new `division`s which are `country`s.... So should be aware of these. These should be fairly easy to fix (all should be present as `country` entries - so just copy over and change `country` to `division`!), but putting so nobody gets alarmed.

You're very welcome, lovely template.

### DIFF
--- a/bin/transform
+++ b/bin/transform
@@ -91,6 +91,9 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
     gisaid_data['region']      = geographic_data[0]
     gisaid_data['country']     = geographic_data[1]
     gisaid_data['division']    = geographic_data[2]
+
+    gisaid_data.loc[pd.isnull(gisaid_data['division']), 'division'] = gisaid_data[pd.isnull(gisaid_data['division'])]['country']
+
     gisaid_data['location']    = geographic_data[3]
 
     return gisaid_data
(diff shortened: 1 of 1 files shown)


emmahodcroft:
My apologies that I keep pushing to this branch and having to force to clean it up again.... sorry!!!!!

kairstenfay:
Thanks for the work here, @emmahodcroft! I rebased onto master and pushed a commit simplifying the pandas `.loc` expression a bit. Would you take a look at it and let me know if everything looks alright?

emmahodcroft:
Hi @kairstenfay - thanks so much! I am pretty new to the intricacies of pandas etc, so I appreciate your simplification! 

Luckily I thought I'd just give it one last quick check - and realised I had put the code in the wrong place! If it sets the division that early, then if the country is wrong (rare - but actually does happen!) then it'll set the division wrong there, before any fixes from `annotations.tsv` have been applied.

Anyway, I've moved the code to the after all fixes have been applied. 
The only last thing to note is that once this is pushed, it'll probably generate a bunch of 'missing division' errors for whoever does the next run. 

I'll message about it on the Slack so we can coordinate and not surprise anyone!

PR #38, opened 2020-05-01
Annotation errors

jameshadfield:
This PR fixes the root cause of a number of errors noticed by @lmoncla and @trvrb.

These errors arise in regional subsampled builds due to our handling of exposure history (which explains the stochasticity as they depend on which samples are included in the build). Specifically, if a genome had `country_exposure` defined in the annotation but not `region_exposure` then this `country_exposure` would appear as a country in the final auspice build. If this country was outside the desired focal region, the result would be a legend such as 👇  ("Israel" should be "Asia", and "Mexico" should be "North America" in this example which is a South American regional build). The fix is to define `region_exposure` for these samples.

image

---

This PR changes a number of things:

1. Duplicate entries are removed from the annotations TSV. There were a few ones with different values, most of which were obvious to choose between except for NSW05 which was annotated as Iran (me) but China (Emma). cc @emmahodcroft 

2. A helper script is added to check that if `country_exposure` is set then `region_exposure` is explicitly set. Similarly for `division_exposure` and `country_exposure`. This script is not run automatically but it would be a good idea in my option.

3. Annotations are updated for the samples identified by (2). 

### Testing
I've regenerated data locally and am running builds on AWS now to make sure this works as desired

### DIFF
--- a/bin/check-annotations
+++ b/bin/check-annotations
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+"""
+Check annotations for potential errors / missing data
+"""
+
+## Not part of this script, but you can check if a strain has duplica
+# 
+
+import pandas as pd
+import sys
+
+annotations = pd.read_csv("./source-data/annotations.tsv", header=None, sep='\t', comment="#")
+annotations.columns = ['strain', 'trait', 'value']
+
+try:
+  annotations = annotations.pivot(index="strain", columns="trait", values="value")
+except ValueError:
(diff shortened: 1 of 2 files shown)


jameshadfield:
AWS test builds look good 👍

emmahodcroft:
Thanks James, this is great! And sorry, I have no idea why I changed NSW05 to China. May have been a copying error while trying to do something else... apologies! Seems clear from GISAID that it's Iran.

Issue #33, opened 2020-04-14
Alert for 'Future Dates' When Pre-Processing (Ingesting)


emmahodcroft:
**Context**  
Currently full runs take 2 hours. If sequences aren't checked when they come in (by scrolling through the incredibly helpful messages that appear in Slack) then it's very easy to miss if dates are seriously incorrect - being from the *future*. 

However, these dates mess up the whole run, as they ruin the tree, and the only fix is to add to `exclude.txt` and re-run. It only takes missing a couple to waste half a day on runs! 

Often we can get quick feedback from submitters on the correct date (or, they can be corrected online), so we wouldn't necessarily want these auto-kicked out (I think?), but it would be great to be alerted to these at the same time as we are alerted to new sequences, so we can immediately add to `exclude.txt` and reach out to submitters.

**Description**   
When sequence come in, can they be checked against today's date. If they are > today's date ('from the future'), then can we get a big alert on Slack, giving a list of the 'futuristic sequences' so that we know immediately. Then we are able to take action!

**Examples**  & **Possible solution**   

I think a message like the ones we currently get would be great. Maybe another Icon so that they are really easy to spot - if a lot comes through sometimes the messages are easy to miss!

tsibley:
> Often we can get quick feedback from submitters on the correct date (or, they can be corrected online), so we wouldn't necessarily want these auto-kicked out (I think?), but it would be great to be alerted to these at the same time as we are alerted to new sequences, so we can immediately add to `exclude.txt` and reach out to submitters.

If ncov-ingest doesn't drop the sequences on a known bad date, then the build it starts is destined to be bad, which I think just makes more work for humans and computers alike. Someone either has to go cancel the known-to-be-bad build or remember to ignore it later when it completes. And if manually added to excludes.txt, then someone has to remember to remove it later.

If the sequences with bad dates _are_ dropped, then the build can still be used if desired. The alert will help folks track down corrections and update annotations, which will then launch a new build automatically. If the date is corrected externally, then the sequence stops being dropped automatically and no one has to go remove anything from exclude.txt.

I'm thinking that this pipeline could even generate a list of automatically excluded sequences, commented with the reason for each, and that file could then be used by the build in conjunction with the manually-curated exclude.txt.

emmahodcroft:
[...]

PR #43, opened 2020-05-03
transform: Send Slack message with future dates

kairstenfay:
Sometimes we receive data from GISAID with collection dates in the
future. This causes a big headache for the Nextstrain nCoV build
maintainers when left uncaught until the end of a multi hour long build.
    
Send a Slack message containing the GISAID EPI ISL ID and collection
date for every strain that has a collection date from the future (e.g.
a date > now). Emma specifically requested that this Slack message be
formatted differently from the others so that it stands out.

### Related issue(s)  
#33

### DIFF
--- a/bin/flag-metadata
+++ b/bin/flag-metadata
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+"""
+Creates a list of flagged sequences from the provided metadata TSV in the
+Nextstrain exclusions.txt convention. Prints the list to stdout.
+"""
+import logging
+import argparse
+import pandas as pd
+from pathlib import Path
+from datetime import date
+
+
+LOG = logging.getLogger(__name__)
+
+
+if __name__ == '__main__':
+    base = Path(__file__).resolve().parent.parent
+
+    parser = argparse.ArgumentParser(
+        description="Warn about future dates in a metadata TSV file.",
+        formatter_class=argparse.RawTextHelpFormatter
+    )
+    parser.add_argument("metadata",
+        nargs="?",
+        default=base / "data/gisaid/metadata.tsv",
+        help="Location of generated metadata tsv. Defaults to `data/gisaid/metadata.tsv`")
+    args = parser.parse_args()
+
+    metadata = pd.read_csv(args.metadata, sep="\t")
+    metadata.loc[metadata.date > str(date.today()), 'reason'] = '# Collection date in the future'
+    flagged_strains = metadata.loc[metadata.reason.notnull()]
+
+    print(flagged_strains[['strain', 'reason']].to_csv(header=False, index=False, sep='\t'))
--- a/bin/ingest-gisaid
+++ b/bin/ingest-gisaid
@@ -51,9 +51,12 @@ fi
   --output-metadata data/gisaid/metadata.tsv \
   --output-fasta data/gisaid/sequences.fasta
 
+./bin/flag-metadata data/gisaid/metadata.tsv > data/gisaid/flagged_metadata.txt
+
 if [[ "$branch" == master ]]; then
     ./bin/notify-on-metadata-change data/gisaid/metadata.tsv "$S3_SRC/metadata.tsv.gz" gisaid_epi_isl
     ./bin/notify-on-additional-info-change data/gisaid/additional_info.tsv "$S3_SRC/additional_info.tsv.gz"
+    ./bin/notify-on-flagged-metadata-change data/gisaid/flagged_metadata.txt "$S3_SRC/flagged_metadata.txt.gz"
 
     ./bin/rebuild-staging data/gisaid/metadata.tsv "$S3_SRC/metadata.tsv.gz" \
       data/gisaid/sequences.fasta "$S3_SRC/sequences.fasta.gz"
@@ -61,4 +64,5 @@ fi
 
 ./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/metadata.tsv "$S3_DST/metadata.tsv.gz"
 ./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/additional_info.tsv "$S3_DST/additional_info.tsv.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/flagged_metadata.txt "$S3_DST/flagged_metadata.txt.gz"
 ./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/sequences.fasta "$S3_DST/sequences.fasta.gz"
--- a/bin/notify-on-flagged-metadata-change
+++ b/bin/notify-on-flagged-metadata-change
@@ -1,0 +1,31 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${SLACK_TOKEN:?The SLACK_TOKEN environment variable is required.}"
+
+bin="$(dirname "$0")"
+
+src="${1:?A source flagged metadata txt file is required as the first argument.}"
+dst="${2:?A destination flagged metadata txt s3:// URL is required as the second argument.}"
+
+dst_local="$(mktemp -t flagged-metadata-XXXXXX.txt)"
+
+diff="$(mktemp -t flagged-metadata-additions-XXXXXX)"
+
+trap "rm -f '$dst_local' '$diff'" EXIT
+
+aws s3 cp --no-progress "$dst" - | gunzip -cfq > "$dst_local"
+
+comm -13 \
+    <(sort "$dst_local") \
+    <(sort "$src") \
+    > "$diff"
+
+if [[ -s "$diff" ]]; then
+    echo
+    echo "Notifying Slack about flagged metadata additions."
+    "$bin"/notify-slack ":waving_black_flag: Newly flagged metadata"
+    "$bin"/notify-slack --upload "flagged-metadata-additions.txt" < "$diff"
+else
+    echo "No flagged metadata additions."
+fi


kairstenfay:
Here's an example of the current format of this "future dates" message (using fake dates I manipulated):
image

kairstenfay:
[ncov-ingest: warn about future dates](https://trello.com/c/iYoOt9fe/565-ncov-ingest-warn-about-future-dates)

emmahodcroft:
I haven't tested (not even sure I can!) but from a quick look over the code, this looks great @kairstenfay , thank you so much!! This will be a *huge* help to the runs!

bin/transform
```diff
@@ -245,6 +253,60 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        for every GISAID EPI ISL ID and future date, respectively.
+        """
+        for datetime_format in ('%Y-%m-%d', '%Y-%m', '%Y'):
+            try:
+                date = datetime.strptime(row.date, datetime_format)
+            except ValueError:
+                continue
+
+            if date > datetime.now():
+                future_dates[row.gisaid_epi_isl] = row.date
```
tsibley, line 274: I think this whole bit could make use of the observation that ISO-8601 dates (even truncated ones) are string-comparable and skip the whole `strptime` step by directly comparing the current date to `row.date`. For example, consider the following:

```
In [1]: from datetime import date

In [2]: today = str(date.today())

In [3]: today
Out[3]: '2020-05-04'

In [4]: '2020-08' > today
Out[4]: True

In [5]: '2020-05' > today
Out[5]: False

In [6]: '2020-05-30' > today
Out[6]: True

In [7]: '2020-04' > today
Out[7]: False

In [8]: '2020' > today
Out[8]: False
```

NAs might have to be dropped first in the actual code.

bin/transform
```diff
@@ -245,6 +253,60 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+
+            return
+
+    future_dates = {}
+    gisaid_data.apply(future_date, axis=1)
```
tsibley, line 279: I try to avoid functions that successively accumulate values in an external variable, as I think it makes the flow a bit more convoluted. In general, I think it's clearer to either a) use a for loop or b) use a functional chain of list comprehensions or maps and filters. I tend to prefer (b) as it more clearly signals that the expression is taking a list and producing a transformed list, but I think both are more clear than mixing functional (`apply`) and imperative (modifying an external variable) styles. This specific case is ~inconsequential and fine as-is, but it's a good example of the pattern so I wanted to mention it for you to think on.

bin/transform
```diff
@@ -293,6 +355,12 @@ if __name__ == '__main__':
     gisaid_data = parse_sex(gisaid_data)
     gisaid_data = update_metadata(gisaid_data)
 
+    # Warn about future dates
+    try:
+        warn_future_dates(gisaid_data)
+    except HTTPError:
+        LOG.info("Slack message couldn't be sent, but continuing anyway.")
```
tsibley, line 362: The whole `warn_future_dates` function feels like it should be a separate script that runs after this one. Particularly since it doesn't affect output at all, is used for side-effects only, and is ignored if it fails.

bin/transform
kairstenfay: That's really clever! I'll incorporate that in a new commit.

bin/transform
kairstenfay: That makes sense to me.

bin/transform
kairstenfay: Yes, I thought this code was a bit smelly, too, but couldn't think of how else to build this dict. However, in my new approach, I discard the dict entirely.

kairstenfay:
I made the requested changes by Tom and also incorporated feedback from the conversation in #33. 

This PR now runs an `exclude` script after `transform` and uses the existing `notify-slack` script to upload files to Slack. Exclusions are automatically saved to a new S3 file called `exclude.txt` that will be used by `ncov` along with the manually created `exclude.txt`.

bin/exclude
```diff
@@ -0,0 +1,38 @@
...
+"""
+import os
+import json
+import logging
+import requests
+import argparse
+import pandas as pd
+from pathlib import Path
+from datetime import date
+from urllib3.exceptions import HTTPError
```
tsibley, line 14: Several of these imports aren't used.

bin/notify-on-exclude-change
```diff
@@ -0,0 +1,31 @@
...
+trap "rm -f '$dst_local' '$diff'" EXIT
+
+aws s3 cp --no-progress "$dst" - | gunzip -cfq > "$dst_local"
+
+comm -13 \
+    <(sort "$dst_local") \
+    <(sort "$src") \
+    > "$diff"
+
+if [[ -n "$(< "$diff")" ]]; then
```
tsibley, line 24: A more idiomatic test that doesn't require reading in the file contents is:

```sh
if [[ -s "$diff" ]]; then
```

see <https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html>.

kairstenfay:
> Looks good to me. A couple minor things to clean up.
> 
> We'll need to sort out the fate of [nextstrain/ncov#401](https://github.com/nextstrain/ncov/pull/401) before merging this, right?

@tsibley Yes. I think the simplest resolution is to make pulling `exclude.txt` from S3 optional.

bin/notify-on-exclude-change
kairstenfay: Thanks. I also changed a similar conditional in `notify-on-metadata-change` in the latest commit.

bin/exclude
kairstenfay: Ugh. Setting up python linting in VS Code is not very straightforward. I'm going to start using `pylint` at the command line.

bin/notify-on-exclude-change
tsibley: Ah, in this case, the use of `-n` in `notify-on-metadata-change` was intentional, see https://github.com/nextstrain/ncov-ingest/commit/2e7093fb58894bb028b96450bc1e1ac87ce6d0af. I should have left a comment in the program as well.

kairstenfay: Gotcha. I documented that in a comment in the fixed up version of the latest commit

kairstenfay:
> General comment: it's nice to have the notifications for the future dates, but I wonder if it's necessary to generate an additional `exclude.txt` file. Can we not use the `augur filter --max-date` option to automatically filter out future dates in the ncov build?

Great question, @joverlee521. It sounds like that would be a better solution than creating a new `exclude.txt` file for future dates only (I say "future dates only" because there may be more reasons down the road we'd want to automatically exclude sequences). 

I don't know the answer to your question but I asked it in #33.

emmahodcroft:
I like this idea, if there's a way we can use today's date in the call? (I am sure this is possible, but unsure the most robust way to do this). This would be nice as it would mean as dates were updated, they'd be auto-included without having to update any files (always a plus). 

Would still be good to get Slack notifications of the future dates though, so we can flag them to submitters!

kairstenfay:
> I like this idea, if there's a way we can use today's date in the call? (I am sure this is possible, but unsure the most robust way to do this). This would be nice as it would mean as dates were updated, they'd be auto-included without having to update any files (always a plus).
> 
> Would still be good to get Slack notifications of the future dates though, so we can flag them to submitters!

I pushed a commit to nextstrain/ncov#401 that uses `augur filter --max-date` to exclude future dates.  Instead of creating a new script called `exclude` here, I wonder how we can semanitcally reflect that we're not going to store sequences with future dates in an automatically generated _exclude.txt_ for augur exclusions, but we'll still track them for Slack notifications.

emmahodcroft:
I wonder if we should just not call it `exclude.txt` but instead `future_dates.txt` or similar. This shows exactly what we're tracking, and can help us to alert submitters, but doesn't imply that it going to go into the exclude file used by filter. (As they won't need to, with the max-date option.) Or, if we think we may want to keep track of a larger variety of files, we could call it something like `flagged_sequences.txt` (though we'll want to note this is different from `ncov/results/flagged_sequences.tsv` which stores the results of the sequence diagnostics...

kairstenfay:
> I wonder if we should just not call it `exclude.txt` but instead `future_dates.txt` or similar. This shows exactly what we're tracking, and can help us to alert submitters, but doesn't imply that it going to go into the exclude file used by filter. (As they won't need to, with the max-date option.) Or, if we think we may want to keep track of a larger variety of files, we could call it something like `flagged_sequences.txt` (though we'll want to note this is different from `ncov/results/flagged_sequences.tsv` which stores the results of the sequence diagnostics...

I like where you're going with `flagged_sequences.txt`. Naming things is hard and I agree we don't want to confuse this file with `ncov/results/flagged_sequences.tsv`. I'll keep thinking on this.

kairstenfay:
> I like where you're going with `flagged_sequences.txt`. Naming things is hard and I agree we don't want to confuse this file with `ncov/results/flagged_sequences.tsv`. I'll keep thinking on this.

What are your thoughts on `flagged_metadata.txt`? This file would contain any and all sequences that have problematic metadata and be sent to slack. These sequences won't be dropped in `ingest` but possibly filtered out downstream (like in `augur filter --max-date xx-xx-xxxx`). 

One thing to keep in mind is this will eventually live alongside `exclude`. In a new PR, I will re-create the `exclude` script and the `exclude.txt` file for automatic exclusions (see #45 for an example).

emmahodcroft:
> What are your thoughts on flagged_metadata.txt? This file would contain any and all sequences that have problematic metadata and be sent to slack. These sequences won't be dropped in ingest but possibly filtered out downstream (like in augur filter --max-date xx-xx-xxxx).

I think that's a good name, but it'll be a bit confusing if it also contains sequences which will be excluded due to sequence attributes (like length, in the example) you linked to. I think the other thing is we'll want to ensure this file only contains what's 'new' if it's going to the slack... getting a long list of everything ever won't be helpful in the Slack! (Too much info!) (Though might be useful to store somewhere.)  

However if we're using this mostly for metadata stuff, like dates, I think the name is good! Perhaps the 'latest additions' can go to the slack, and all can be included in a file with reasons for why it's 'bad'. Hope I'm understanding!

kairstenfay:
> I think that's a good name, but it'll be a bit confusing if it also contains sequences which will be excluded due to sequence attributes (like length, in the example) you linked to. 

It won't, ideally. The example I linked to would put sequences in an `exclude.txt` file like this PR originally proposed.  

> I think the other thing is we'll want to ensure this file only contains what's 'new' if it's going to the slack... getting a long list of everything ever won't be helpful in the Slack! (Too much info!) (Though might be useful to store somewhere.)

I see two solutions here. 
1. We could append all "flagged metadata" to one file and send only the diff via Slack (like what we do for _metadata.tsv_).
2. We could overwrite a "flagged metadata" file every ingest that contains only bad metadata from new or changed rows. 

I think 1 may be simpler, but I definitely agree we don't want to clutter Slack unnecessarily. 


> However if we're using this mostly for metadata stuff, like dates, I think the name is good! Perhaps the 'latest additions' can go to the slack, and all can be included in a file with reasons for why it's 'bad'. Hope I'm understanding!

Sounds good, thanks. I would like to format this file in the same style as `exclude.txt` where each sequences has an inline comment describing the problem.

emmahodcroft:
> We could append all "flagged metadata" to one file and send only the diff via Slack (like what we do for metadata.tsv).

I think this would be great, if it's not too much trouble to do! If sequences were updated (to have a correct date), would they disappear from this file? 

> Sounds good, thanks. I would like to format this file in the same style as exclude.txt where each sequences has an inline comment describing the problem.
Sounds perfect!

kairstenfay:
> I think this would be great, if it's not too much trouble to do! If sequences were updated (to have a correct date), would they disappear from this file?

Not in the way I described, but this is even better and could certainly be arranged.

kairstenfay:
This PR now adds strains with future dates to a file called _flagged_metadata.txt_ that is in the same style as the `ncov` _exclude.txt_. It sends Slack notifications for any new exclusions not in the old file stored on S3 and overwrites that file at the end of each ingest.

tsibley:
I rebased this onto master, squashing commits in the process to reduce code churn, and re-pushed. If the workflows run without error, I'll merge.

PR #46, opened 2020-05-13
Fetch all GenBank sequences + metadata from NCBI Virus

tsibley:
This adds a new file, genbank.ndjson.gz, which currently contains about
~3,600 sequences.  The previous genbank.fasta.gz file will no longer be
updated; it was a small subset of just 62 sequences which we'd manually
linked between GISAID and GenBank.  58 of those sequences are in the new
file; the remainder were removed from GenBank in the interim.

Note that the GenBank data is still not ingested, just fetched and
tracked.  The next steps are to incorporate it into the "transform" step
such that it ends up mixed in with the GISAID data in metadata.tsv and
sequences.fasta by:

  1. Aligning the fields and values between GISAID and GenBank

  2. ~Matching sequences in both repositories, likely by strain name
     heuristics, and then deduplicating matches.  Emma has already started
     scoping out the feasibility of this.~ Nevermind! It looks like we're leaning
     towards separate builds here for the two data sources.

### DIFF
--- a/bin/fetch-from-genbank
+++ b/bin/fetch-from-genbank
@@ -1,40 +1,72 @@
 #!/usr/bin/env python3
-import pandas as pd
-from Bio import Entrez
-from os import environ
-from shutil import copyfileobj
-from sys import argv, exit, stdin, stdout, stderr
+"""
+Download all SARS-CoV-2 sequences and their curated metadata from GenBank via
+NCBI Virus.
 
-Entrez.email = environ.get("NCBI_EMAIL", "hello@nextstrain.org")
+Outputs newline-delimited JSON records.
 
-if not argv[1:]:
-    print(f"Reading metadata.tsv from stdin", file = stderr)
-    metadata_file = stdin
+The request this program makes is based on observing the network activity that
 
-elif len(argv[1:]) == 1:
-    print(f"Reading metadata.tsv from {argv[1]!r}", file = stderr)
-    metadata_file = argv[1]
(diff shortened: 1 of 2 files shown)


kairstenfay:
I typically see python scripts wrapped in a `if __name__ == '__main__':` block. Is there a reason here why you didn't do it that way?

joverlee521:
Just FYI, I'm going to start working on the transform script based on the genbank.ndjson produced.

bin/fetch-from-genbank
```diff
@@ -1,40 +1,72 @@
...
+            ('location',                'CountryFull_s'),
+            ('collected',               'CollectionDate_s'),
+            ('submitted',               'CreateDate_dt'),
+            ('length',                  'SLen_i'),
+            ('host',                    'Host_s'),
+            ('isolation_source',        'Isolation_csv'),
+            ('biosample_accession',     'BioSample_s'),
+            ('authors',                 'Authors_csv'),
+            ('publications',            'PubMed_csv'),
+            ('sequence',                'Nucleotide_seq'),
```
joverlee521, line 56: I found some additional fields that would be useful while playing around with the API on Postman:
```
('region', 'Region_s`),
('strain', 'Isolate_s')
```

tsibley: Awesome! :raised_hands: I've added these in another commit.

tsibley:
> I typically see python scripts wrapped in a `if __name__ == '__main__':` block. Is there a reason here why you didn't do it that way?

That conditional is primarily useful if the code expects to be (or wants to support being) run both as a script (the program entrypoint) _and_ imported as a module. I don't expect to need that here, so didn't add those extra bits.

PR #48, opened 2020-05-15
fetch-from-genbank: Add contact information to our NCBI requests

tsibley:
The NCBI's Entrez API requires an "email" parameter¹ and while this
isn't Entrez, include it anyway to be nice.  For good measure, also
include contact information in the User-Agent header.  Both of these can
help the NCBI reach out if our fetches are causing issues, rather than
having no clue where they're coming from.

¹ https://www.ncbi.nlm.nih.gov/home/about/policies/

### DIFF
--- a/bin/fetch-from-genbank
+++ b/bin/fetch-from-genbank
@@ -62,9 +62,17 @@ params = {
     # Stable sort with newest last so diffs work nicely.  Columns are source
     # data fields, not our output columns.
     'sort': 'SourceDB_s desc, CollectionDate_s asc, id asc',
+
+    # This isn't Entrez, but include the same email parameter it requires just
+    # to be nice.
+    'email': 'hello@nextstrain.org',
 }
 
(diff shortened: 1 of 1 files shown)


bin/fetch-from-genbank
```diff
@@ -62,9 +62,17 @@ params = {
     # Stable sort with newest last so diffs work nicely.  Columns are source
     # data fields, not our output columns.
     'sort': 'SourceDB_s desc, CollectionDate_s asc, id asc',
+
+    # This isn't Entrez, but include the same email parameter it requires just
+    # to be nice.
```
kairstenfay, line 67: Why is this is a nice thing to do? In case someone else wants to run this code but against Entrez?

tsibley: Nice to NCBI, who may notice an `email` parameter in their request logs when they're digging into burdensome requests.

PR #50, opened 2020-05-18
Use EPI_ISL number to add annotation information

emmahodcroft:
Currently we use `strain` to match up the `annotations.tsv` file. Unfortunately, if `strain` changes (and it often does) this means our annotations fall out of date. We try to keep up with this manually in runs, but it's really easy to miss some name changes.

This *should* make all this a thing of the past, as EPI_ISL numbers are much more stable than `strain`. We'll still need to keep an eye out, as they have been known to change (mostly in error), but in general this should make runs easier!

However, to aid readability I kept the `strain` column in there - it usually makes things clearer as it contains the country and often the division or other info - just makes it much easier to tell at a glance if changes make sense.

### Testing
Using the newly added `source-data/annotations-EPI_ISL.tsv` you can test the build to see how you think it's running, by passing this in instead of the current `annotations.tsv` file.

Before merging, a new version of this file (with EPI_ISL numbers) should be generated to ensure it's up-to-date with any updates to `annotations.tsv` that happened since this PR. 

@tsibley  & @kairstenfay if you are able to give this a quick look I'd appreciate it! If you approve, I can generate a final 'new' version of the file, and merge all of it together (with the EPI_ISL version of the file replacing `annotations.tsv`).


emmahodcroft:
(The `ingest` is failing on the check for this because it's using the 'old' annotations file... they'll have to both change in sync!)

kairstenfay:
@emmahodcroft I just realized we'll also need to update the documentation in `./bin/transform` for the [`annotations` argument](https://github.com/nextstrain/ncov-ingest/blob/9efff443834a862f4f2775cd3536635189dee72c/bin/transform#L261)

emmahodcroft:
Ah good catch Kairsten! I'm not going to work more on this tonight, but I can do this tomorrow morning before I merge. Will it cause problems if I just rename the new annotations file to be `annotations.tsv` and delete the one in order to do this 'switch'...? Or will that mess up Github?

kairstenfay:
> Ah good catch @kairstenfay ! I'm not going to work more on this tonight, but I can do this tomorrow morning before I merge. Will it cause problems if I just rename the new annotations file to be `annotations.tsv` and delete the one in order to do this 'switch'...? Or will that mess up Github?

Replacing the existing `annotations.tsv` with your new one should work fine as long as your new annotations file contains all the latest data from `master`.

PR #52, opened 2020-05-18
transform: strip spaces before dropping prefix in strain name

joverlee521:
Currently there are four genomes with `hCoV-19` in the strain name in
`metadata.tsv` because the original strain name had leading white spaces.

Strip all spaces before dropping the `hCoV-19` prefix to fix this.

### DIFF
--- a/bin/transform
+++ b/bin/transform
@@ -63,7 +63,7 @@ def preprocess(gisaid_data: pd.DataFrame) -> pd.DataFrame:
     # consistent, predictable string comparisons.
     for column in gisaid_data:
         if gisaid_data[column].dtype == "string":
-            gisaid_data[column] = gisaid_data[column].str.normalize("NFC")
+            gisaid_data[column] = gisaid_data[column].str.normalize("NFC").str.strip()
 
     # Calculate sequence length.  GISAID provides a "sequence_length" column,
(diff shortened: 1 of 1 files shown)


kairstenfay:
It seems like there's string stripping happening all over the place in `./bin/transform`. Maybe it would be better to strip white space early on in `./bin/transform` for every column with a `dtype == 'string'` (building on [Thomas' recent commit](https://github.com/nextstrain/ncov-ingest/commit/0bec632a596cb8a16ea7b3fab5e2ab1fe75ef52b)).

kairstenfay:
Doing so would handle the `.strip()` for strain name, originating lab, submitting lab, and authors.

joverlee521:
> It seems like there's string stripping happening all over the place in `./bin/transform`. Maybe it would be better to strip white space early on in `./bin/transform` for every column with a `dtype == 'string'` (building on [Thomas' recent commit](https://github.com/nextstrain/ncov-ingest/commit/0bec632a596cb8a16ea7b3fab5e2ab1fe75ef52b)).

I agree. I didn't know if there was some reason `strip()` wasn't centralized.

tsibley:
> I agree. I didn't know if there was some reason `strip()` wasn't centralized.

No good reason.

PR #53, opened 2020-05-20
fix: Fill null division before division_exposure

kairstenfay:
Interpolate values for a null `division` before interpolating null
values for `division_exposure`, because, if missing, the latter uses the
value of the former.

### DIFF
--- a/bin/transform
+++ b/bin/transform
@@ -222,6 +222,10 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
                 value = value.rstrip()
             curated_gisaid_data.loc[curated_gisaid_data['gisaid_epi_isl'] == epi_isl, key] = value
 
+    # if division is blank, replace with country data, to avoid unexpected effects when subsampling by division
+    # (where an empty division is counted as a 'division' group)
+    curated_gisaid_data.loc[pd.isnull(curated_gisaid_data['division']), 'division'] = curated_gisaid_data['country']
+
     # Set `region_exposure` equal to `region` if it wasn't added by annotations
     if 'region_exposure' in curated_gisaid_data:
         curated_gisaid_data['region_exposure'].fillna(curated_gisaid_data['region'], inplace=True)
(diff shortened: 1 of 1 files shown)


joverlee521:
I think @emmahodcroft moved this to the end on purpose? (referring to this [commit](https://github.com/nextstrain/ncov-ingest/commit/6f059c198d07587ce0fdcce27765ebaf2d7e4fa4))

kairstenfay:
> I think @emmahodcroft moved this to the end on purpose? (referring to this [commit](https://github.com/nextstrain/ncov-ingest/commit/6f059c198d07587ce0fdcce27765ebaf2d7e4fa4))

Oh good find, @joverlee521. 

@emmahodcroft this code is causing null division_exposure even though division gets filled in. I believe this is the culprit behind the error you messaged me about in #ncov-gisaid-updates.

emmahodcroft:
Ah ok. The problem I found before was that if you fill an empty `division` up top, then if the `division` is changed due to `annotations.tsv` then the old `division_exposure` was remaining without being changed to match! 

So then we had `division_exposure` not being updated to match `division` after change in the `annotations.tsv`, making it look erroneously as if we had some information about the exposure.

I guess the best path would probably be to fix this at the top, but then ensure that if any `division` `country` or `region` is modified from the `annotations.tsv` that this is applied also to `division_exposure`, `country_exposure` `region_exposure`.... except if there is also an entry for those things in `annotations.tsv`....
I hope that makes sense!

kairstenfay:
> Ah ok. The problem I found before was that if you fill an empty `division` up top, then if the `division` is changed due to `annotations.tsv` then the old `division_exposure` was remaining without being changed to match!
> 
> So then we had `division_exposure` not being updated to match `division` after change in the `annotations.tsv`, making it look erroneously as if we had some information about the exposure.
> 
> I guess the best path would probably be to fix this at the top, but then ensure that if any `division` `country` or `region` is modified from the `annotations.tsv` that this is applied also to `division_exposure`, `country_exposure` `region_exposure`.... except if there is also an entry for those things in `annotations.tsv`....
> I hope that makes sense!

It does make sense @emmahodcroft. This PR fills in missing `division`s after applying all annotations.

emmahodcroft:
Great! In that case it looks good on paper to me, if it seems to do the right thing in testing, that's good by me!

Issue #47, opened 2020-05-14
Improve error handling in `check-annotations`


kairstenfay:
@lmoncla and I currently ran into an error raised by `check-annotations` with the following output:
```
The following strains have a division_exposure but not a country_exposure.
This is ok if the country_exposure happens to match the country, but problems arise if this is not the case.
To make this clear, country_exposure should be explicitly set for the following:
strain
France/CIBU200107/2020    NaN
Name: strain, dtype: object
strains have a division_exposure but not a country_exposure
```

France/CIBU200107/2020 was correctly annotated for sampling and exposure locations. It had `division` and `division_exposure` set but not `country` or `country_exposure` annotated. However, in the metadata, the country and country exposure were France.

This issue was resolved by removing `division_exposure` annotations for France/CIBU200107/2020. If it's the case where we don't want to have a `division` and `division_exposure` annotation without a `country` or `country_exposure` annotation, then we should describe that the `division_exposure` can be dropped in the error message sent to the user.

kairstenfay:
Another thought: would it be possible, without creating a circular dependency, to read from the metadata to determine if an annotated strain has an appropriately filled `country`, `country_exposure`, `region`, or `region_exposure`? Another issue popped up with additional sequences annotated with country exposures but not region exposures, because the region is the same in the metadata. Having to clarify in `annotations.tsv` that the region for these hasn't changed ends up creating additional work for the Nextstrain team.

kairstenfay:
One more thought: the `check-annotations` script currently exits after the first error is encountered. Consider collecting all errors that arise from the `annotations.tsv` and providing all of them to the user at the end of runtime.

jstoja:
I've just seen that the check caused issues on master. Sorry to read this.
I'll take a look at your findings here and make sure that the PR correctly pass CI before merging.

jstoja:
> If it's the case where we don't want to have a division and division_exposure annotation without a country or country_exposure annotation, then we should describe that the division_exposure can be dropped in the error message sent to the user.

@jameshadfield Would you be able to confirm this? You mention auspice build errors in #39 

> would it be possible, without creating a circular dependency, to read from the metadata to determine if an annotated strain has an appropriately filled country, country_exposure, region, or region_exposure?

The annotations are used to generate the metadata with `bin/transform`. https://github.com/nextstrain/ncov-ingest/blob/master/bin/transform#L217
Therefore, I'm not sure which one should override the other one.

[...]

Issue #39, opened 2020-05-01
Run `./bin/check-annotations` as CI (or similar)


jameshadfield:
#38 introduced `./bin/check-annotations` which helps identify problems with the contents of `annotations.tsv`. This highlights issues that may only become evident stochastically after running the auspice build. 

It would be nice to run this as a CI (or similar). For this to happen the script would have to be modified to exit appropriately which it doesn't currently do.

emmahodcroft:
This would be really good. Additionally - just realised another check we should do - if a sequence name listed in `annotations.tsv` is not found in the metadata. 

This can indicate a sequence name has changed, and thus the file needs to be updated. I'll see if I can add this if I have time. 
Would clearly be most valuable if this is run automatically as James suggests above!

jstoja:
Hey @emmahodcroft I just saw your suggestion about sequences checking.
I'm just wondering:
* Should it check the sequences against from [`ncov/data/metadata.tsv`](https://github.com/nextstrain/ncov/blob/master/data/metadata.tsv) or [`ncov-ingest/data/metadata.tsv` (this repository)]() ?
* Since I don't have the `metadata.tsv` file from this repository, can I base myself from [`ncov/data/metadata.tsv`](https://github.com/nextstrain/ncov/blob/master/data/metadata.tsv) to implement this additional check in #40 ?

Thanks

kairstenfay:
> Hey @emmahodcroft I just saw your suggestion about sequences checking.
> I'm just wondering:
> 
>     * Should it check the sequences against from [`ncov/data/metadata.tsv`](https://github.com/nextstrain/ncov/blob/master/data/metadata.tsv) or `ncov-ingest/data/metadata.tsv` (this repository) ?
> 
>     * Since I don't have the `metadata.tsv` file from this repository, can I base myself from [`ncov/data/metadata.tsv`](https://github.com/nextstrain/ncov/blob/master/data/metadata.tsv) to implement this additional check in #40 ?
> 
> 
> Thanks

Hi @jstoja. Please run the checks against the `metadata.tsv` file in this repo (`ncov-ingest`). You can base yourself from another copy of the `metadata.tsv` (e.g. one from the `ncov` repo) for development, because both files have the same structure.

emmahodcroft:
[...]

PR #54, opened 2020-05-20
Location hierarchy

kairstenfay:
Validates incoming metadata, paving the way for rolling `./bin/check-annotations` into `./bin/transform`.
Validates metadata region, country, and division against a defined set of known hierarchies. This set is made via `./bin/create-hierarchy`. 
If new (previously unseen) hierarchies are found, outputs them to a local file, `data/new_location_hierarchies.tsv` and sends a Slack notification alerting ncov build maintainers about them.

### Related issue(s)  
Creates list of hierarchies as described by #47
Collects all failures instead of exiting on first error as described by #47
Will eventually pull in additional checks and replace `check-annotations` as described in #39 

~### Remaining tasks~
~- [x] invoke `./bin/create-hierarchy` in `ingest`~
~- [x] allow annotations to explicitly permit unknown hierarchies in metadata~
~- [x] write _draft_metadata.tsv_ before annotations?~
~- [ ] incorporate other functionality in `./bin/check_annotations` or save for a future PR?~ save for a future PR 

### Questions
~- Do we want a hard `ingest` fail when we encounter new location hierarchies or emit a warning instead?~ No, we decided to send a slack notification 
~- should _location_hierarchy.tsv_ live on S3 or under _./source-data/_?~ I put it on S3 so that it stays up to date


bin/ingest
```diff
@@ -47,10 +47,15 @@ if [[ "$branch" == master ]]; then
 fi
 ./bin/upload-to-s3 --quiet data/gisaid.ndjson "$S3_DST/gisaid.ndjson.gz"
 
+if [[ -f "$S3_SRC/metadata.tsv.gz" ]]; then
+    ./bin/create-hierarchy "$S3_SRC/metadata.tsv.gz" > source-data/location_hierarchy.tsv
+fi
```
kairstenfay, line 52: We may want to just assume `metadata.tsv.gz` exists here

lib/utils/hierarchy_dataframe.py
```diff
@@ -21,6 +21,7 @@ def hierarchy_dataframe(metadata: pd.DataFrame, hierarchy_columns: list) -> pd.D
             sep='_',
             suffix='\w+') \
         .reset_index()[hierarchy_columns] \
+        .fillna('') \
```
kairstenfay, line 24: I am now questioning if this is desirable behavior or if we'd prefer no side effects of hidden type casting.

bin/transform
```diff
@@ -257,6 +257,33 @@ def validate_metadata(metadata: pd.DataFrame,
...
+        # Avoid a KeyError if the annotations file doesn't contain all of
+        # our resolution columns
+        columns = [ col for col in annotations.columns if col in keep_columns ]
+
+        wide_annotations = annotations \
+            .pivot(index='gisaid_epi_isl', columns='variable', values='value')
+        wide_annotations = wide_annotations[columns].dropna()
+
+        wide_annotations['gisaid_epi_isl'] = wide_annotations.index
+        wide_annotations = wide_annotations.reset_index(drop=True)
```
kairstenfay, line 283: This is kind of clunky IMO. It's some data prep that allows `hierarchy_dataframe()` to run using the expeted `gisaid_epi_isl` column, but it makes me wonder if it's better if `hierarchy_dataframe()` relies on an index instead of a specific, named column for its reshaping.

kairstenfay:
[ncov-ingest: Location hierarchy to validate metadata](https://trello.com/c/QCQhsO2A/588-ncov-ingest-location-hierarchy-to-validate-metadata)

bin/transform
```diff
@@ -247,6 +251,91 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        """
+        assert metadata[location_hierarchy_columns].notnull().all().all(), \
+            "There are missing values in the location hierarchy columns " \
+            f"«{location_hierarchy_columns}». At this point, we expect all " \
+            "missing values to be interpolated."
+
+    def no_unknown_hierarchies(known_location_hierarchies: pd.DataFrame):
+        """
+        Raises an ``AssertionError`` if the metadata contain hierarchies not
+        found in the defined location hierarchy.
```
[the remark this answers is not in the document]
kairstenfay: Thanks, good catch.

bin/transform
```diff
@@ -247,6 +251,91 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        """
+        metadata_hierarchies = hierarchy_dataframe(metadata, location_hierarchy_columns)
+
+        all_hierarchies = metadata_hierarchies.merge(
+            known_location_hierarchies,
+            how='outer',
+            indicator=True
+        )
+
+        unknown_hierarchies = all_hierarchies[all_hierarchies._merge == 'left_only']
```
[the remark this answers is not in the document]
kairstenfay: Yep! It comes from the `indicator` argument in Pandas. It is new to me as well.

bin/transform
```diff
@@ -247,6 +251,91 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+
+    try:
+        no_missing_hierarchy_data()
+    except AssertionError as e:
+        exceptions.append(e)
+
+    try:
+        no_unknown_hierarchies(known_location_hierarchies)
+    except AssertionError as e:
+        exceptions.append(e)
```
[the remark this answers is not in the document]
kairstenfay: Not for this PR but I plan to port more validations from `check-annotations` where it will be needed.

lib/utils/hierarchy_dataframe.py
```diff
@@ -0,0 +1,27 @@
...
+    'region_exposure').
+    """
+    metadata = metadata.rename(columns={
+        resolution: f'{resolution}_strain' for resolution in hierarchy_columns
+    })
+
+    return pd \
+        .wide_to_long(metadata,
+            stubnames=hierarchy_columns,
+            i=['gisaid_epi_isl'],
```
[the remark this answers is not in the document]
kairstenfay: Ah good idea. I can update the function to accept an ID argument.

kairstenfay:
> After going through this PR, I think I might be a little out of the loop of how the flow is supposed to work here. From what I've seen, we are comparing the latest location hierarchy with the previously saved location hierarchy, then writing out a the new hierarchies to Slack. Are we then correcting invalid hierarchies using annotations?

You've nearly got it. We're comparing the latest location hierarchy with metadata originating from the previous run of `transform` and out the new hierarchies to Slack _after_ applying all annotations in _annotations.tsv_.

bin/transform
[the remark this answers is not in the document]
kairstenfay: Updated in a new commit.

lib/utils/hierarchy_dataframe.py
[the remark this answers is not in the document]
kairstenfay: Updated in a new commit.

kairstenfay:
I wonder if the metadata validations should live in a separate script, e.g. one called `validate`.

bin/create-hierarchy
```diff
@@ -0,0 +1,34 @@
...
+        "can be used to validate future metadata and annotations. The output "
+        "hierarchy TSV is printed to stdout, and you likely want to redirect "
+        "it to a new file.",
+    formatter_class=argparse.RawTextHelpFormatter
+)
+
+parser.add_argument(
+    "metadata",
+    nargs="?",
+    default="s3://nextstrain-ncov-private/metadata.tsv.gz",
```
tsibley, line 26: This path will want to be updated during/after the rebase.

bin/create-hierarchy
```diff
@@ -0,0 +1,34 @@
...
+    help="Metadata file (in TSV format) used to build hierarchy.\n")
+
+args = parser.parse_args()
+
+metadata = pd.read_csv(args.metadata, sep='\t')
+
+known_hierarchies = hierarchy_dataframe(metadata, 'gisaid_epi_isl', HIERARCHY_COLUMNS)
```
tsibley, line 33: The `gisaid_epi_isl` key may want to be parameterized.

bin/transform
```diff
@@ -294,7 +395,15 @@ if __name__ == '__main__':
     gisaid_data = parse_authors(gisaid_data)
     gisaid_data = parse_age(gisaid_data)
     gisaid_data = parse_sex(gisaid_data)
-    gisaid_data = update_metadata(gisaid_data)
+
+    # Write intermediate metadata to file for comparison before annotations
+    gisaid_data.to_csv(base / 'data/draft_metadata.tsv', sep='\t', index=False)
```
tsibley, line 400: This file path seems like it should be based on the `args.output_metadata` path rather than hardcoded.

bin/transform
```diff
@@ -247,6 +251,95 @@ def update_metadata(curated_gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        unknown_hierarchies = all_hierarchies[all_hierarchies._merge == 'left_only']
+
+        if not unknown_hierarchies.empty:
+            unknown_hierarchies[location_hierarchy_columns] \
+                .to_csv(args.output_new_location_hierarchies, sep='\t', index=False)
+
+    # Combine metadata location hierarchies with those coded in annotations
+    known_location_hierarchies = pd.concat([
+        pd.read_csv(args.location_hierarchy, sep='\t'),
+        annotation_hierarchies(annotations)]
```
tsibley, line 323: I don't understand why this combines the location hierarchy file with the annotations file in order to produce known locations.

Separately, reading the annotations file as a table is not quite kosher and might lead to subtly incorrect results. This is because earlier lines in the annotations file can be overridden by later lines. It is essentially a procedural list of changes to make. To make it function more like an overlay table, we'd want to enforce uniqueness on (record id, field id).

joverlee521:
Can we borrow the [geo-synonyms.tsv](https://github.com/nextstrain/fauna/blob/master/source-data/geo_synonyms.tsv) from fauna to do location hierarchy validation? 

Also, I feel like `location-hierarchy.tsv` should be a curated list that lives within `source-data` of this repo rather than hidden in the private S3 bucket. When we see new hierarchies, they can be validated and added to the list or if they are incorrect, update the record using annotations. (This is the flow I use in maintaining the seasonal flu builds)

bin/transform
kairstenfay: I think I was thinking in too-constrained terms when implementing the location hierarchy validation. The desire to also check the annotations file comes from the check I wanted to replace in _./bin/check-annotations_ that every country annotation also has a division annotation to match, etc.

bin/transform
kairstenfay: Yes, that makes sense.

bin/create-hierarchy
kairstenfay: I believe it is parametrized?

bin/create-hierarchy
kairstenfay: Thank Tom for your feedback. I'll try to address it in comments, but I'm happy to jump on a call with you to further discuss the intended functionality here.

> For example, why does `ingest` run `create-hierarchy` to produce a local version of `location_hierarchy.tsv` from the _existing_ `s3://nextstrain-ncov-private/metadata.tsv.gz` but `transform` reads from `s3://nextstrain-ncov-private/location_hierarchy.tsv.gz`? And then the local version of `location_hierarchy.tsv` is uploaded to S3. It seems like at no point is an updated `location_hierarchy.tsv` made from the _new_ metadata.

You're right, this seems like an oversight on my end. I believe this branch got large enough that I started to lose sight of the bigger picture here. It seems like we should create a new location hierarchy at the end of ingest using new metadata and upload it to S3.

> I'm also confused about how `draft_metadata.tsv` is intended to be used. It is written but never uploaded anywhere…? (When we talked about a draft metadata table vs. the final metadata table before, the idea was that the the draft metadata makes it easier to compare the effect of annotations because the draft metadata + annotations = final metadata.)

You're right, it's not used anywhere here, just written. I thought it would be useful for nCoV build maintainers but wasn't entirely clear on how it could be useful other than through its pure existence. Do you have any recommendations on how to make it more useful? 

> It also strikes me as a bit strange to bundle up production of the `new_location_hierarchies.tsv` into `transform` rather than as a separate diff program that takes the old and new metadata files. I'd think an old vs. new diff is what we want, but maybe I'm missing the point of comparing against `location_hierarchy.tsv` instead?

The intent is to capture any new resolution combinations. Are you thinking an old vs new diff of unique rows of the resolution (e.g. region, country, division) columns? 

> If you could rebase this onto the latest version of master, that'd also be appreciated.

I will certainly do that.

kairstenfay:
Thanks, Tom, for your feedback. I'll try to address it in comments, but I'm happy to jump on a call with you to further discuss the intended functionality here.

> For example, why does `ingest` run `create-hierarchy` to produce a local version of `location_hierarchy.tsv` from the _existing_ `s3://nextstrain-ncov-private/metadata.tsv.gz` but `transform` reads from `s3://nextstrain-ncov-private/location_hierarchy.tsv.gz`? And then the local version of `location_hierarchy.tsv` is uploaded to S3. It seems like at no point is an updated `location_hierarchy.tsv` made from the _new_ metadata.

You're right, this seems like an oversight on my end. I believe this branch got large enough that I started to lose sight of the bigger picture here. It seems like we should create a new location hierarchy at the end of ingest using new metadata and upload it to S3.

> I'm also confused about how `draft_metadata.tsv` is intended to be used. It is written but never uploaded anywhere…? (When we talked about a draft metadata table vs. the final metadata table before, the idea was that the the draft metadata makes it easier to compare the effect of annotations because the draft metadata + annotations = final metadata.)

You're right, it's not used anywhere here, just written. I thought it would be useful for nCoV build maintainers but wasn't entirely clear on how it could be useful other than through its pure existence. Do you have any recommendations on how to make it more useful? 

> It also strikes me as a bit strange to bundle up production of the `new_location_hierarchies.tsv` into `transform` rather than as a separate diff program that takes the old and new metadata files. I'd think an old vs. new diff is what we want, but maybe I'm missing the point of comparing against `location_hierarchy.tsv` instead?

The intent is to capture any new resolution combinations. Are you thinking an old vs new diff of unique rows of the resolution (e.g. region, country, division) columns? 

> If you could rebase this onto the latest version of master, that'd also be appreciated.

I will certainly do that.

kairstenfay:
> Can we borrow the [geo-synonyms.tsv](https://github.com/nextstrain/fauna/blob/master/source-data/geo_synonyms.tsv) from fauna to do location hierarchy validation?

I was not aware of that file, but I can take a look at it and let you know how I see it fitting (or not fitting) into this workflow. 

> Also, I feel like `location-hierarchy.tsv` should be a curated list that lives within `source-data` of this repo rather than hidden in the private S3 bucket. When we see new hierarchies, they can be validated and added to the list or if they are incorrect, update the record using annotations. (This is the flow I use in maintaining the seasonal flu builds)

Yes, I had some trouble deciding where to put the _location_hierarchy.tsv_ file and opted for putting it on S3 such that `transform` would still run even if the user had not run `create-hiearchy` locally.

bin/backfill-location-hierarchy
```diff
@@ -0,0 +1,75 @@
...
+    parser.add_argument("metadata",
+        default="s3://nextstrain-ncov-private/metadata.tsv.gz",
+        nargs="?",
+        help="A GISAID metadata TSV file. \nDefaults to "
+            "s3://nextstrain-ncov-private/metadata.tsv.gz.")
+    parser.add_argument("--output",
+        default=base / "source-data/location_hierarchy.tsv",
+        nargs="?",
+        help="Output location for the new location hierarchy TSV file.\n"
+            "Defaults to ../source-data/location_hierarchy.tsv.")
```
kairstenfay, line 62: I'm not sure what's an appropriate default value to put in the help string here. Should it be relative to this file? The top of the repo? Should I use string substitution and replace it with a path on the user's computer?

lib/utils/hierarchy_dataframe.py
```diff
@@ -0,0 +1,39 @@
...
+def hierarchy_dataframe(metadata: pd.DataFrame, unique_id: str,
+    hierarchy_columns: List[str]) -> pd.DataFrame:
+    """
+    Given a ``pd.DataFrame`` `metadata`, returns a `pd.DataFrame` with one row
+    for every unique location hierarchy defined by the `unique_id` and
+    `hierarchy_columns` (e.g. region, country, and division). Looks for location
+    columns matching the original resolution name (e.g. 'region') and any
+    stubname of it (e.g. 'region_exposure').
+
+    Raises a :class:``KeyError`` if the given *unique_id* isn't in the
```
kairstenfay, line 15: I am definitely not getting the dostring documentation style right here. Which standard are we using for this repo?

bin/create-hierarchy
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
joverlee521, line 1: Is this script replaced by `check-locations`? I don't see this being used anywhere.

kairstenfay:
> Thank you for reworking this PR! I've left some comments but I also have two major suggestions:
> 
>     1. After looking at the current `source-data/location_hierarchy.tsv`, I think it might not be worth incorporating the fauna data.

I think proportionally much more of the fauna data will go unused than used if we do incorporate it, given the non-titlecasing, lack of spaces, and lack of accents in fauna resolution names. So, I'm ok with discarding it. 

>     2. I think we should include `location` in the hierarchy as well. It would be helpful to the ncov build maintainers to get alerts of new `locations` since they need to add them to the `lat_longs.tsv` file in the ncov repo.

I have been wondering for some time if `location` should be added to the location hierarchy columns. I'm happy to add it in.

bin/check-locations
```diff
@@ -0,0 +1,63 @@
...
+base = Path(__file__).resolve().parent.parent
+
+
+def validate_metadata(metadata: pd.DataFrame, columns: List[str]):
+    def no_missing_hierarchy_data():
+        """
+        Raises an ``AssertionError`` if the metadata have missing values in the
+        given location hierarchy *columns*.
+        """
+        assert metadata[columns].notnull().all().all(), \
```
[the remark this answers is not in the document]
kairstenfay: Yes

bin/create-hierarchy
kairstenfay: Oops, you're right. it's no longer being used anywhere.

bin/check-locations
```diff
@@ -0,0 +1,63 @@
...
+            "There are missing values in the location hierarchy columns " \
+            f"«{columns}». At this point, we expect all " \
+            "missing values to be interpolated."
+
+    no_missing_hierarchy_data()
```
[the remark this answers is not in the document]
kairstenfay: So the inner function here is a bit of overkill because it's the only validation being run. But, I separated the `no_missing_hierarchy_data()` logic into its own function because future PRs will add more functionality to `validate_metadata()`. 

I defined the function inside of `validate_metadata()` because the larger scope of this script doesn't need to know about `no_missing_hierarchy_data()`. Really, defining it as an inner function is really just a personal preference. When looking at the `validate_metadata()` function, I find it easier to read all the validations that will be run within the definition of `validate_metadata()` than to go looking for the function definitions elsewhere in the script.

bin/check-locations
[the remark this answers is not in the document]
kairstenfay: Side note: it looks like we don't interpolate missing `location`, so I will exclude that column from this missing data check.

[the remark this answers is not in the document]
joverlee521: Yup I believe `location` is the only column in the hierarchy that can be missing.

bin/notify-on-location-hierarchy-addition
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${SLACK_TOKEN:?The SLACK_TOKEN environment variable is required.}"
```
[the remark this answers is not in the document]
kairstenfay: Alternatively, `SLACK_TOKEN` could be removed from all the `notify-*` scripts (except `notify-slack`, which checks for both `SLACK_TOKEN` and `SLACK_CHANNLES`.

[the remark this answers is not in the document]
joverlee521: Hmm, I want to say that we've required `SLACK_TOKEN` in all of the `notify-*` scripts so that they can fail loudly early on instead of waiting for it to fail at the `notify-slack` command.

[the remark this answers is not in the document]
kairstenfay: Hmm but then we'd do some data processing just to throw it away for a missing env var... you're right, I'll instead add the check for `SLACK_CHANNELS` to all `notify-*` scripts.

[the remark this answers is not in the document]
kairstenfay: Ha we had nearly the same thought at the same time. I wrote my follow up response before seeing your comment. :see_no_evil:

kairstenfay:
I guess I won't add `location` until an agreement is made in [this Slack thread](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1593649049230500).

kairstenfay:
It looks like Emma is preferring keeping all locations, so I'll go ahead and add those changes now, possibly reverting them later.

emmahodcroft:
Sorry I maybe should have been clearer - I think keeping locations for the USA (which are usually counties) is important, but I'm open to letting other locations go if we're feeling this is too much time. I don't think we should delete/remove locations we've already found though!

But I think by the sound of it your script could still flag them, we just decide if we put them in orderings/lat-longs or not? Could ignore if we want?

I am going to try and implement a better way to source lat-longs though - so maybe we could bring them back if I can do that!

kairstenfay:
> Sorry I maybe should have been clearer - I think keeping locations for the USA (which are usually counties) is important, but I'm open to letting other locations go if we're feeling this is too much time. I don't think we should delete/remove locations we've already found though!
> 
> But I think by the sound of it your script could still flag them, we just decide if we put them in orderings/lat-longs or not? Could ignore if we want?
> 
> I am going to try and implement a better way to source lat-longs though - so maybe we could bring them back if I can do that!

No problem! Thanks for providing additional context. As this PR is written, new locations could be ignored, but the list will grow very long until a manual intervention is made at the new `./source-data/location_hierarchy.tsv`. I understand that the discussion on lat/long and locations is still in flux. I'm trying to design this to be as flexible as possible, and so I'm completely open to adapting the process to whatever the ultimate decision is (which is hopefully whatever requires the least additional work for you all). Keeping or discarding locations or programmatically keeping locations based on a country should all be feasible.

Issue #19, opened 2020-03-09
Allow data from outside GISAID


jameshadfield:
We're often getting genomes + metadata shared prior to GISAID release. It would be good to allow this to be incorporated into this repo for consumption by the ncov pipeline. Not sure of the best way to address this -- we could commit a "non-GISAID" FASTA + TSV, but that may cause the repo to grow too big over time.

kairstenfay:
@joverlee521 and @tsibley are currently working on importing data from GenBank which may close this issue.

kairstenfay:
Closing this issue since we've now split our workflow to process data from GISAID or Genbank. The refactoring done in #55 should allow other, external sources of data to be ingested as well.

PR #55, opened 2020-05-21
Transform GenBank

joverlee521:
Parse the GenBank JSON load into a metadata TSV and a FASTA file.

### Related issue(s)  

Related to #19 

### Remaining tasks
- [x] notify on GenBank change (Should this be a different Slack channel?)
- [x] notify on metadata change
- [x] notify on problem records that need manual annotations (this is currently records that don't have `region`, `country` or a valid `strain`)


bin/transform-genbank
```diff
@@ -0,0 +1,317 @@
...
+
+
+# Note: 'sequence' should NEVER appear in this list!
+METADATA_COLUMNS = [  # Ordering of columns in the existing metadata.tsv in the ncov repo
+    'strain', 'virus', 'gisaid_epi_isl', 'genbank_accession', 'date', 'region',
+    'country', 'division', 'location', 'region_exposure', 'country_exposure',
+    'division_exposure', 'segment', 'length', 'host', 'age', 'sex',
+    'originating_lab', 'submitting_lab', 'authors', 'url', 'title',
+    'date_submitted'
+]
```
kairstenfay, line 20: I think we'd ideally import this list from a shared file for genbank and gisaid transformations.

bin/transform-genbank
```diff
@@ -0,0 +1,317 @@
...
+    if 'country_exposure' in genbank_data:
+        genbank_data['country_exposure'].fillna(genbank_data['country'], inplace=True)
+    else:
+        genbank_data['country_exposure'] = genbank_data['country']
+
+    # Set `division_exposure` equal to `division` if it wasn't added by annotations
+    if 'division_exposure' in genbank_data:
+        genbank_data['division_exposure'].fillna(genbank_data['division'], inplace=True)
+    else:
+        genbank_data['division_exposure'] = genbank_data['division']
```
kairstenfay, line 239: Same comment re: redundancy in the first half of this function and `transform-gisaid`'s `update_metadata()` as well

bin/transform-genbank
```diff
@@ -0,0 +1,317 @@
...
+
+    return genbank_data
+
+
+def write_fasta_file(sequence_data: pd.DataFrame):
+    """ """
+    with fsspec.open(str(args.output_fasta), 'wt') as fastafile:
+        for index, row in sequence_data.iterrows():
+            fastafile.write(f">{row['strain']}\n")
+            fastafile.write(f"{row['sequence']}\n")
```
kairstenfay, line 268: This is identical to `transform-gisaid`'s `write_fasta_file()`

joverlee521:
>  I have some concerns regarding duplication of code between this and the gisaid transformations.

A very legitimate concern! I duplicated the code because I'm not sure how these two transform scripts will diverge in the future. I'd rather keep them completely separate from the start than detangle them later on when things drift apart (Although that might just be me overthinking things again 😄)

kairstenfay:
> > I have some concerns regarding duplication of code between this and the gisaid transformations.
> 
> A very legitimate concern! I duplicated the code because I'm not sure how these two transform scripts will diverge in the future. I'd rather keep them completely separate from the start than detangle them later on when things drift apart (Although that might just be me overthinking things again smile)

That's a valid point. At the very least it may be worth refactoring out objects/functions that are identical across both scripts. I started a [small example of importing shared functions](https://github.com/nextstrain/ncov-ingest/tree/location-hierarchy/bin/utils) on a WIP branch.

bin/transform-genbank
```diff
@@ -0,0 +1,317 @@
...
+    Returns a DataFrame with a column for GenBank accession plus
+    additional columns containing hardcoded metadata.
+    """
+    hardcoded_metadata = pd.DataFrame(genbank_data['genbank_accession'])
+    hardcoded_metadata['virus']             = 'ncov'
+    hardcoded_metadata['gisaid_epi_isl']    = '?'
+    hardcoded_metadata['segment']           = 'genome'
+    hardcoded_metadata['age']               = '?'
+    hardcoded_metadata['sex']               = '?'
+    hardcoded_metadata['originating_lab']   = '?'
```
joverlee521, line 193: age, sex, and originating lab seems to be available for some records in BioSample. However, the `Entrez.efetch` API does not work with the BioSample accession that we get from the GenBank pull. The only way I've been able to get this to work is to use `Entrez.esearch` to search the BioSample accession to get the BioSample ID, then use that BioSample ID to get the metadata via `efetch`. The caveat here is that I can't seem to use `esearch` to do a search for a batch of accessions, so we would need to make a separate `esearch` request for each accession, then do a batch `efetch` with all the returned BioSample IDs.

joverlee521: Another discovery, `Entrez.read()` does not work with BioSample's XML (see: https://github.com/biopython/biopython/issues/2817). So we would have use a python XML parser to parse the data returned by `efetch`.

tsibley: I believe you can use `esearch` to find a batch of accessions by constructing a query string like this:

    SAMPLE1[accn] or SAMPLE2[accn] or …

The raw link for example: <https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=biosample&term=SAMEA2467098[accn]+or+SAMN07663113[accn]&retmode=json>

A long list of accessions will hit URL-length limits, so best to use POST params instead of GET + query params. I don't know if `Entrez.esearch` uses GET or POST by default or if it's dynamic depending on the data. It's also possible to use the [server-side search history feature](https://www.ncbi.nlm.nih.gov/books/NBK25497/#chapter2.Using_the_Entrez_History_Server) of Entrez to store the results of your esearch there and then tell efetch to use that saved id list, if you want to avoid round-tripping them.

joverlee521: Thanks for digging into this! Looking at the [source code](https://biopython.org/DIST/docs/api/Bio.Entrez-pysrc.html) it seems that it dynamically uses POST if the query is over 1000 characters.

tsibley: Ahh, you're correct. I missed that!

bin/transform-genbank
```diff
@@ -0,0 +1,282 @@
...
+        (r'^USA-', 'USA/'),
+        (r'^USACT-', 'USA/CT-'),
+        (r'^USAWA-', 'USA/WA-'),
+        (r'^HKG.', 'HongKong/'),
+    ]
+    # Add country codes to regex_replacement
+    country_codes = pd.read_csv(base / 'source-data/country-codes.tsv', header=None, sep='\t', comment="#")
+    country_codes.apply(lambda c: regex_replacement.extend([
+        (rf'^{c[0]}/', f'{c[1]}/'),
+        (rf'/{c[0]}/', f'/{c[1]}/')]), axis=1)
```
tsibley, line 54: To be safe, it'd be good to call `re.escape` on the values used to construct these regexes.

If I'm understanding correctly, I think the two patterns appended to `regex_replacement` could also be reduced to `(rf'(^|/){c[0]}/', rf'\1{c[1]}/')`. I wonder if this makes the intent more clear or less clear.

bin/utils/transform.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+import fsspec
+import pandas as pd
+
```
tsibley, line 4: Great to have a shared utils library, but it's a bit unconventional to have things other than programs in a bin directory. I would put this file into a lib directory, e.g. `lib/utils/transform.py` and then make sure `lib` is on `sys.path`.

source-data/country-codes.tsv
```diff
@@ -0,0 +1,248 @@
+# ISO 3166-1 alpha-3 codes
```
tsibley, line 1: The names here are non-standard and it seems like there are some missing codes too. Is there a reason to mangle the names?

For comparison, this is a standard list as pulled from Wikipedia (with two small name tweaks for disambiguation/geopolitical reasons): https://gist.github.com/tsibley/c5916b7d738e9826f8cdc9669a9d2543

source-data/us-state-codes.tsv
```diff
@@ -0,0 +1,58 @@
+# from https://gist.github.com/rogerallen/1583593
```
tsibley, line 1: Some of the names here don't match the official names from the US Census. Any reason not the use the official names like we do in ID3C? https://github.com/seattleflu/id3c-customizations/blob/master/locations/data/states.txt

source-data/country-codes.tsv
joverlee521: Mangled some to fit what I've seen in the GISAID data. If we don't care about matching the strain names between the two sources, then we can use the standard.

source-data/us-state-codes.tsv
joverlee521: ditto mangling to fit GISAID data.

bin/utils/transform.py
joverlee521: ☝️ @kairstenfay (for reference on your wip branch)

bin/transform-genbank
joverlee521: Yes, it can be reduced to `(rf'(^|/){c[0]}/', rf'\1{c[1]}/')` (although I had to look up what `\1` was doing here). Regex also always looks like gibberish to me so not sure which way is more clear here.

source-data/country-codes.tsv
tsibley: I think we landed on "don't care about matching", and we can always mangle the names as part of any future matching process itself.

I'd definitely prefer to start from a standard point and mangle, rather than try to unmangle later.

bin/utils/transform.py
tsibley: We discussed this on our calls, but wanted to record here for posterity: The (`bin/augur` dev wrapper](https://github.com/nextstrain/augur/blob/42a26edc229c32cd83e4bf272ce0bdada1c100c6/bin/augur) includes an example of import path manipulation. It's a bit verbose, and if desired, could be shortened into something like:

```python
import sys
from pathlib import Path
sys.path.insert(0, str(Path(__file__).parent.parent / "lib"))

from utils.transform import (…)
```

bin/transform-genbank
tsibley: I'd say stick with what you had then. :-)

joverlee521:
[Ingest sequences from GenBank in ncov-ingest](https://trello.com/c/jaY9wpvL/580-ingest-sequences-from-genbank-in-ncov-ingest)

.github/workflows/ingest-genbank.yml
```diff
@@ -0,0 +1,33 @@
+name: Ingest 2019-nCov/SARS-CoV-2 data from GenBank for nextstrain.org/ncov
+
+on:
+  push:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # "At minute 0 past every 4th hour"
+    - cron:  '0 */4 * * *'
```
tsibley, line 8: This should be `*/6` to reflect the latest changes on master (and is what ingest-gisaid.yml uses because of the rebase).

.github/workflows/ingest-genbank.yml
```diff
@@ -0,0 +1,33 @@
...
+        python3 -m pip install --upgrade pip setuptools
+        python3 -m pip install pipenv
+        pipenv sync
+        pipenv run ./bin/ingest-genbank
+      env:
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        GISAID_API_ENDPOINT: ${{ secrets.GISAID_API_ENDPOINT }}
+        GISAID_USERNAME_AND_PASSWORD: ${{ secrets.GISAID_USERNAME_AND_PASSWORD }}
```
tsibley, line 31: These should be removed for GenBank.

bin/transform-genbank
```diff
@@ -0,0 +1,291 @@
...
+    country_codes = pd.read_csv(base / 'source-data/country-codes.tsv', header=None, sep='\t', comment="#")
+    # Remove accents and non-alphabet characters from country names
+    country_codes[1] = country_codes[1] \
+        .str.normalize("NFD") \
+        .str.encode("ascii", errors="ignore") \
+        .str.decode("utf-8") \
+        .str.replace(r'[^a-zA-Z]', '')
+    country_codes.apply(lambda c: regex_replacement.extend([
+        (rf'^{re.escape(c[0])}/', f'{c[1]}/'),
+        (rf'/{re.escape(c[0])}/', f'/{c[1]}/')]), axis=1)
```
tsibley, line 63: I looked at this again because the forcing things into ASCII caught my eye. In general, things shouldn't be forced into ASCII. But then when re-reading the surrounding context, I'm not really sure why we'd replace ISO country codes with mangled names in these strain names? Country codes seem very appropriate for use in strain names, so I think we could just leave them as-is.

bin/ingest-gisaid
```diff
@@ -43,25 +43,22 @@ set -x
...
 
-./bin/upload-to-s3 ${silent:+--quiet} data/metadata.tsv "$S3_DST/metadata.tsv.gz"
-./bin/upload-to-s3 ${silent:+--quiet} data/additional_info.tsv "$S3_DST/additional_info.tsv.gz"
-./bin/upload-to-s3 ${silent:+--quiet} data/sequences.fasta "$S3_DST/sequences.fasta.gz"
-
-./bin/fetch-from-genbank > data/genbank.ndjson
-./bin/upload-to-s3 ${silent:+--quiet} data/genbank.ndjson "$S3_DST/genbank.ndjson.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/metadata.tsv "$S3_DST/gisaid_metadata.tsv.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/additional_info.tsv "$S3_DST/gisaid_additional_info.tsv.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/sequences.fasta "$S3_DST/gisaid_sequences.fasta.gz"
```
tsibley, line 64: I believe these path changes on S3 will mean communicating changes to GISAID and letting them adjust, slowing down deploying this PR. I'd suggest leaving the GISAID files on S3 named as-is for now? This should be ~ok, as we're not putting the GenBank files in this same bucket anymore.

bin/ingest-genbank
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${S3_SRC:=s3://nextstrain-data}"
```
tsibley, line 4: I think this wants to include an object prefix so things are a bit more clear. I'd suggest `s3://nextstrain-data/ncov` which then with the full paths below makes objects like `s3://nextstrain-data/ncov/genbank.ndjson.gz`.

bin/transform-genbank
joverlee521: Hmm, I think it was weird that some strain names contain the full country name while others are just country codes. If it's not weird to have a mixture of both, then I'm fine with leaving the country codes as is.

bin/ingest-gisaid
joverlee521: Oh right, I forgot that GISAID also has access to these files. I'll leave the files on S3 as is.

bin/transform-genbank
tsibley: Nod. I think fine to leave them. Strain names as submitted will continually use non-standard nomenclature in various ways. If we did want to standardize geolocality in strain names, I think using the ISO codes would be more appropriate than country names anyway.

joverlee521:
I should have noticed this sooner, but there's some inconsistency in the format of the location field.

Working based on the US locations, I thought the location field was formatted to be `country: location, division` but this is not the case. According to NCBI (from SPHERES [slack](https://cspheres.slack.com/archives/C0112BJ6W4F/p1591990641283800?thread_ts=1591987903.282100&cid=C0112BJ6W4F)), the only controlled part is `country:` but everything in the sub-regions is free text. They recommend sub-regions to go from largest -> smallest, but it is not required. 

Looks like submissions from outside of the US follows the `country: division, location` format while submissions from the US mostly follows the `country: location, division` format. I can work with this assumption in the parsing of the location field, but I was wondering if there is a better way to tackle this?

bin/transform-genbank
```diff
@@ -0,0 +1,280 @@
...
+        "MN908947	collection_date 2019-12-26 # Manually corrected date")
+    parser.add_argument("--output-metadata",
+        default=base / "data/genbank/metadata.tsv",
+        help="Output location of generated metadata tsv. Defaults to `data/genbank/metadata.tsv`")
+    parser.add_argument("--output-fasta",
+        default=base / "data/genbank/sequences.fasta",
+        help="Output location of generated FASTA file. Defaults to `data/genbank/sequences.fasta`")
+    parser.add_argument("--problem-data",
+        default=base / "data/genbank/problem_data.tsv",
+        help="Output location of generated tsv of problem records missing geography region or a valid strain name")
```
kairstenfay, line 264: I wonder if instead of naming this file _problem_data.tsv_, we could create an _exclude.txt_ and concatenate it with the one downstream in `ncov` like #43 originally proposed.

joverlee521: This is a little different from `exclude.txt`. The `problem_data.tsv` contains records of sequences that are not included in the `metadata.tsv` at all because they are missing required location fields or they have improperly formatted strain names.  So they are not even passed downstream to the ncov build at all.

bin/transform-genbank
```diff
@@ -0,0 +1,288 @@
...
+        else:
+            assert False, f"Found unknown format for geographic data: {value}"
+
+        # Special parsing for US locations because the format varies
+        if value[0] == 'USA' and division:
+            # Switch location & division if location is a US state
+            if location and any(location.strip() in s for s in us_states.items()):
+                state = location
+                location = division
+                division = state
```
tsibley, line 119: FYI, this is a nice idiom for swapping values:

```python
division, location = location, division
```

bin/transform-genbank
```diff
@@ -0,0 +1,288 @@
...
+            # Convert US state codes to full names
+            if us_states.get(division.strip()):
+                division = us_states[division.strip()]
```
tsibley, line 122: Might be more robust to normalize (internal space collapsing, lowercase, etc) `location` and `division` and/or `us_states` before comparing them in the `if … any(location … in us_states…)` and `if us_states.get(…)` lines.

PR #58, opened 2020-05-31
notify scripts should just exit cleanly if the destination key doesn't exist

ttung:
Right now, the notify scripts will error out if the destination key doesn't exist.  That means the ingest script cannot be run until it has succeeded in uploading data, which is a bootstrapping obstacle.

### Testing
Deleted the metadata key, then ran the script and it started the batch run.

### DIFF
--- a/bin/notify-on-additional-info-change
+++ b/bin/notify-on-additional-info-change
@@ -8,6 +8,12 @@ bin="$(dirname "$0")"
 src="${1:?A source additional info TSV file is required as the first argument.}"
 dst="${2:?A destination additional info TSV s3:// URL is required as the second argument.}"
 
+# if the file is not already present, just exit
+s3path="${2#s3://}"
+bucket="${dst_s3path%%/*}"
+key="${dst_s3path#*/}"
+aws s3api head-object --bucket $bucket --key $key || exit 0
+
 # Remove rows where columns 3 (additional_host_info) and 4 (additional_location_info) are empty.
 # Compare the S3 version with the local version.
 diff="$(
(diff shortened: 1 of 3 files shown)


kairstenfay:
Thanks for making these scripts more robust, @ttung! 

I think my only minor comment would be that we could make AWS `key` and `bucket` parsing similar to what we've done previously in `./bin/rebuild-staging`. There are obviously many ways to do this, but it might be nice to conform to the example in the [`hashes-match()` function](https://github.com/nextstrain/ncov-ingest/blob/ddff1c6421d1dcb33c09b460f7cbac40ecd7b6be/bin/rebuild-staging#L70). To be clear, I don't think this is a blocker for merging, and I'm eager to see what my teammates think.

ttung:
I updated to use the same three-liner that `bin/rebuild-staging` uses.

bin/notify-on-additional-info-change
```diff
@@ -8,6 +8,12 @@ bin="$(dirname "$0")"
 src="${1:?A source additional info TSV file is required as the first argument.}"
 dst="${2:?A destination additional info TSV s3:// URL is required as the second argument.}"
 
+# if the file is not already present, just exit
+s3path="${2#s3://}"
+bucket="${dst_s3path%%/*}"
+key="${dst_s3path#*/}"
```
tsibley, line 14: Oh, actually, these use an undeclared variable name `dst_s3path`. I'll fix this locally and then merge.

tsibley:
I fixed the copy/paste typo and put the incantation in their little program for less repetition/places to adjust it later. Merged as #63 with @ttung credited as co-author. Thanks!

PR #59, opened 2020-06-01
Standardize locations

### LINKED ISSUE #57: standardize characters and casing in location information during ingest
Currently, when new data is ingested, manual adjustments are frequently required for `division` and `location` fields to fix mundane, recurring issues. For example, `division` and `location` fields frequently include an underscore to separate words, rather than a space. Other times, these fields are spelled correctly, but with incorrect casing (all uppercase, rather than titlecase, for example). An example of this is sequences from Valencia, Spain, which almost always ingested with `division Comunitat_Valenciana`. Currently, these require manually adding each strain to `annotations.tsv` to set `division Comumitat Valenciana`. 

A previous example was that we received sequences from the USA, which were ingested as `region NORTH AMERICA`. These had to be manually altered to `region North America`.

It would be amazing if all location fields (`region`, `country`, `division`, and `location`) could be automatically processed to: 
1. replace underscores between words with spaces
2. convert to title case

These 2 checks/fixes would substantially reduce the amount of manual annotation that would need to occur for some batches of sequences in which this is common. This had been reasonably manageable until yesterday, when we received ~ 400 sequences from Spain, most of which required manual underscore removal across multiple fields.

kairstenfay:
~Replaces underscores with spaces in resolution names.~ (done via #60) 
Sets them to title case.

### Related issue(s)  
Resolves #57

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ fsspec = "*"
 pandas = ">=1.0.1"
 s3fs = "*"
 biopython = "*"
+regex = "*"
 
 [requires]
 python_version = "3.6"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "c76521db7a1266c622dd13cf7219af39be4e04a9ed53023573f6e1bcb2c11394"
+            "sha256": "295401b51ad3b0910d751938d3ea390e8379a747e2d1e4f50bd3f017e9ce5664"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -40,17 +40,17 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:26f8564b46d009b8f4c6470a6d6cde147b282a197339c7e31cbb0fe9fd9e5f5d",
-                "sha256:f59d0bd230ed3a4b932c5c4e497a0e0ff3c93b46b7e8cde54efb6fe10c8266ba"
+                "sha256:bcaa88b2f81b88741c47da52f3414c876236700441df87b6198f860e6a200d6f",
+                "sha256:e974e7a3bbdbd6a73ffc07bea5fa0c0744a5a8b87dcca94702597176e3de465e"
             ],
-            "version": "==1.13.20"
+            "version": "==1.13.23"
         },
         "botocore": {
             "hashes": [
-                "sha256:990f3fc33dec746829740b1a9e1fe86183cdc96aedba6a632ccfcbae03e097cc",
-                "sha256:d4cc47ac989a7f1d2992ef7679fb423a7966f687becf623a291a555a2d7ce1c0"
+                "sha256:5831068c9b49b4c91b0733e0ec784a7733d8732359d73c67a07a0b0868433cae",
+                "sha256:7778957bdc9a25dd33bb4383ebd6d45a8570a2cbff03d1edf430fdacec2b7437"
             ],
-            "version": "==1.16.20"
+            "version": "==1.16.23"
         },
         "certifi": {
             "hashes": [
@@ -154,29 +154,29 @@
         },
         "numpy": {
             "hashes": [
-                "sha256:00d7b54c025601e28f468953d065b9b121ddca7fff30bed7be082d3656dd798d",
-                "sha256:02ec9582808c4e48be4e93cd629c855e644882faf704bc2bd6bbf58c08a2a897",
-                "sha256:0e6f72f7bb08f2f350ed4408bb7acdc0daba637e73bce9f5ea2b207039f3af88",
-                "sha256:1be2e96314a66f5f1ce7764274327fd4fb9da58584eaff00b5a5221edefee7d6",
-                "sha256:2466fbcf23711ebc5daa61d28ced319a6159b260a18839993d871096d66b93f7",
-                "sha256:2b573fcf6f9863ce746e4ad00ac18a948978bb3781cffa4305134d31801f3e26",
-                "sha256:3f0dae97e1126f529ebb66f3c63514a0f72a177b90d56e4bce8a0b5def34627a",
-                "sha256:50fb72bcbc2cf11e066579cb53c4ca8ac0227abb512b6cbc1faa02d1595a2a5d",
-                "sha256:57aea170fb23b1fd54fa537359d90d383d9bf5937ee54ae8045a723caa5e0961",
-                "sha256:709c2999b6bd36cdaf85cf888d8512da7433529f14a3689d6e37ab5242e7add5",
-                "sha256:7d59f21e43bbfd9a10953a7e26b35b6849d888fc5a331fa84a2d9c37bd9fe2a2",
-                "sha256:904b513ab8fbcbdb062bed1ce2f794ab20208a1b01ce9bd90776c6c7e7257032",
-                "sha256:96dd36f5cdde152fd6977d1bbc0f0561bccffecfde63cd397c8e6033eb66baba",
-                "sha256:9933b81fecbe935e6a7dc89cbd2b99fea1bf362f2790daf9422a7bb1dc3c3085",
-                "sha256:bbcc85aaf4cd84ba057decaead058f43191cc0e30d6bc5d44fe336dc3d3f4509",
-                "sha256:dccd380d8e025c867ddcb2f84b439722cf1f23f3a319381eac45fd077dee7170",
-                "sha256:e22cd0f72fc931d6abc69dc7764484ee20c6a60b0d0fee9ce0426029b1c1bdae",
-                "sha256:ed722aefb0ebffd10b32e67f48e8ac4c5c4cf5d3a785024fdf0e9eb17529cd9d",
-                "sha256:efb7ac5572c9a57159cf92c508aad9f856f1cb8e8302d7fdb99061dbe52d712c",
-                "sha256:efdba339fffb0e80fcc19524e4fdbda2e2b5772ea46720c44eaac28096d60720",
-                "sha256:f22273dd6a403ed870207b853a856ff6327d5cbce7a835dfa0645b3fc00273ec"
+                "sha256:0172304e7d8d40e9e49553901903dc5f5a49a703363ed756796f5808a06fc233",
+                "sha256:34e96e9dae65c4839bd80012023aadd6ee2ccb73ce7fdf3074c62f301e63120b",
+                "sha256:3676abe3d621fc467c4c1469ee11e395c82b2d6b5463a9454e37fe9da07cd0d7",
+                "sha256:3dd6823d3e04b5f223e3e265b4a1eae15f104f4366edd409e5a5e413a98f911f",
+                "sha256:4064f53d4cce69e9ac613256dc2162e56f20a4e2d2086b1956dd2fcf77b7fac5",
+                "sha256:4674f7d27a6c1c52a4d1aa5f0881f1eff840d2206989bae6acb1c7668c02ebfb",
+                "sha256:7d42ab8cedd175b5ebcb39b5208b25ba104842489ed59fbb29356f671ac93583",
+                "sha256:965df25449305092b23d5145b9bdaeb0149b6e41a77a7d728b1644b3c99277c1",
+                "sha256:9c9d6531bc1886454f44aa8f809268bc481295cf9740827254f53c30104f074a",
+                "sha256:a78e438db8ec26d5d9d0e584b27ef25c7afa5a182d1bf4d05e313d2d6d515271",
+                "sha256:a7acefddf994af1aeba05bbbafe4ba983a187079f125146dc5859e6d817df824",
+                "sha256:a87f59508c2b7ceb8631c20630118cc546f1f815e034193dc72390db038a5cb3",
+                "sha256:ac792b385d81151bae2a5a8adb2b88261ceb4976dbfaaad9ce3a200e036753dc",
+                "sha256:b03b2c0badeb606d1232e5f78852c102c0a7989d3a534b3129e7856a52f3d161",
+                "sha256:b39321f1a74d1f9183bf1638a745b4fd6fe80efbb1f6b32b932a588b4bc7695f",
+                "sha256:cae14a01a159b1ed91a324722d746523ec757357260c6804d11d6147a9e53e3f",
+                "sha256:cd49930af1d1e49a812d987c2620ee63965b619257bd76eaaa95870ca08837cf",
+                "sha256:e15b382603c58f24265c9c931c9a45eebf44fe2e6b4eaedbb0d025ab3255228b",
+                "sha256:e91d31b34fc7c2c8f756b4e902f901f856ae53a93399368d9a0dc7be17ed2ca0",
+                "sha256:ef627986941b5edd1ed74ba89ca43196ed197f1a206a3f18cc9faf2fb84fd675",
+                "sha256:f718a7949d1c4f622ff548c572e0c03440b49b9531ff00e4ed5738b459f011e8"
             ],
-            "version": "==1.18.4"
+            "version": "==1.18.5"
         },
         "pandas": {
             "hashes": [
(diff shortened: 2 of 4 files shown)


bin/transform
```diff
@@ -95,13 +95,18 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
-    gisaid_data['location']    = geographic_data[3].str.strip()
-
+    geographic_data = gisaid_data['location'].str.split(r'\s*/\s*', expand=True)
+    location_columns = ['region', 'country', 'division', 'location']
+
+    for column, index in zip(location_columns, range(len(location_columns))):
+        gisaid_data[column] = geographic_data[index] \
+            .str.replace('_', ' ') \
+            .str.strip() \
+            .str.title()
```
joverlee521, line 105: Do we want to write a custom `title()` function that doesn't capitalize articles such as the, of, de, do, etc. or are we okay with manually correcting those?

lmoncla: hmm yeah we definitely don't want to capitalize those, nor do we want to manually correct them. I had thought that title case did automatically not capitalize those

joverlee521: There are Python libraries like https://github.com/ppannuto/python-titlecase that try to be smart about title casing, but it only works for English words.

lmoncla: I see. Ok, then yes, it would be preferable maybe to have a function to specify this, as it will definitely be not great to have to fix those manually

kairstenfay: Another perhaps simpler option than finding all the non-English articles and states in locations (e.g. `Madison WI`) could be to cast to title case only if the incoming string is in all uppercase.

kairstenfay: I pushed a new commit that only casts resolution names that are already in uppercase to title case.

bin/transform
```diff
@@ -99,7 +99,14 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
     for column, index in zip(location_columns, range(len(location_columns))):
-        gisaid_data[column] = geographic_data[index].str.replace('_', ' ').str.strip()
+        gisaid_data[column] = geographic_data[index] \
+            .str.replace('_', ' ') \
+            .str.strip() \
+            .str.title()
+
+        # Manual fixes for non-title case locations
+        gisaid_data.loc[
+            (pd.notnull(gisaid_data[column]) & (gisaid_data[column] == 'Usa')), column] = 'USA'
```
kairstenfay: (although I haven't tested, my guess is) It would probably be more computationally efficient to prevent casting to 'Usa' altogether and add `gisaid_data[column] != 'USA'` to the mask used to cast to title case.

bin/transform
```diff
@@ -95,13 +95,22 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
     """
-    geographic_data = gisaid_data['location'].str.split('\s*/\s*', expand=True)
+    geographic_data = gisaid_data['location'].str.split(r'\s*/\s*', expand=True)
+    location_columns = ['region', 'country', 'division', 'location']
 
-    gisaid_data['region']      = geographic_data[0].str.strip()
-    gisaid_data['country']     = geographic_data[1].str.strip()
-    gisaid_data['division']    = geographic_data[2].str.strip()
-    gisaid_data['location']    = geographic_data[3].str.strip()
+    for column, index in zip(location_columns, range(len(location_columns))):
```
joverlee521, line 101: I believe this can be simplified with `enumerate()`

bin/transform
```diff
@@ -95,13 +95,22 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        gisaid_data[column] = geographic_data[index].str.replace('_', ' ').str.strip()
 
+        not_null = pd.notnull(gisaid_data[column])
+        is_uppercase = gisaid_data[column].str.upper() == gisaid_data[column]
```
joverlee521, line 105: I'm not entirely sure this is what we want? I assume we want to convert all locations to title case to standardize names. For example, in the live site, there's currently `New Taipei city` and `New Taipei City`.

bin/transform
joverlee521, line 101: I believe this can be simplified with `enumerate()`?

kairstenfay: Ah yes I forgot about that function.

bin/transform
joverlee521: Would it be weird if everything was just in uppercase?

kairstenfay: I see the following non-mutually exclusive programmatic options:
* we could use a third-party library like `titlecase` that handles English articles
* we could "whitelist" non-English articles like "de", "la", "do", etc. 
* we could whitelist US state abbreviations like DC, WA, WI that appear in location names

Locations that slip through could continue to be manually annotated or added to the whitelist. Which seems like more work -- the status quo of annotating locations or fixing improperly title cased locations (either in annotations.tsv or in this script)?

joverlee521: I think it's worthwhile to standardize as much as possible within the script and manually annotate things that slip through. If a certain pattern is seen often enough, then we add it to the hardcoded whitelist?

lmoncla: I think it will be less work to have as much as possible be standardized in the script, and fixing things that slip through manually. I think that would still be an improvement over the status quo. To be honest, I am most excited about the underscore replacement and I did not give adequate thought to how complicated the casing is.

kairstenfay: @lmoncla @joverlee521 I opened #59 to be able to immediately merge the underscore replacement while we continue working/thinking through title casing on this PR.

kairstenfay:
Rebased onto master after merging #60

bin/transform
kairstenfay: @joverlee521 Ok, I have reverted my old title casing commits in favor of a new one. 
This one uses `.title()` because the benefit of using the third party `titlecase` library is that it uses the context of the whole string. For example, 
```
>>> from titlecase import titlecase
>>> titlecase('of')
'Of'
>>> titlecase('united stats of america')
'United Stats of America'
```

Because `titlecase` does not work with non-English articles (de, la, etc.), it was necessary to re-process any `titlecase`-produced strings either by tokenization and iteration or string substitution to set the articles back to their correct capitalization. Therefore, I opted in favor of capitalizing each token (i.e. word) separately to process each string only once. This reduced the advantage of `titlecase` which is why I instead used `.title()`. 

I also found another edge case for Francophone locations where hyphen-delimited locations (e.g. Braine-le-Compte) don't register as separate tokens. However all parts do get capitalized by `.title()`, so I wrote a separate regex that looks for this pattern and prevents manipulation on it.

kairstenfay: I'm open to criticism here if we decide we do want to process entire resolution name strings with `titlecase` then retroactively substitute articles back to their correct capitalization.

bin/transform
```diff
@@ -95,11 +97,47 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
     location_columns = ['region', 'country', 'division', 'location']
 
+    # Manually curate list of tokens and patterns that should not be cast to
+    # title case
+    articles = [
+        'de', 'del', 'di', 'do', 'la', 'le', 'las', 'los', 'nad', 'of', 'op',
+        'the', 'y'
+    ]
+    locations = ['USA', 'DC']
+    whitelist_tokens = articles + locations
```
tsibley, line 128: A `set` is a better data structure for these variables.

bin/transform
```diff
@@ -95,11 +97,47 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+    # Some Francophone locations are hyphen-delimited (e.g. Braine-le-Comte).
+    # These names don't get split into separate tokens, but all pieces are
+    # capitalized by `.title()`. Assume these names are already properly cased,
+    # and prevent them from being transformed.
+    whitelist_regex = r'\S*-le?-?'
```
tsibley, line 133: You can avoid this issue and other similar issues by splitting on Unicode word boundaries (`\b`) instead of whitespace. The standard `re` library doesn't support splitting on zero-length matches, but the `regex` library does. We use it elsewhere for better Unicode and regex support. For example:

```python
# Run with: python3 -m doctest -v titlecase.py
import regex

def titlecase(text, articles = {}, abbrev = {}):
    """
    >>> articles = {'a', 'and', 'of', 'the', 'le'}
    >>> abbrev = {'USA', 'DC'}

    >>> titlecase("the night OF THE LIVING DEAD", articles)
    'The Night of the Living Dead'

    >>> titlecase("BRAINE-LE-COMTE, FRANCE", articles)
    'Braine-le-Comte, France'

    >>> titlecase("auvergne-RHÔNE-alpes", articles)
    'Auvergne-Rhône-Alpes'

    >>> titlecase("washington DC, usa", articles, abbrev)
    'Washington DC, USA'
    """
    words = enumerate(regex.split(r'\b', text, flags = regex.V1))

    def changecase(index, word):
        casefold = word.casefold()
        upper = word.upper()
        title = word.title()

        if upper in abbrev:
            return upper
        elif casefold in articles and index != 1:
            return casefold
        else:
            return title

    return ''.join(changecase(i, w) for i, w in words)
```

tsibley:
Also, to avoid churn, I'd say drop the first four commits (two commits + their reverts).

kairstenfay:
> Also, to avoid churn, I'd say drop the first four commits (two commits + their reverts).

I agree and plan to do that before merge. I just didn't want to lose track of our decision making on this PR by force-pushing the commits into oblivion too early.

bin/transform
kairstenfay: Thanks for writing this handy function. I'll incorporate it into the latest commit. 
Do we want to use `.casefold()` if it might risk losing special characters? e.g. "ß" -> "ss"

tsibley: Ah, good catch! I think the condition should be `casefold in articles` but that branch should `return lower`.

kairstenfay: I incorporated these changes into a new commit.

bin/transform
```diff
@@ -97,47 +97,61 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+        'The Night of the Living Dead'
+
+        >>> titlecase("BRAINE-LE-COMTE, FRANCE", articles)
+        'Braine-le-Comte, France'
+
+        >>> titlecase("auvergne-RHÔNE-alpes", articles)
+        'Auvergne-Rhône-Alpes'
+
+        >>> titlecase("washington DC, usa", articles, abbrev)
+        'Washington DC, USA'
```
kairstenfay, line 120: Does it make sense to keep the doc tests here? They fail when I run `python3 -m doctest -v transform` locally.

bin/transform
```diff
@@ -95,11 +97,61 @@ def parse_geographic_columns(gisaid_data: pd.DataFrame) -> pd.DataFrame:
...
+
+        return ''.join(changecase(i, w) for i, w in words)
+
     geographic_data = gisaid_data['location'].str.split(r'\s*/\s*', expand=True)
     location_columns = ['region', 'country', 'division', 'location']
 
+    # Manually curate set of tokens that should not be cast to title case
+    articles = {
+        'de', 'del', 'di', 'do', 'l', 'la', 'le', 'las', 'los', 'nad', 'of',
+        'op', 'the', 'y'
```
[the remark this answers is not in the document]
kairstenfay: > @kairstenfay Thank you for tackling the casing problem!
> 
> Something I noticed while testing is that the casing in the `annotations.tsv` can vary as well. It's outside the scope of this PR, but just wanted to point it out. (It might be something to add as a check in #54?)

Oh, good catch. I didn't realize annotations needed title casing as well. It's not necessarily outside the scope of this PR if we move the title casing form `parse_geographic_columns()` to a new function after `update_metadata()`. Maybe the function could be named `titlecase_columns()` and accept a list of columns that should be title cased. Perhaps then we could even extend this to other columns like author names, etc. Thoughts?

kairstenfay:
> @kairstenfay Thank you for tackling the casing problem!
> 
> Something I noticed while testing is that the casing in the `annotations.tsv` can vary as well. It's outside the scope of this PR, but just wanted to point it out. (It might be something to add as a check in #54?)

Oh, good catch. I didn't realize annotations needed title casing as well. It's not necessarily outside the scope of this PR if we move the title casing from parse_geographic_columns() to a new function after update_metadata(). Maybe the function could be named titlecase_columns() and accept a list of columns that should be title cased. Perhaps then we could even extend this to other columns like author names, etc. Thoughts?

bin/transform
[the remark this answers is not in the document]
kairstenfay: +1 on des and en. 
I see "sur" occurs commonly in Francophone countries (Montignies-Sur-Sambre). It's also its own word meaning "south" in Spanish that may show up in some locations where it ought to be capitalized. Given "sur"s prevalence as an article in Francophone locations, though, I think it would be fair to manually annotate Hispanic locations.

kairstenfay:
> Oh, good catch. I didn't realize annotations needed title casing as well. It's not necessarily outside the scope of this PR if we move the title casing from parse_geographic_columns() to a new function after update_metadata(). Maybe the function could be named titlecase_columns() and accept a list of columns that should be title cased. Perhaps then we could even extend this to other columns like author names, etc. Thoughts?

After commenting on the "sur" article you proposed, I realized this is not so straightforward. We are using _annotations.tsv_ to manually update locations that are edge cases in our title casing logic. Therefore we should probably not title case _annotations.tsv_.

joverlee521:
> > Oh, good catch. I didn't realize annotations needed title casing as well. It's not necessarily outside the scope of this PR if we move the title casing from parse_geographic_columns() to a new function after update_metadata(). Maybe the function could be named titlecase_columns() and accept a list of columns that should be title cased. Perhaps then we could even extend this to other columns like author names, etc. Thoughts?
> 
> After commenting on the "sur" article you proposed, I realized this is not so straightforward. We are using _annotations.tsv_ to manually update locations that are edge cases in our title casing logic. Therefore we should probably not title case _annotations.tsv_.

I was going to say the same thing 😄 
Maybe we just need to check that at least _within_ the annotations.tsv the casing is the same?

kairstenfay:
> I was going to say the same thing smile
> Maybe we just need to check that at least _within_ the annotations.tsv the casing is the same?

Yes, I think that is a good validation step to run.

joverlee521:
> Maybe we just need to check that at least within the annotations.tsv the casing is the same?

An additional step is to make sure the final casing matches the casing provided in the annotations.tsv if the location exists in the annotations.tsv?

kairstenfay:
I just forced a re-push to remove potentially offensive and racially charged language used to refer to a list of items to be excluded from transformation. My apologies.

bin/transform
tsibley: Would be good to keep the doc tests so we can check them for any future changes.

But ugh. This error is because doctest assumes that files which don't end in `.py` are doctests embedded in text, not code, files:

> Because the file name does not end with .py, doctest infers that it must be run with testfile(), not testmod().

From the end of [Simple usage: checking examples in a text file](https://docs.python.org/3/library/doctest.html#simple-usage-checking-examples-in-a-text-file).

I think an easy fix here is to move `titlecase` into the new lib/… module once #55 is merged (soon).

tsibley:
I rebased this onto master and fixed the doctest issue, but running locally, I get:

```
$ pipenv run ./bin/transform-gisaid data/gisaid.ndjson
Traceback (most recent call last):
  File "./bin/transform-gisaid", line 256, in <module>
    gisaid_data = preprocess(gisaid_data)
  File "./bin/transform-gisaid", line 58, in preprocess
    gisaid_data = standardize_dataframe(gisaid_data, mapper)
  File "lib/utils/transform.py", line 36, in standardize_dataframe
    if df[column].dtype == "string":
TypeError: data type "string" not understood
```

thoughts? I've run out of time today (childcare), but will dig more tomorrow if someone doesn't know what's going on.

kairstenfay:
> pipenv run ./bin/transform-gisaid data/gisaid.ndjson

This is happening for me on the `master` branch.

joverlee521:
> This is happening for me on the master branch.

From my digging, this seems to be the culprit:
```Python
for field in column_mapper:
        if field not in df:
            df[field] = pd.NA
```
The empty column has a `numpy.dtype` which does not recognize the data type "string".

kairstenfay:
Rebased onto master after merging #65 which addresses the issue @tsibley raised.

PR #65, opened 2020-06-16
utils: Ensure empty columns have a Pandas-native dtype

kairstenfay:
See commit messages.

Resolves the issue [described by @tsibley in #59](https://github.com/nextstrain/ncov-ingest/pull/59#issuecomment-644450742) using @joverlee521's debugging findings.

### DIFF
--- a/lib/utils/transform.py
+++ b/lib/utils/transform.py
@@ -31,7 +31,7 @@ def standardize_dataframe(df: pd.DataFrame, column_mapper: dict) -> pd.DataFrame
     # Normalize all string columns to Unicode Normalization Form C, for
     # consistent, predictable string comparisons.
     for column in df:
-        if df[column].dtype == "string":
+        if df[column].notnull().any() and df[column].dtype == "string":
             df[column] = df[column].str.normalize("NFC").str.strip()
 
     # Drop entries with length less than 15kb and reset index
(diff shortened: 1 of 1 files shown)


tsibley:
Nice debugging! Conditionalizing the dtype comparison seems like fixing the symptom, not the root problem, though. What about making sure the empty columns have a valid Pandas-native nullable dtype?

```python
df[field] = pd.Series(dtype = "string")
```

will fill `field` with `pd.NA` and `df[field].dtype` will still work with Pandas-native dtypes (as opposed to just numpy-native dtypes). An empty Series of `Int64` or `Boolean`, the other Pandas-native dtypes, would also work.

kairstenfay:
> Nice debugging! Conditionalizing the dtype comparison seems like fixing the symptom, not the root problem, though. What about making sure the empty columns have a valid Pandas-native nullable dtype?
> 
> ```python
> df[field] = pd.Series(dtype = "string")
> ```
> 
> will fill `field` with `pd.NA` and `df[field].dtype` will still work with Pandas-native dtypes (as opposed to just numpy-native dtypes). An empty Series of `Int64` or `Boolean`, the other Pandas-native dtypes, would also work.

Agreed, that's a superior solution. I incorporated this suggestion into a new commit.

PR #66, opened 2020-06-16
upload-to-s3: Set Content-Type and Content-Encoding headers

tsibley:
Crucial when serving files out of the S3 bucket via Cloudfront.

### DIFF
--- a/bin/upload-to-s3
+++ b/bin/upload-to-s3
@@ -35,7 +35,7 @@ main() {
             gzip -c "$src"
         else
             cat "$src"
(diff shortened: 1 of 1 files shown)


tsibley:
The difference in headers can be shown by these two commands:

```
curl -I http://data.nextstrain.org/ncov-ingest/genbank_metadata.tsv.gz
curl -I http://data.nextstrain.org/ncov-ingest/branch/genbank/s3-content-headers/genbank_metadata.tsv.gz
```

PR #70, opened 2020-06-19
transform: standardize date format

### LINKED ISSUE #69: Standardize dates
**Current Behavior**  
We don't do any date pre-processing or standardization, allowing invalid ISO 8601 dates (like 2020-5-8) to slip through. This was detected by the recent addition of the `flagged_metadata.txt` file where a bunch of invalid dates were flagged for not being less than or equal to the current date. 

**Expected behavior**  
Non-zero-padded or otherwise invalid dates like 2020-5-8 should be standardized to their expected format, e.g. 2020-05-08.

joverlee521:
Standardize date columns within shared function `standardize_dataframe`
instead of just in `transform-genbank` since it's become apparent that
GISAID dates may not be in ISO 8601 date format.

Stop relying on pandas `to_datetime` function to format date columns
since it infers the dates from incomplete date strings
(e.g. 2020 -> 2020-01-01 and 2020-01 -> 2020-01-01).

### Related issue(s)  

Fixes #69

### DIFF
--- a/lib/utils/transform.py
+++ b/lib/utils/transform.py
@@ -146,6 +146,9 @@ def format_date(date_string: str, expected_formats: set) -> str:
     >>> format_date("2020-1-15", expected_formats)
     '2020-01-15'
 
+    >>> format_date("2020-1-1", expected_formats)
+    '2020-01-01'
+
     >>> format_date("2020-01-15", expected_formats)
     '2020-01-15'
 
--- a/lib/utils/transform.py
+++ b/lib/utils/transform.py
@@ -2,6 +2,7 @@
 import regex
 import fsspec
 import pandas as pd
+from datetime import datetime
 from typing import List, Optional, Set, Union
 
 # Note: 'sequence' should NEVER appear in these lists!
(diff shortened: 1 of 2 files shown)


joverlee521:
[standardize dates in ncov-ingest](https://trello.com/c/zUPS3Bsb/628-standardize-dates-in-ncov-ingest)

lib/utils/transform.py
```diff
@@ -121,3 +128,34 @@ def changecase(index, word):
...
+    >>> format_date("2020-1-15", expected_formats)
+    '2020-01-15'
+
+    >>> format_date("2020-01-15", expected_formats)
+    '2020-01-15'
+
+    >>> format_date("2020-01-15T00:00:00Z", expected_formats)
+    '2020-01-15'
+    """
+    for format in expected_formats:
```
kairstenfay, line 158: `format` is a built-in function in python, so it might be better to choose a different variable name here.

joverlee521: Ah, thanks. Updated in latest commit.

Issue #78, opened 2020-07-08
Location hierarchy formatting issues


kairstenfay:
The [source-data/location_hierarchy.tsv](https://github.com/nextstrain/ncov-ingest/blob/master/source-data/location_hierarchy.tsv) file contains some issues where 
* some rows only have 3 columns instead of 4 (which GitHub currently points out when you view this file in the browser), i.e. a missing final tab
* some rows have 5 columns instead of 4, i.e. a surplus final tab

To prevent future formatting issues with the manually updated location hierarchy file, I suggest adding a script like `bin/clean-location-hierachy` that either
* does some bash magic (something like `cut -f1-4 source-data/location-hierarchy.tsv` to grab only the first four columns, then something else I haven't quite figured out to add a trailing tab to rows missing the 4th column -- awk and gsub()?), or
* loads it into pandas and print it back out again with only 4 columns, tab-delimited.

I also suggest sorting the file on region, country, division, and location in case a row is misplaced. 

This cleaning script should be added to the ingest scripts.

joverlee521:
Just had a thought maybe we should make this a new workflow to run whenever there's a push to `source-data/location_hierarchy.tsv` instead of within the ingest scripts.

kairstenfay:
> Just had a thought maybe we should make this a new workflow to run whenever there's a push to `source-data/location_hierarchy.tsv` instead of within the ingest scripts.

That's a great idea!

eharkins:
We could also add an option to the script to add in the the output from [the slack location hierarchy additions file](https://github.com/nextstrain/ncov-ingest/blob/eca007ed26a8832c0ebfcc47d43cb016e8776f26/bin/notify-on-location-hierarchy-addition) so that you could run it locally to add those new location hierarchies and then manually remove the wrong ones or duplicates (there will be fewer duplicates with this clean up script included in automation) instead of manually adding the correct new ones. 

Sorry if this belongs better in https://github.com/nextstrain/ncov/pull/455

kairstenfay:
> We could also add an option to the script to add in the the output from [the slack location hierarchy additions file](https://github.com/nextstrain/ncov-ingest/blob/eca007ed26a8832c0ebfcc47d43cb016e8776f26/bin/notify-on-location-hierarchy-addition) so that you could run it locally to add those new location hierarchies and then manually remove the wrong ones or duplicates (there will be fewer duplicates with this clean up script included in automation) instead of manually adding the correct new ones.
> 
[...]

PR #82, opened 2020-07-24
tsv cleanup script and workflow

eharkins:
This PR introduces the following changes (hopefully as discussed in #78):
- A script to automatically sort and clean up the number of tabs in tsv metadata files in this repo
- An automated workflow to run that script on source-data/gisaid_annotations.tsv and source-data/location_hierarchy.tsv and commit those changes to the branch on which the commit that triggered the workflow was pushed
- A ton of changes to source-data/gisaid_annotations.tsv and source-data/location_hierarchy.tsv according to the way the script sorted them. Some of these are because things were out of alphabetical order (both source-data/gisaid_annotations.tsv and source-data/location_hierarchy.tsv), and others have to do with comments and splitting of paper annotations (source-data/gisaid_annotations.tsv). I included all these changes since if we start using this workflow on master, those changes will happen almost immediately anyway but we will have to update these changes when we are ready to merge. Any feedback on how the new sorting meets the expectations discussed ([e.g.](https://github.com/nextstrain/ncov-ingest/issues/78#issuecomment-657076680)) would be appreciated! 

To merge this, we will also need to :
- [ ] change the branch name in the workflow to master
- [ ] rerun the sorting and cleanup according to the latest metadata files.
- [ ] decide whether we want to apply the script (automatically or otherwise) to any manually maintained tsvs in this repo (such as the genbank annotations?)

### Related issue(s)  
https://github.com/nextstrain/ncov-ingest/issues/78

### Testing
Tested sorting and these two cases from #78:
> some rows only have 3 columns instead of 4 (which GitHub currently points out when you view this file in the browser), i.e. a missing final tab
> some rows have 5 columns instead of 4, i.e. a surplus final tab


emmahodcroft:
This is looking good @eharkins ! Though hard to wrap head around all the reorganisation. Just to clarify, this would run automatically after pushing, for example, new location_hierarchy or gisaid_annotations?
So I suppose we don't want to push this until we've decoupled those changes from triggering new data ingest (pulling from GISAID), or we'll have even more chances to get new data 🙃 

A good final test for this branch might be to double-check the metadata files produced from the 'original' and 'reorganised' files and ensure all is coming out the same - that's probably fastest way to check we didn't accidentally mess anything up!

eharkins:
> Just to clarify, this would run automatically after pushing, for example, new location_hierarchy or gisaid_annotations? So I suppose we don't want to push this until we've decoupled those changes from triggering new data ingest (pulling from GISAID), or we'll have even more chances to get new data 🙃

This is accurate, although if I remember correctly I was only having to cancel ingests upon manually pushing changes to those files and not when the automated commits from the new github actions workflow were pushed - this could be different on master since I know there are slight differences in those workflows on master vs other branches. I agree with being 100% sure we dont risk extra ingests since those are the main bottleneck / slow down for builds as I see it.

> A good final test for this branch might be to double-check the metadata files produced from the 'original' and 'reorganised' files and ensure all is coming out the same - that's probably fastest way to check we didn't accidentally mess anything up!

Great idea. In order to not mess with "production" builds, would that involve basically running [bin/ingest-gisaid](https://github.com/nextstrain/ncov-ingest/blob/master/bin/ingest-gisaid) but without the steps where the latest metadata is pushed to aws and the results are alerted in slack? I.e. a "local" run of that ingest process which would not affect the outcome of the next "official" ingest?

eharkins:
Just noting that maybe we can reuse the sorting script to sort some files in the ncov repo as well (e.g. defaults, and [some of the files](https://github.com/nextstrain/ncov/pull/475#issuecomment-666213081) used by the metadata script in that repo).

emmahodcroft:
I've asked in the 'stop ingest from annotations push' issue if we can go ahead and merge that (I think it seems ready?) - would be nice in any case to stop worrying about new data any time we ingest! And then we could merge this.

> Great idea. In order to not mess with "production" builds, would that involve basically running bin/ingest-gisaid but without the steps where the latest metadata is pushed to aws and the results are alerted in slack? I.e. a "local" run of that ingest process which would not affect the outcome of the next "official" ingest?

Yes, exactly. That should generate a metadata file that we can then compare to the 'official' metadata from AWS, and hopefully they'll match exactly and we're good to go! They should be done as closely together as possible to avoid grabbing any new data in one but not the other 🙃

eharkins:
> Yes, exactly. That should generate a metadata file that we can then compare to the 'official' metadata from AWS, and hopefully they'll match exactly and we're good to go!

👍 
> They should be done as closely together as possible to avoid grabbing any new data in one but not the other

Maybe I can avoid this constraint / risk of fetching new stuff if I just run the[ transform-gisaid script which applies the annotations and the check-locations script](https://github.com/nextstrain/ncov-ingest/blob/master/bin/ingest-gisaid#L50-L57)?

emmahodcroft:
I think just starting them roughly at the same time should be fine - you just wouldn't want to compare the AWS run from the afternoon to a fresh-download via local code in the evening, that's all! I am never 100% sure exactly what pulls from our local GISIAD copy vs GISAID directly, but one way to do it would definitely just be to start a local download the same time you trigger an ingest on AWS. But if you can find a fancier way, that would work too!

bin/clean-tsv-metadata
```diff
@@ -2,7 +2,7 @@
 import pandas as pd
 import argparse
 
-def clean_metadata_file(file_name, n_cols, header):
+def clean_metadata_file(file_name, n_cols, header, sort_col):
```
kairstenfay, line 5: I think changes in this file accidentally got pushed to this commit that adds a new workflow; can you split them out?

.github/workflows/clean-tsv.yml
```diff
@@ -11,7 +11,7 @@ on:
       - source-data/gisaid_annotations.tsv
 
 jobs:
-  locations:
+  clean-metadata:
```
kairstenfay, line 14: Can you fixup the changes in this commit to the previous commit? It looks like this commit is just edits that belong in the prior one.

.github/workflows/clean-tsv.yml
```diff
@@ -15,7 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: setup
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
kairstenfay, line 21: Ditto for fixing up these changes into the clean-tsv workflow original commit.

.github/workflows/clean-tsv.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: '3.x'
+    - name: Install dependencies
```
kairstenfay, line 22: Ditto re: fixup

bin/clean-tsv-metadata
```diff
@@ -0,0 +1,45 @@
...
+import argparse
+
+def clean_metadata_file(file_name, n_cols, header):
+    """
+    1. Read the metadata TSV file
+    2. Sort alphabetically and make sure n columns (n-1 tabs) each row
+    3. Write out (overwrite) metadata file
+    """
+    col_names = list(pd.read_csv(file_name, sep="\t", header=None, nrows=1).iloc[0]) if header else list(range(n_cols))
+    metadata = pd.read_csv(file_name,
```
kairstenfay, line 12: If you want this to work with an annotations file, you'll want to add `comment = '#'` as an argument.

eharkins: I just tried doing this and some effects of it are:
1. it removes all of the commented lines from the beginning of the file - this is ok since we could parse and re-add them with grep
2. it removes comments on the end of lines, e.g. 
```
Argentina/PAIS-A0003/2020      EPI_ISL_430795  division        Buenos Aires # previously Ciudad Autonoma de Buenos Aires
```
. This I don't immediately see a way around, but there could be one!

The way that the script works now, i.e. without `comment = '#'` passed to pd.read_csv, it seems to do reasonably well with:
1. fully #-commented lines; they still end up at the top (albeit sorted alphabetically). 
2. lines with comments on the end; they maintain their informative comments. 

Looking at b1f143a, is there anything in particular that it does that we don't want?

bin/clean-tsv-metadata
```diff
@@ -0,0 +1,45 @@
...
+import pandas as pd
+import argparse
+
+def clean_metadata_file(file_name, n_cols, header):
+    """
+    1. Read the metadata TSV file
+    2. Sort alphabetically and make sure n columns (n-1 tabs) each row
+    3. Write out (overwrite) metadata file
+    """
+    col_names = list(pd.read_csv(file_name, sep="\t", header=None, nrows=1).iloc[0]) if header else list(range(n_cols))
```
kairstenfay, line 11: It's unclear to me what value reading in the column names (header) separately adds to this script. 

If I run `pd.read_csv()` with `header=None`, then the columns are automatically named using zero-indexed integers. 

Additionally, I can think of an edge case where we may want to specify `--header` and also enforce a number of columns, which the code as written does not currently support.

I would recommend instead dropping the `col_names` assignment line, reading in the TSV with the `header` argument passed, and handling surplus columns in a separate part of the code.

bin/clean-tsv-metadata
```diff
@@ -14,6 +14,11 @@ def clean_metadata_file(file_name, n_cols, header):
                            header=0 if header else None, #needs to happen if a header already exists and you are overriding with "names"
                            names=col_names, # overriding column names; this adds missing tabs when there are too few
                            usecols=col_names) # using first n only; this removes extra tabs
+    if sort_col is not None:
+        if not header:
+            sort_col = int(sort_col) + 1
+        col_names.remove(sort_col)
+        col_names = [sort_col] + col_names
```
kairstenfay, line 21: You may want to consider how to handle ValueErrors that `col_names.remove()` throws in the event the given `sort_col` does not exist in the given TSV.

bin/clean-tsv-metadata
```diff
@@ -36,10 +41,15 @@ if __name__ == "__main__":
             type=int,
             help="Number of columns to enforce.\n")
 
+    parser.add_argument(
+            "--sort-col",
+            type=str,
+            help="A column name (or 1-based index for headerless tsvs) to prioritize sorting; if not used, sorting will be alphabetical by columns in the order they appear.`.\n")
+
     parser.add_argument(
```
kairstenfay: Out of curiosity, which TSV motivated you to add this option?

.github/workflows/clean-tsv.yml
```diff
@@ -0,0 +1,45 @@
...
+        pipenv run ./bin/clean-tsv-metadata --metadata source-data/location_hierarchy.tsv --header
+    - name: clean annotations
+      run: |
+        pipenv sync
+        grep -v "paper\|title" source-data/gisaid_annotations.tsv > nonpaper_annotations.tsv
+        pipenv run ./bin/clean-tsv-metadata --metadata nonpaper_annotations.tsv --n-cols 4
+        grep "paper\|title" source-data/gisaid_annotations.tsv > paper_annotations.tsv
+        pipenv run ./bin/clean-tsv-metadata --metadata paper_annotations.tsv --n-cols 4 --sort-col 1
+        cat nonpaper_annotations.tsv > source-data/gisaid_annotations.tsv
+        cat paper_annotations.tsv >> source-data/gisaid_annotations.tsv
```
kairstenfay, line 39: Why do you clean paper and non paper annotations separately?

.github/workflows/clean-tsv.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  push:
+    branches:
+      - clean-metadata
+    tags-ignore:
+      - '**'
+    paths:
+      - source-data/location_hierarchy.tsv
+      - source-data/gisaid_annotations.tsv
+
```
kairstenfay, line 12: I know you mentioned changing this `branches:` value to "master" before merging.
Just to be clear, you want this workflow to run on every new push to master?

bin/clean-tsv-metadata
```diff
@@ -36,10 +41,15 @@ if __name__ == "__main__":
             type=int,
             help="Number of columns to enforce.\n")
 
+    parser.add_argument(
+            "--sort-col",
+            type=str,
+            help="A column name (or 1-based index for headerless tsvs) to prioritize sorting; if not used, sorting will be alphabetical by columns in the order they appear.`.\n")
```
kairstenfay, line 47: Why did you choose to use 1-based indexing? I would assume that most python users are familiar with 0-based indexing and may even overlook this point.

kairstenfay:
It may be worth documenting somewhere (in the README or in the help description of this command/the n-cols option) how many columns are expected in the usual suspects: _gisaid_annotations.tsv_ and _location_hierarchy.tsv_.

bin/clean-tsv-metadata
kairstenfay: Ah yes, you're right. My apologies for recommending something without testing it myself. Please ignore that recommendation for now.

Tangentially related to annotations comments, I see that the first row that is a comment in an annotations TSV gets wrapped in double quotation marks (and extra double quotation marks appear around previously quoted strings). Here are my steps to reproduce:
```
head -n 5 source-data/gisaid_annotations.tsv > source-data/test_annotations.tsv 
tail -n 5 source-data/gisaid_annotations.tsv >> source-data/test_annotations.tsv 
cp source-data/test_annotations.tsv source-data/copy_test_annotations.tsv
./bin/clean-tsv-metadata --metadata source-data/copy_test_annotations.tsv --n-cols 4
```
I'm not sure what could possibly be causing this bug.

bin/clean-tsv-metadata
kairstenfay: Additionally, I'm not sure that this section of the code is doing what we want. e.g. 

```
head -n 5 source-data/gisaid_annotations.tsv > source-data/test_annotations.tsv 
tail -n 5 source-data/gisaid_annotations.tsv >> source-data/test_annotations.tsv 
cp source-data/test_annotations.tsv source-data/copy_test_annotations.tsv
./bin/clean-tsv-metadata --metadata source-data/copy_test_annotations.tsv --n-cols 4 --sort-col 3
```

results in the following error:
```
Traceback (most recent call last):
  File "./bin/clean-tsv-metadata", line 55, in <module>
    clean_metadata_file(args.metadata, args.n_cols, args.header, args.sort_col)
  File "./bin/clean-tsv-metadata", line 20, in clean_metadata_file
    col_names.remove(sort_col)
ValueError: list.remove(x): x not in list
```

Would you expect sorting by column `3` to work?

.github/workflows/clean-tsv.yml
eharkins: Yes, every push with changes to either of these files:
```
     paths:
      - source-data/location_hierarchy.tsv
      - source-data/gisaid_annotations.tsv
```
similar to https://github.com/nextstrain/ncov-ingest/blob/master/.github/workflows/ingest-gisaid-master.yml#L5-L10

.github/workflows/clean-tsv.yml
eharkins: https://github.com/nextstrain/ncov-ingest/issues/78#issuecomment-657076680

bin/clean-tsv-metadata
eharkins: See f95f2de, I couldn't think of a smarter way than reading it in a separate time.

bin/clean-tsv-metadata
eharkins: Should be fixed by 74c4dcb!

bin/clean-tsv-metadata
eharkins: The gisaid annotations so that I could keep paper annotations for the same sequences together.

.github/workflows/clean-tsv.yml
eharkins: Let me know if this is enough context, sorry to be cryptic/blunt!

bin/clean-tsv-metadata
eharkins: You're right, changed it in 74c4dcb!

bin/clean-tsv-metadata
eharkins: Should be fixed by ab18071!

eharkins:
@kairstenfay thanks for all the comments! I made almost all of the changes in https://github.com/nextstrain/ncov-ingest/pull/82#pullrequestreview-468830281 since I agreed with them overwhelmingly. Instead of `--ignore-header`, I went with `--no-header`, since it seems more likely you would be telling the script that no header exists in the file rather than ignoring one that does.

If this looks good, I can test a little bit more and address some of the final to-dos, e.g.:
- https://github.com/nextstrain/ncov-ingest/pull/82#issuecomment-675145894
- https://github.com/nextstrain/ncov-ingest/pull/82#issue-456044668

eharkins:
One thing to note about forcing files to have the correct amount of columns/tabs as I do here is that it doesn't alert us to cases where all the information is there but someone forgot to enter a tab between two columns in a row.

.github/workflows/clean-tsv.yml
kairstenfay: Got it, thanks for the context.

kairstenfay:
In my rebase, I accidentally caused a huge merge conflict in the automatically cleaned TSVs. I've dropped the automated commit for now until we are sure that the clean script works exactly as expected.

.github/workflows/clean-tsv.yml
```diff
@@ -27,14 +27,14 @@ jobs:
...
+        pipenv run ./bin/clean-tsv source-data/location_hierarchy.tsv source-data/location_hierarchy.tsv --n-cols 4
     - name: clean annotations
       run: |
         pipenv sync
         grep -v "paper\|title" source-data/gisaid_annotations.tsv > nonpaper_annotations.tsv
-        pipenv run ./bin/clean-tsv nonpaper_annotations.tsv --n-cols 4 --no-header
+        pipenv run ./bin/clean-tsv nonpaper_annotations.tsv nonpaper_annotations.tsv --n-cols 4 --no-header
         grep "paper\|title" source-data/gisaid_annotations.tsv > paper_annotations.tsv
-        pipenv run ./bin/clean-tsv paper_annotations.tsv --n-cols 4 --sort-col 1 --no-header
+        pipenv run ./bin/clean-tsv paper_annotations.tsv paper_annotations.tsv --n-cols 4 --sort-col 1 --no-header
```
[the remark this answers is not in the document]
kairstenfay: Oh, I wrote this in error. This is the output as well as the input. Got it.

kairstenfay:
> One thing to note about forcing files to have the correct amount of columns/tabs as I do here is that it doesn't alert us to cases where all the information is there but someone forgot to enter a tab between two columns in a row.

Hmm I feel like there are some important edge cases like that one that could be missed in an automated cleanup. We could post the automatically cleaned differences to Slack for review. 

Another edge case would be if someone accidentally entered an extra tab in between two columns and produced, for e.g., five columns in one row in the location_hierarchy.tsv. That fifth column would be dropped. 

I don't know if it makes sense to write code for all these possible edge cases or to rely on human review of the machine output. Thoughts?

.github/workflows/clean-tsv.yml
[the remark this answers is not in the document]
eharkins: This was just me passing the same name for the input and output files (overwriting), since I was assuming we didn't want preserve intermediate output in the github workflow.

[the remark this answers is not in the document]
kairstenfay: Yes, thank you, I eventually realized that.

eharkins:
Thanks @kairstenfay! 
> In my rebase, I accidentally caused a huge merge conflict in the automatically cleaned TSVs. I've dropped the automated commit for now until we are sure that the clean script works exactly as expected.

I've been using the following to accept the versions of those files from master (and then subsequently running the cleaning on them), but removing the automated commit also makes sense until this is ready to merge.
```
git checkout --theirs source-data/gisaid_annotations.tsv source-data/location_hierarchy.tsv
```

> Truncating columns w/ n-cols

This is fixed in 86c271b :
```
(nextstrain) MC02T50AUGYGR:ncov-ingest eharkins$ head -n 20 data/gisaid/metadata.tsv > data/gisaid/head_metadata.tsv
./bin/clean-tsv data/gisaid/head_metadata.tsv temp --n-cols 2
(nextstrain) MC02T50AUGYGR:ncov-ingest eharkins$ ./bin/clean-tsv data/gisaid/head_metadata.tsv temp --n-cols 2
(nextstrain) MC02T50AUGYGR:ncov-ingest eharkins$ head temp
strain  virus
Algeria/G0638_2264/2020 ncov
Algeria/G0640_2265/2020 ncov
```

> Exceeding the number of columns found in the TSV with --n-cols produces this error:

Error is more informative and human friendly now (47668cc):
```
(nextstrain) MC02T50AUGYGR:ncov-ingest eharkins$ ./bin/clean-tsv data/gisaid/head_metadata.tsv temp --n-cols 100
Too many columns specified: expected 100 and found 26
--n-cols 100 was passed, but there are not this many columns in data/gisaid/head_metadata.tsv. --n-cols can't add extra columns, it just enforces up to the existing number of columns in the tsv.
```
I can't think of an example where I've wanted to add an extra empty column in the context of the ncov builds, but if this is a function we want then it seems doable!


> Hmm I feel like there are some important edge cases [...] I don't know if it makes sense to write code for all these possible edge cases or to rely on human review of the machine output. Thoughts?

Maybe it's better to just have us run it locally when we are making additions to manually maintained tsvs during the course of a usual ingest. That way we are more likely to review the results of the clean up as a part of the diff we commit including our human-entered annotations, etc. And remove the workflow altogether. I don't want to over-automate things in a way that causes us more trouble than it's worth!

eharkins:
@emmahodcroft @MoiraZuber I talked to @kairstenfay on slack who expressed support for this plan, given the above:
1. remove the workflow in this PR since maybe it's more trouble than it's worth to accommodate edge cases / alert us to corrections made so we can review them by eye.
2. Make a final commit when we are ready to merge which cleans/sorts the locations and gisaid annotations.
3. From there on out, we would use bin/clean-tsv locally to sort and clean locations and annotations as we add them each day before committing. 

Kairsten is planning a few edits to the PR, but just wanted to check in about that plan in the meantime.

kairstenfay:
> @emmahodcroft @MoiraZuber I talked to @kairstenfay on slack who expressed support for this plan, given the above:
> 
>     1. remove the workflow in this PR since maybe it's more trouble than it's worth to accommodate edge cases / alert us to corrections made so we can review them by eye.
> 
>     2. Make a final commit when we are ready to merge which cleans/sorts the locations and gisaid annotations.
> 
>     3. From there on out, we would use bin/clean-tsv locally to sort and clean locations and annotations as we add them each day before committing.
> 
> 
> Kairsten is planning a few edits to the PR, but just wanted to check in about that plan in the meantime.

Thanks for reiterating our Slack convo, @eharkins ! I agree with Eli's intuition that automatically cleaning the TSV and committing its changes without human review may cause more headaches than is worthwhile. To be clear, we are still going to write this `clean-tsv` script, just not automate it (unless your feedback suggests we should do otherwise).

kairstenfay:
@eharkins When you get a moment, could you please test out the latest changes I pushed? 

If they look good, and assuming we want to move forward dropping the workflow, I can do some rebasing to remove the workflow and squash some of the development commits into larger ones.

kairstenfay:
> To me the thinking around dropping the automation and running this locally seems a good one, as far as catching any issues and ensuring we know it's all working right. If we get very confident in the future, we can always add in automation later if we feel it would help!

Great :+1:

> It would be great if we could have a quick overview of the main issues being checked/caught for each of the files in the script - it's mostly formatting, right? And checking for missing columns, etc? Also, am I understanding that we would run this after making any changes to `gisaid_annotations` & `location_hierarchy`, for example, before pushing back up to trigger a new `ingest`? (In other words, where in the 'process' is the best place to run this script?)

The script drops surplus columns (extra tabs), fills in missing columns (adds an empty tab to the final column(s) of each row where necessary), and performs a sort which is, by default, is in ascending order of each of the columns as they appear in the TSV. My understanding is this script should be run after making any changes to `gisaid_annotations` & `location_hierarchy` as you said. Great question. This should be documented. 
- [ ] Document how/when to run script in README

> RE some comments above about whether 'missing columns' (forgotten tabs) might be skipped or 'extra columns' (extra tab) might be dropped at the end, I wonder if it would be possible to just quickly run through the file at the start and look for any lines that have the incorrect number of columns, and print these for the user? Then they can easily catch these cases and figure out how to address them. Then it could be re-run with the corrections, in case they hadn't been handled correctly the first time. Is that possible?

`transform-gisaid` emits a warning and skips annotations that cannot be applied, for example if the formatting got messed up on a row. This `clean-tsv` script does not skip any annotations, but as you pointed out, it will drop extra columns at the end of a row. Running through the file once at the start to emit improperly formatted rows is possible, but it will require a decent amount of extra work on this PR. So my question for you, @emmahodcroft, is, do you consider this feature to be a blocker? Because `gisaid_annotations` & `location_hierarchy` are version controlled, you will be able to read the git diff after running `clean-tsv`. To be fair, though, the diff will also include rows moved around due to the automatic sorting. Maybe sorting should be optional so you can run the script once, preview the tab/column differences, then run again to sort?

emmahodcroft:
I've just been trying to test this out (sorry for the delay), to get an idea of how much things might change without warning before/after running the script. However, I'm running into some other behaviour for the `gisaid_annotations.tsv` file that I'm not sure we want.

The first thing is that it doesn't seem like the 'paper' data (`paper_url` and `strain`) are being sorted separately at the bottom - they are now intermixed with the rest of the data. I just think this makes the file much harder to 'scan' usefully (to get an overall idea of the corrections that are made for a country) - the paper data is pretty universally uninteresting - and is really huge in comparison to our other edits. I know I've discussed this with Eli in the past, but it was a while ago!

One solution to this would be to break the paper data out into a separate file and have the `ncov-ingest` script read in both of them and 'cat' the contents before processing as usual. Or, set a special way to process the `gisaid_annotations.tsv` file in the script. The first might be easier if we want the script to stay more general.


The other thing is that I think probably the comments should be tidied up a bit (this isn't just a problem after using the script - though some of them get more apparent). Small things like, if all comments start `# `  then they're sorted alphabetically after that (at the moment some are `#HongKong` so appear 'out of order'). And we can make things like our faux-'header' (`# strain EPI_ISL key value #comment`) start `##` so that it hopefully then appears just before the real entries. And finally some comments were apparently below whatever they referenced, but since some sorting (present already in the file, so not the script's fault!) 
they are just floating around at the top and don't make much sense.

I'll try to do the above for the `gisaid_annotations.tsv` file in `master` and then hopefully it can be pulled into this branch!

------

I've tried to take the first 1000 lines and then delete the comments, and just run this, in order to better compare the script and what might be missing - but there's so many order changes it's hard! I didn't think that through... 
Do you think we can we assume that the current `ncov-ingest` `transform` script is picking up and warning us of any lines with missing or excess tabs - just so we can be sure we're 'starting clean' with this script?

emmahodcroft:
Cleaned up comments now pushed to master!

eharkins:
> Cleaned up comments now pushed to master!

Thanks!
> The main thing is that it doesn't seem like the 'paper' data (paper_url and strain) are being sorted separately at the bottom - they are now intermixed with the rest of the data. I just think this makes the file much harder to 'scan' usefully (to get an overall idea of the corrections that are made for a country) - the paper data is pretty universally uninteresting - and is really huge in comparison to our other edits. I know I've discussed this with Eli in the past, but it was a while ago!

How are you running the script on the annotations file? It should sort the paper information separately and put it at the bottom when running[ as in the workflow](https://github.com/nextstrain/ncov-ingest/pull/82/files#diff-4de49ffd06685602811c7e1761fd83bcR31-R39), but with the `--output-tsv` option added before the second argument, like so:
```
pipenv sync
grep -v "paper\|title" source-data/gisaid_annotations.tsv > nonpaper_annotations.tsv
pipenv run ./bin/clean-tsv nonpaper_annotations.tsv --output-tsv nonpaper_annotations.tsv --n-cols 4 --no-header
grep "paper\|title" source-data/gisaid_annotations.tsv > paper_annotations.tsv
pipenv run ./bin/clean-tsv paper_annotations.tsv --output-tsv paper_annotations.tsv --n-cols 4 --sort-col 1 --no-header
cat nonpaper_annotations.tsv > source-data/gisaid_annotations.tsv
cat paper_annotations.tsv >> source-data/gisaid_annotations.tsv
```

emmahodcroft:
Oh dear - had no idea there's workflow!! That will explain a lot :) Ok, how can I run the whole workflow - and apologies!

kairstenfay:
Since there are so many steps, I wonder if it's worth creating a workflow that only runs on a manual trigger?

eharkins:
That could work, or a bash script that we run locally if we want to check the diff before committing sorting /cleaning changes as discussed above.

kairstenfay:
> That could work, or a bash script that we run locally if we want to check the diff before committing sorting /cleaning changes as discussed above.

Ah yes that's probably the right call since we want to run the changes locally.

emmahodcroft:
Yes, sorry - I think I'd clearly missed a step. I didn't realise there was a workflow which essentially 'shows' how to run each file 🙃 ! Though of course with the original automation idea, that makes sense!

Seems to me that a local file, shell or python, would be the best solution if we want to run locally. Perhaps this can be crafted from the existing workflow without too much trouble?

Just anticipating when I can run again locally but more 'informedly' 🙃, I wanted to ask again from above - it's going to be a little hard to see what might have been dropped/modified by the script due to all the ordering changes. Do we think we're confident that the current `ncov-ingest` ('new') `transform` would be throwing up errors for any lines that don't have the right number of columns - so that we're sure we're starting 'clean'? (And don't accidentally lose stuff which we could possibly fix if we catch it now?)

kairstenfay:
I'm working on addressing your concerns, @emmahodcroft . Meanwhile, in creating a script that runs sorting on all of our manually maintained TSVs, I found some behavior we likely don't want.

1. This script strips the quotations from `title` in `gisaid_annotations.tsv`. e.g.
```diff
-Argentina/PAIS-A0020/2020      EPI_ISL_430811  title   "COVID-19 in Latin America: Contrasting phylodynamic inference with epidemiological surveillance."
+Argentina/PAIS-A0020/2020      EPI_ISL_430811  title   COVID-19 in Latin America: Contrasting phylodynamic inference with epidemiological surveillance.
```

2. We are adding extra tabs to commented lines. e.g.
```diff
+# ZZZ note UPHL-05 and -06 are same patient                    
```

emmahodcroft:
Good catch @kairstenfay ! I don't know off the top of my head how important the quotes are - but I guess in general they probably mostly protect in case there's a random tab in a title or something (at least, most CSV parsers would ignore such a thing if in quotes). Here's how the title shows up in builds at the moment: https://nextstrain.org/ncov/europe_2020-07-31?s=Argentina/PAIS-A0020/2020

Looking at `metadata.tsv` it looks like at this point all quotes have been stripped. Dunno how that plays into how important the quotes are, either.

The extra tabs on comments certainly aren't a killer but they might be a bit confusing. Or might they cause trouble at some point?

kairstenfay:
Hi @emmahodcroft,

> The other thing is that I think probably the comments should be tidied up a bit (this isn't just a problem after using the script - though some of them get more apparent). Small things like, if all comments start `# ` then they're sorted alphabetically after that (at the moment some are `#HongKong` so appear 'out of order'). And we can make things like our faux-'header' (`# strain EPI_ISL key value #comment`) start `##` so that it hopefully then appears just before the real entries. And finally some comments were apparently below whatever they referenced, but since some sorting (present already in the file, so not the script's fault!) they are just floating around at the top and don't make much sense. (I'll either add EPI_ISL, clarify, or remove entirely)

I'm not as familiar with annotations, so I have a few questions here.
* Are you suggesting we enforce a space between `#` and any text after it, so `#HongKong` gets corrected to `# HongKong`? 
* are you suggesting adding a double `##` plus space before any annotation that beings with `strain EPI_ISL key value #comment`? 
* Are there any other suggestions you're making for the clean script in this paragraph you've written?

> Seems to me that a local file, shell or python, would be the best solution if we want to run locally. Perhaps this can be crafted from the existing workflow without too much trouble?

Yes, that's exactly what I've done. I've taken @eharkins's workflow logic and put it into a new script called `./bin/clean-source-data-tsvs`. 

> Just anticipating when I can run again locally but more 'informedly' upside_down_face, I wanted to ask again from above - it's going to be a little hard to see what might have been dropped/modified by the script due to all the ordering changes. 

I have tried to address this by adding a `--sort` option to `./bin/clean-source-data-tsvs`. By default, the script will not sort anything in the annotations & location hierarchy files. The idea here is you can run the script, inspect the diff (via `git diff` or other diffing program), and vet the changes. If they look OK, you can then re-run `./bin/clean-source-data-tsvs --sort`. It still needs a little more testing, but feel free to give it a try and see if the sort/no-sort split addresses your concerns!

> Do we think we're confident that the current ncov-ingest ('new') transform would be throwing up errors for any lines that don't have the right number of columns - so that we're sure we're starting 'clean'? (And don't accidentally lose stuff which we could possibly fix if we catch it now?)

I'll have to do a little more testing on this to know for sure, but can you explain what you mean by staring "clean"? Could you also please elaborate on your concerns about losing stuff?

emmahodcroft:
Hi @kairstenfay 

>I'm not as familiar with annotations, so I have a few questions here.

I'm sorry! I really wasn't clear here - my apologies. Mostly I was just referring to things there that I thought we should fix ourselves just to make the sorting neater for commented things - like since they come out alphabetically we can just try to stick to a format so that it looks tidier after sorting - putting a space after # (since that's what majority did), and using ## to help keep our 'header' near the bottom, just above the real data. But these were all changes I made already in master and pulled in. These could go into the script but given it's comments I don't think it's so critical - just something we can try to do ourselves (and really, had mostly done anyway) to keep things tidy!

> It still needs a little more testing, but feel free to give it a try and see if the sort/no-sort split addresses your concerns!

Oh, that sounds great! I will try to test this tomorrow, then hopefully can use diff to explore any differences :) 

> I'll have to do a little more testing on this to know for sure, but can you explain what you mean by staring "clean"? Could you also please elaborate on your concerns about losing stuff?

Mostly I suppose I'm just wondering if columns might get stripped or added that will make it harder to detect any mistakes that exist right now, in future. I think if we have anything without the right number of columns at the moment we *would* be getting a warning - from my looking at the `gisaid_transform` code today - so I'm actually feeling more relaxed about this now. I guess I was just a little concerned that if we did have something that only has, say, 3 columns now, we could maybe find this more easily (?) now, since it's 'wrong' - but once we have a copy where it has 4 columns due to one being added, it might more easily 'slip through the cracks'. This is rather theoretical - I wasn't sure if this is a real risk or not. But from my look today at `transform` (sorry I didn't think to do that earlier) I know that we'll get warnings for anything that has != 4 columns and isn't a comment, which leaves me thinking this isn't a problem!

kairstenfay:
> I'm sorry! I really wasn't clear here - my apologies. Mostly I was just referring to things there that I thought we should fix ourselves just to make the sorting neater for commented things - like since they come out alphabetically we can just try to stick to a format so that it looks tidier after sorting - putting a space after # (since that's what majority did), and using ## to help keep our 'header' near the bottom, just above the real data. But these were all changes I made already in master and pulled in. These could go into the script but given it's comments I don't think it's so critical - just something we can try to do ourselves (and really, had mostly done anyway) to keep things tidy!

No worries. These are interesting considerations, and it's worth filing an issue if you eventually desire some of these additional comment-parsing behaviors.

> Oh, that sounds great! I will try to test this tomorrow, then hopefully can use diff to explore any differences :)

Great! I look forward to your feedback on whether or not this is a useful tool.

> Mostly I suppose I'm just wondering if columns might get stripped or added that will make it harder to detect any mistakes that exist right now, in future. I think if we have anything without the right number of columns at the moment we _would_ be getting a warning - from my looking at the `gisaid_transform` code today - so I'm actually feeling more relaxed about this now. I guess I was just a little concerned that if we did have something that only has, say, 3 columns now, we could maybe find this more easily (?) now, since it's 'wrong' - but once we have a copy where it has 4 columns due to one being added, it might more easily 'slip through the cracks'. This is rather theoretical - I wasn't sure if this is a real risk or not. But from my look today at `transform` (sorry I didn't think to do that earlier) I know that we would be getting warnings for anything that has != 4 columns and isn't a comment, which leaves me thinking this isn't a problem!

Ah, I see. So you're rightfully concerned about deleting data hanging out in extra tabs when we're trimming number of columns down to 4. You've raised this concern before and it makes sense. So long as we run `transform` before running `clean-tsv` (or its wrapper script, `clean-source-data-tsvs`), then we should see warnings emitted for misshapen annotations.
e.g. This workflow would result in misshapen annotations being flagged in Slack.
1. edit annotations file 
2. run fetch-and-ingest (which calls `transform`) 
3. run `clean-source-data-tsvs`

This, second workflow, however, would not emit warnings about misshapen annotations and would "silently" apply changes (although these changes would be viewable in the diff that must be committed).
1. edit annotations
2. run `clean-source-data-tsvs`
3. run fetch-and-ingest (which calls `transform`)

emmahodcroft:
Ok! Took me a minute to get a good testing strategy but what I ended up doing was stripping all the comments at the top (to avoid diffs due to adding tabs to comments) and removing all quotes from titles (to avoid diffs due to this alone) - only then running the script on this version.

Then running without sorting, things look really good - all I found was that some comments hidden 'in the data' so these had extra tabs added (not concerning, just missed my 'delete all comments'). 

When I ran with `--sort` I did notice a couple things a bit odd - first, the sorting of the 'paper data' at the end seems... odd? Here's the end of the data (it seems to put animal samples last), transitioning to the 'paper data':
image

It's ordered Wuhan, Guangdong, Thailand. I can't really tell if this is sorted, or what's happened. It's not in the order it is in the original file! Probably good to check something sensible is happening?

Finally, what should (ideally?) be the header for the 'paper data' was stuck at the end:
image

If it's tricky to put this at the beginning of the paper data, I think it would also be fine for it to go at the top of the entire file - but it doesn't really make sense to be at the end :)  Also not really sure how it happened!

I'll edit the `master` `gisaid_annotations` to move the comments out of the data and move around a couple of other bits that shouldn't be where they are, which will ease any future testing if needed. However, overall I'm feeling pretty good about this (data being dropped/missed) now 👍 

I guess the only last remaining unknown is maybe to run `gisaid_transform` with one of the cleaned `gisaid_annotations.tsv` data and ensure the lack of quotes & everything doesn't impact how `metadata.tsv` ends up? 

----

>edit annotations file
run fetch-and-ingest (which calls transform)
run clean-source-data-tsvs

I'm happy to propose to use this workflow - we can add this to the docs maybe with a brief note why this is the prescribed order. It doesn't look like we're missing a lot now, though, so that makes me think we have managed to do this ok mostly in the past - hopefully continuing in the future!

emmahodcroft:
I made the changes to `gisaid_annotations` on `master` and have pulled into this branch 👍

eharkins:
> It's ordered Wuhan, Guangdong, Thailand. I can't really tell if this is sorted, or what's happened. It's not in the order it is in the original file! Probably good to check something sensible is happening?

It's sorted by EPI_ISL [using `--sort-col 1`](https://github.com/nextstrain/ncov-ingest/pull/82/files#diff-3043244df2f361961c7bd4c938c6b6e1R43). I did this early on in the workflow - I guess I thought you might want to group annotations this way, to ensure `paper_url` and `title` for the same sequence were always next to each other, but I'm not sure this makes any more sense than using strain or something else. Feel free to change this part to accommodate how you want those annotations sorted (maybe even by `--sort-col 3` ; sort them by the value of the annotation which might be helpful if that is often a unique way to identify the paper; not sure if you like to search by paper or by sequence).


> edit annotations file
run fetch-and-ingest (which calls transform)
run clean-source-data-tsvs

What would this look like in practice as it relates to pushing to github / automation? I was imagining two reasons for doing this sorting: 1. to save time adding annotations so that we can just append them to the end of the file in any order before sorting and 2. to have the version on github be sorted so that annotations could be easily found from the version on github. Right now in terms of automation it seems like we would have

(manual): edit annotations file [and then push to github]
(automated on github): run fetch-and-ingest (which calls transform)
(manual after ingest finishes and we handle any warnings from transform about "misshapen annotations"?): run clean-source-data-tsvs
Then would we push again to github (triggering another ingest / transform) to make sure the version on github was sorted?

I was initially imagining doing it this way:
> This, second workflow, however, would not emit warnings about misshapen annotations and would "silently" apply changes (although these changes would be viewable in the diff that must be committed).
edit annotations
run clean-source-data-tsvs
run fetch-and-ingest (which calls transform)

but if that defeats the purpose of the transform script's warnings, I agree we should do it the other way (or go with a third option that gets the best of both worlds but requires more work on this PR in order to incorporate the warnings about "misshapen" annotations into the tsv cleaning process before it asserts anything about number of columns).

> Finally, what should (ideally?) be the header for the 'paper data' was stuck at the end:

Ah, that's because [we are just sticking any line with the word paper in it in that section](https://github.com/nextstrain/ncov-ingest/pull/82/files#diff-3043244df2f361961c7bd4c938c6b6e1R38-R39), and sorting all of them as if they are paper annotations. Two ways to fix that:
1. be more specific with the grep command by searching for `paper_url` and then adding another grep to stick the paper annotations header before them
2. do the separation of paper stuff in python so that we can have it more specialized.

Thanks @kairstenfay @emmahodcroft for all your work on this, I hope it remains a potential time-saver, improvement for all the trouble!

kairstenfay:
@emmahodcroft how do you ideally want paper annotations sorted? Do you want to keep them separated within the file from the non-paper annotations?

PR #86, opened 2020-08-11
Better info missing locations assert

### LINKED ISSUE #84: Better Info when Assert in `check-locations`
**Context**  
We don't permit any entries to have missing information in `division_exposure` `country_exposure` or `region_exposure`, which is checked in `check-locations` script.

If an entry is blank, an `assert` is triggered, and the `ncov-ingest` pipeline files. Unfortunately, there's little other information provided, so it can be hard to figure out what triggered the assert. 

**Description**   
It would be great to give a list of sequences/epi_isl numbers of any sequences falling foul of the rule, so that it's easier to investigate and fix the problem.

**Examples**  
An example [happened on 27 July](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1595866779487300), with sequence EPI_ISL_498226 - the location was listed only as 'Latvia' with no region. Thus, the `region` was 'Latvia' and `country` and `division` (thus `country_exposure` & `division_exposure`) were empty. This triggered the `assert`. 

However, this was done after making many changes to `gisaid_annotations`, which historically has been the cause of this error. Even though 37 new sequences were also ingested after the changes to `gisaid_annotations`, as the `ncov-ingest` pipeline wasn't completing, it was hard to see what these were, and we assumed it was the `gisaid_annotations` changes that had somehow caused the problem.

Being able to see it was the Latvia/test1/2020 sequence that was causing the problem would have saved us a lot of time trying to find the isssue!

kairstenfay:
Improve the assertion error message when strains have missing location hierarchy data.

Resolves #84. 

Also fixes a bug where we were not enforcing no missing hierarchy data for ['region', 'country', 'division'] (only their "_exposure" counterparts).

### DIFF
--- a/bin/check-locations
+++ b/bin/check-locations
@@ -30,7 +30,7 @@ def validate_metadata(metadata: pd.DataFrame, columns: List[str]):
         except ValueError:
             non_null_columns = columns
 
-        non_null_columns = [ f"{resolution}_exposure" for resolution in non_null_columns ]
+        non_null_columns += [ f"{resolution}_exposure" for resolution in non_null_columns ]
 
         assert metadata[non_null_columns].notnull().all().all(), \
             "There are missing values in the location hierarchy columns " \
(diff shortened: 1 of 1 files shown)


emmahodcroft:
This looks great @kairstenfay ! Thank you for working on this! This should make things much easier in future if we run into this again! 🙌

PR #87, opened 2020-08-12
Alert certain users of new GISAID data from their interested location hierarchy areas

### LINKED ISSUE #79: Alert Certain Users to New Sequences from GISAID
**Context**  
At least two Nextstrain members are currently maintaining more custom builds (Emma maintains USA South-Central/Texas/Louisiana builds, Richard maintains ECDC Europe builds, and both maintain a Swiss build). With decreasing frequency of `ncov` builds, it's become a little harder to notice when new sequences from a country/division/region of interest come in.

It would be cool if there were a way to ping one or more users on Slack about the arrival of new sequences from an area of interest. For example, Emma would find it useful to know whenever we get new sequences from Texas or Louisiana so she can re-run the custom local builds there. Both Emma & Richard would probably find it useful to know when new Swiss sequences appear.

**Description**   
It would be neat if somewhere in 'ingest' it could check against a list of areas of interest and alert the appropriate people when new sequences from these areas pop up. Could be either just tags in a message in #ncov-gisaid-updates channel, or could be private messages - whatever would work easier I guess!
 
Perhaps somewhere in the 'ingest' script there could be a text/JSON/etc file where one can put their slack username and the area they'd like to be alerted to - then any Nextstrain member could easily add or remove alerts as they need/want! For example:

```
username    level     area-name
emmahodcroft    division    Texas
emmahodcroft    division    Louisiana
rneher    country    Switzerland
emmahodcroft    country    Switzerland
...
```

kairstenfay:
Notify certain users of new GISAID data from specified location hierarchy areas of interest. 
Resolves #79.

Open questions:
* What should this config JSON be called?
* What's the best place for this new config JSON to live?
* Where's the best place to document instructions for retrieving a Slack member ID? The README?

### DIFF
--- a/bin/notify-users-on-new-locations
+++ b/bin/notify-users-on-new-locations
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+"""
+Notifies specified users of new data for their interested locations
+"""
+import sys
+import json
+import argparse
+import subprocess
+import pandas as pd
+from pathlib import Path
+from typing import Dict, Iterable, List, Union
+
+
+LOCATION_HIERARCHY_COLUMNS = ['region', 'country', 'division', 'location']
+base = Path(__file__).resolve().parent.parent
+
+
+def create_new_areas_lists(metadata: pd.DataFrame) -> Dict[str, List[str]]:
+    """
+    Returns a dictionary with keys of every location hierarchy resolution and
+    values containing a list of unique area names at each resolution in the
+    given *metadata*.
+    """
+    def unique_resolution_areas(series: Iterable[pd.Series]) -> List[str]:
+        """ Returns a list of unique, non-null items in the given *series*. """
+        return list(
+            pd.concat(series) \
+                .dropna() \
+                .unique()
+        )
+
+    # Locations do not have an associated exposure column
+    resolutions_with_exposure = LOCATION_HIERARCHY_COLUMNS.copy()
+    resolutions_with_exposure.remove('location')
+
+    areas: Dict[str, List] = {}
+
+    for r in resolutions_with_exposure:
+        areas[r] = unique_resolution_areas(
+            [incoming_metadata[r], incoming_metadata[f'{r}_exposure']]
+        )
+
+    # Handle location edge case
+    areas['location'] = unique_resolution_areas([incoming_metadata['location']])
+
+    return areas
+
+
+def alert_user_of_new_strains(entry: Dict[str, Union[str, List[str]]],
+    new_areas: Dict[str, List[str]]):
+    """
+    Using the given configuration *entry*, sends a Slack alert to the configured
+    user if there are new strains in the user's interested location hierarchy
+    areas.
+    """
+    member_id = entry.get('slack_member_id')
+
+    if not member_id:
+        # We can't send a Slack notification to a specified user without this
+        # field
+        return
+
+    alerts: List[str] = []
+
+    for resolution in LOCATION_HIERARCHY_COLUMNS:
+        if not entry.get(resolution):
+            # No alert has been configured at this resolution
+            continue
+
+        alerts += [ area for area in entry[resolution] if area in new_areas[resolution] ]
+
+    if not alerts:
+        return
+
+    alert_text = f"There are new strains from {', '.join(alerts)}"
+
+    subprocess.run([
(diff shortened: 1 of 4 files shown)


bin/notify-users-on-new-locations
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+"""
+Notifies specified parties of new data for their interested locations
```
kairstenfay, line 3: I used the word 'parties' previously in this script but later renamed it to 'users'. I should update the docstring here to reflect that change.

bin/notify-users-on-new-locations
```diff
@@ -0,0 +1,101 @@
...
+import pandas as pd
+from pathlib import Path
+from typing import Dict, Iterable, List, Union
+
+
+LOCATION_HIERARCHY_COLUMNS = ['region', 'country', 'division', 'location']
+base = Path(__file__).resolve().parent.parent
+
+
+def create_new_areas_lists(metadata: pd.DataFrame) -> Dict[str, List]:
```
kairstenfay, line 18: The return type should be `Dict[str, List[str]]`

bin/notify-users-on-new-locations
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Returns a dictionary with keys of every location hierarchy resolution and
+    values containing a list of unique area names for each resolution in the
```
kairstenfay, line 21: "at each resolution" instead of "for each resolution" is probably clearer

bin/notify-users-on-new-locations
```diff
@@ -0,0 +1,101 @@
...
+    new_areas: Dict[str, List[str]]):
+    """
+    Using the given configuration *entry*, sends a Slack alert to the configured
+    user if there are new strains in the user's interested location hierarchy
+    areas.
+    """
+    member_id = entry.get('slack_member_id')
+
+    if not member_id:
+        # We can't send a Slack message without this field
```
kairstenfay, line 59: "We can't send a Slack notification to a specified user..."*

bin/notify-users-on-new-locations
```diff
@@ -0,0 +1,101 @@
...
+        return
+
+    alerts: List[str] = []
+
+    # Ignore the first two keys in the JSON which are the user's name and Slack
+    # member ID
+    for resolution in list(entry)[2:]:
```
kairstenfay, line 66: Perhaps it's better here to loop over the location hierarchy columns instead of the keys of the config object so I don't have this hardcoded assumption of `[2:]` here.

emmahodcroft:
It might be good to merge tomorrow when USA is up, if that's ok - even though it seems low-risk we've had the `ncov-ingest` pipeline fail previously over super random things, and if we didn't write it, it's hard for us to debug! (And if USA is asleep, we are stuck :) ) So good to merge while those who can debug (if needed) are on-hand!

new-sequence-alerts-config.json
```diff
@@ -0,0 +1,8 @@
+[{
```
emmahodcroft, line 1: I think my only comment would just be - if you don't have any entries for `region` etc, can they be left out, or must be there but empty? This would be good to maybe specify explicitly so we don't end up breaking it by removing 'empty' lines... (or make it so it can be removed?)

kairstenfay: Excellent point. I'm going to slightly modify the logic of how we loop over each entry in the config file, and when I do so, I'll make sure that empty `region` etc. can be omitted without throwing an error.

kairstenfay:
> It might be good to merge tomorrow when USA is up, if that's ok - even though it seems low-risk we've had the `ncov-ingest` pipeline fail previously over super random things, and if we didn't write it, it's hard for us to debug! (And if USA is asleep, we are stuck :) ) So good to merge while those who can debug (if needed) are on-hand!

Sure thing! I'll be sure not to merge anything after hours (Seattle time).

eharkins:
> It might be good to merge tomorrow when USA is up 

Sorry, missed this before running today's ingest! Should we plan to merge Monday before ingest and then @kairstenfay can help me debug should any issues arise?

kairstenfay:
> > It might be good to merge tomorrow when USA is up
> 
> Sorry, missed this before running today's ingest! Should we plan to merge Monday before ingest and then @kairstenfay can help me debug should any issues arise?

Yep, I wrote to you in Slack!

Issue #77, opened 2020-07-08
Rewrite transform-gisaid to perform a streaming transform


kairstenfay:
My system's out of memory killer is stopping my local run of `./bin/transform-gisaid`. Because we're using Pandas, we're loading all the GISAID data into memory. Consider rewriting the script to use a streaming transform instead.

huddlej:
It looks like the data are in the correct format to enable pandas's chunking of the JSON as you read it in. For example:

```python
df = pd.read_json("gisaid.ndjson", lines=True, chunksize=1000)
for chunk_df in df:
    print(chunk_df.shape)

# Prints:
# (1000, 19)
# (1000, 19)
# ...
```

Since each chunk is a valid data frame, you may be able to add the chunking in the top-level of the script and keep all the downstream processing functions the same...

tsibley:
One issue with either a chunking or streaming approach here is that the transformation includes both sorting and de-duplicating.

https://github.com/nextstrain/ncov-ingest/blob/2a7fd99f45c4c140d9223a21ba6f1a412fbefbd8/bin/transform-gisaid#L65-L73

It is possible to implement _some_ de-duplication schemes in a streaming/chunking fashion, but ours depends on a sort to choose the preferred record.

huddlej:
What if the de-duplication is decoupled from the other transformations and happens as a first pass through the data? This could look like:

Pass 1: Find and resolve duplicate strains into a set of valid strains/epi ids without sorting (akin to single pass methods to find the max/min values in an array). This might be implemented with a dictionary to track the length and epi id by strain name, but it could be reduced to a set after the pass finishes.

Pass 2: Transform records with standardization function and output metadata and sequences for valid strains epi ids.

A single pass through the compressed JSON takes ~20 seconds and the set of strain/epi ids shouldn't consume much memory compared to the full JSON. Lookups into this set will also be fast for the second pass.

Alternately, this does seem headed in the direction where databases shine, but I get that's not where we want to spend time right now. :)

tsibley:
I agree that sort of two-pass approach could work. There would be some normalization/computation that would have to be applied at the start of both passes, since it effects `strain` and `length`.

[...]

PR #90, opened 2020-08-28
Faster transform-gisaid

kairstenfay:
Processing in transform-gisaid is represented as a pipeline of steps.  Each step either manipulates the data (`Transform`) or filters the data (`Filter`).  At the start of each pipeline is a `DataSource`.

To test this, I processed the same gisaid dataset with both the old script and the new script.  The key differences in the output are:

1. The new script does _not_, by default, sort the sequences before outputting.  It does perform the same deduplication process however.  To sort the sequence data, use the option `--sorted-fasta`.
2. The new script interprets errors in `source-data/gisaid_annotations.tsv` differently than the old script.  It ignores all rows that do not have four columns.  As a result, some annotations are not processed.

Additionally, there is the option `--output-unix-newline`, which forces all output (fasta files and metadata csv files) to use unix newlines.

Fixes https://github.com/nextstrain/ncov-ingest/issues/77

-------
originally opened as #88 by @ttung

### DIFF
--- a/bin/ingest-gisaid
+++ b/bin/ingest-gisaid
@@ -76,9 +76,11 @@ main() {
         aws s3 cp --no-progress "$S3_DST/gisaid.ndjson.gz" - | gunzip -cfq > data/gisaid.ndjson
     fi
 
+    flagged_annotations="$(mktemp -t flagged-annotations-XXXXXX)"
+    trap "rm -f '$flagged_annotations'" EXIT
     ./bin/transform-gisaid data/gisaid.ndjson \
-      --output-metadata data/gisaid/metadata.tsv \
-      --output-fasta data/gisaid/sequences.fasta
+        --output-metadata data/gisaid/metadata.tsv \
+        --output-fasta data/gisaid/sequences.fasta > "$flagged_annotations"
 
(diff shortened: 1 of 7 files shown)


kairstenfay:
Rebased onto master.

bin/ingest-gisaid
```diff
@@ -76,16 +76,19 @@ main() {
...
+        --output-metadata data/gisaid/metadata.tsv \
+        --output-fasta data/gisaid/sequences.fasta > "$flagged_annotations"
 
     ./bin/flag-metadata data/gisaid/metadata.tsv > data/gisaid/flagged_metadata.txt
     ./bin/check-locations data/gisaid/metadata.tsv \
         data/gisaid/location_hierarchy.tsv \
         gisaid_epi_isl
 
     if [[ "$branch" == master ]]; then
+        ./bin/notify-slack --upload "flagged-annotations" < "$flagged_annotations"
```
kairstenfay, line 91: The printed warnings in Tony's `transform-gisaid` are now sent to Slack, but only when on the master branch. I just wanted to check that this is the desirable behavior. @eharkins

eharkins: This makes sense to me!

PR #92, opened 2020-09-01
Line endings

### LINKED ISSUE #91: transform-gisaid produces output TSVs with CRLF line endings
**Current Behavior**  
The latest `transform-gisaid` script, introduced by #90, introduced a bug that produces all output TSVs (metadata and additional info) with Windows (CRLF) line endings (i.e. newlines) on all operating systems (Mac, Linux) by default instead of relying on the user's OS. Right now, a [temporary workaround](https://github.com/nextstrain/ncov-ingest/commit/638af88b81d2e2fa9cbfc6ae49896a3eb9b98c18) is implemented in the `ingest-GISAID` workflow to always invoke the `--output-unix-newline` option when running `transform-gisaid`.

**Expected behavior**  
If the user invokes `transform-gisaid` with the `--output-unix-newline` option, then all line endings are in Unix style (works as expected).
Otherwise, the script outputs newlines using the user's operating system (does not work as expected and instead produces CRLF line endings).

**How to reproduce**  
Steps to reproduce the current behavior:  
1. Run `transform-gisaid` without the `--output-unix-newline` option.
2. Run `file metadata.tsv` or `file additional_info.tsv`. The output is 
```
UTF-8 Unicode text, with very long lines, with CRLF line terminators
```
instead of the expected 
```
UTF-8 Unicode text, with very long lines
```

kairstenfay:
Uses a user's OS newline character(s) by default when writing output files.

### Related issue(s)  
Fixes #91 

This works on Ubuntu, but I need Windows and Mac testers. Can I count on your help, @emmahodcroft and @eharkins? Ping me here or on Slack if you have any questions, thank you!

### DIFF
--- a/bin/transform-gisaid
+++ b/bin/transform-gisaid
@@ -2,6 +2,7 @@
 """
 Parse the GISAID NDJSON load into a metadata tsv and a FASTA file.
 """
+import os
 import argparse
 import csv
 import sys
@@ -82,7 +83,7 @@ if __name__ == '__main__':
         dest="newline",
         action="store_const",
         const="\n",
-        default=None,
+        default=os.linesep,
         help="When specified, always use unix newlines in output files."
     )
     args = parser.parse_args()
@@ -151,9 +152,7 @@ if __name__ == '__main__':
     with open(args.output_fasta, "wt", newline=args.newline) as fasta_fh:
         with open(args.output_additional_info, "wt", newline="") as additional_info_fh, \
              open(args.output_metadata, "wt", newline="") as metadata_fh:
-            dict_writer_kwargs = {}
-            if args.newline is not None:
-                dict_writer_kwargs['lineterminator'] = args.newline
+            dict_writer_kwargs = {'lineterminator': args.newline}
 
             # set up the CSV output files
             additional_info_csv = csv.DictWriter(


kairstenfay:
@ttung -- If you have some time, would you please review this PR? I believe I've fixed the bug we bumped into where when `--output-unix-newline` wasn't invoked, python's `csv.DictWriter` fell back to a default lineterminator of `\r\n` (regardless of OS).  
I also want to make sure I didn't miss any context in my second commit by removing the hardcoding of fasta newlines as `\n`. You likely have more context here than me!

bin/transform-gisaid
```diff
@@ -151,9 +152,7 @@ if __name__ == '__main__':
     with open(args.output_fasta, "wt", newline=args.newline) as fasta_fh:
         with open(args.output_additional_info, "wt", newline="") as additional_info_fh, \
              open(args.output_metadata, "wt", newline="") as metadata_fh:
-            dict_writer_kwargs = {}
-            if args.newline is not None:
-                dict_writer_kwargs['lineterminator'] = args.newline
+            dict_writer_kwargs = {'lineterminator': args.newline}
```
ttung, line 155: I actually noticed this earlier.  I believe python's stated reason for doing this is because Excel does not like `.csv` files with just `\n`.  If this is not a concern, we can do this.  If it is a concern, we should go with python's default behavior.

kairstenfay: Good to know. @emmahodcroft & @eharkins -- if either of you are Excel users, when testing, could you check if you can open the output TSVs from this PR in Excel?

emmahodcroft: So I never normally open up any of the files in Excel. However I've just opened up a `metadata.tsv` with `LF` endings (this is Unix I believe?) and it opened up fine.

emmahodcroft:
This seems to work for me on Windows (in WSL, mind you! - but using the original PR before the argument option was added, I was getting Windows endings). The `metadata.tsv` file generated by running this script has `LF` endings (I believe these are Unix).

emmahodcroft (approved):
This seems to work for me on Windows (WSL), and produces metadata.tsv with LF line endings

bin/transform-gisaid
```diff
@@ -193,5 +192,5 @@ if __name__ == '__main__':
                         | SequenceLengthFilter(15000)
                         | LineNumberFilter(line_numbers)
                 ):
-                    fasta_fh.write(f">{entry['strain']}\n")
-                    fasta_fh.write(f"{entry['sequence']}\n")
+                    fasta_fh.write(f">{entry['strain']}{args.newline}")
+                    fasta_fh.write(f"{entry['sequence']}{args.newline}")
```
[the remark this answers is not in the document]
kairstenfay: Thanks. I've dropped this commit.

bin/transform-gisaid
eharkins: I never open with Excel either when running the ingest.

kairstenfay:
I just pushed a fixup with a minor reword in the commit message. This is ready to merge tomorrow!

PR #96, opened 2020-09-16
Fix how we process comment lines

### LINKED ISSUE #95: Commented Annotations with 4 Columns are Still Applied
**Current Behavior**  
As spotted by @eharkins - currently it seems rows that are commented but have 4 columns are applied, rather than ignored.

In [this commit](https://github.com/nextstrain/ncov-ingest/commit/69cde21eb4a2687a0e3e4bc31b4ca0b7970d15d0), lines 698-707, Eli included some commented information for reference but that shouldn't be processed. 

Example:
```
# USA/OK-ADDL-03/2020	EPI_ISL_535362	location	City Suburb
# USA/OK-ADDL-04/2020	EPI_ISL_535363	location	City Suburb
# USA/PR-B35B/2020	EPI_ISL_534693	location	Ausilio Mutuo # Hospital in San Juan
# USA/PR-B3SB/2020	EPI_ISL_534694	location	Ausilio Mutuo # Hospital in San Juan
```

However, after pushing this update, we got notified that these locations had been applied, despite being commented:
image

**Expected behavior**  
Any line that starts with `#` should be fully ignored, even if it has the right number of columns.

**Possible solution**  
This seems to stem from [this section](https://github.com/nextstrain/ncov-ingest/blob/master/bin/transform-gisaid#L97) of the code in `gisaid_transform` where we check if a line has 4 columns -- only if it _does not_ then do we see if it is commented. If it has 4 columns, it seems to go ahead and try to process it, even if it starts with `#`. 

We should change this so that it first checks to see if the first character (minus whitespace) is `#` and ignore the whole line if so (regardless of number of columns).

**Temporary solution**
Ensure that commented columns we don't want to be applied do *not* have 4 columns.

ttung:
We were only detecting comment lines if it didn't resolve into four columns.  This changes the ordering -- first detect whether it is a comment, and only then do we continue processing.

### Related issue(s)  

Fixes #95

### DIFF
--- a/bin/transform-gisaid
+++ b/bin/transform-gisaid
@@ -94,10 +94,10 @@ if __name__ == '__main__':
         with open(args.annotations, "r") as gisaid_fh:
             csvreader = csv.reader(gisaid_fh, delimiter='\t')
             for row in csvreader:
-                if len(row) != 4:
-                    # ensure that it's a comment
-                    if row[0].lstrip()[0] != '#':
-                        print("WARNING: couldn't decode annotation line " + "\t".join(row))
+                if row[0].lstrip()[0] == '#':
+                    continue
+                elif len(row) != 4:
+                    print("WARNING: couldn't decode annotation line " + "\t".join(row))
                     continue
                 strain, epi_isl, key, value = row
                 annotations[epi_isl].append((


emmahodcroft:
This looks good to me! @eharkins would you mind having a look? If you think all is good, happy for you to merge :)
Thanks for quick work on this @ttung !

eharkins (approved):
Looks good and tested it fixes the issue as experienced today :) thanks @ttung

PR #89, opened 2020-08-26
Local GISAID ingest script

ivan-aksamentov:
### Description of proposed changes    

This adds local, offline script for GISAID ingest.

### Related issue(s)  

This script is meant to be a temporary replacement for the currently non-functional ingest Github Action.

### Testing

The workflow should be roughly as follows:

1 .Download input files from S3 (these are the "rolling" results of the previous ingest):

```
./bin/local-ingest-gisaid --download-inputs
```

The following files will be downloaded:

```
data/gisaid/inputs/metadata.tsv
data/gisaid/inputs/additional_info.tsv
```

2. Download fresh GISAID data

Create a file `.env` in the directory one level above the project. For example, if your project is in  `~/work/ncov-ingest`, make a file `~/work/.env`. Define the following variables in it:

```
GISAID_URL=
GISAID_USERNAME=
GISAID_PASSWORD=
```
(make sure you don't loose or accidentally commit this sensitive information anywhere)

Run:

```
./bin/local-ingest-gisaid --download-gisaid
```

The following files will appear:

```
data/gisaid/tmp/gisaid.ndjson.bz2
data/gisaid/inputs/gisaid.ndjson
```

2.5. Optionally, remove or move the previous outputs:

```
rm -rf data/gisaid/outputs
# or
mv data/gisaid/outputs{,_}

```

3. Run ingest

```
./bin/local-ingest-gisaid --ingest
```

Note: that this part of the script does not use network or any cloud storage. It is currently does not send the Slack notifications either.

Note: additionally do downloaded inputs, files in `source-data/` are also used during ingest.

The results will be produced in `data/gisaid/outputs/`:

```
additional-info-changes.txt
additional_info.tsv
flagged_metadata.txt
location-hierarchy-additions.tsv
location_hierarchy.tsv
metadata-additions.tsv
metadata-changes.txt
metadata.tsv
sequences.fasta

```

4. Perform the usual cleanup and build shepherding. You don't have to commit and push changes to files in `source-data/` to GitHub for the local ingest to be operational. However, it is a good idea to share these changes with the world when you are done.

5. Re-run ingest (3) with cleaned data if needed. You can re-run as many times as you want. Your changes stay on your machine.

6. Upload the new results to S3 bucket:

```
./bin/local-ingest-gisaid --upload-outputs
```

The following files will be uploaded:

```
metadata.tsv
additional_info.tsv
flagged_metadata.txt
sequences.fasta
```

Now the ncov build can be triggered as usual.

[...]

PR #99, opened 2020-09-29
Add nextstrain clades to metadata (version for running locally)

ivan-aksamentov:

During GISAID ingest, it is proposed to produce a version of `metadata.tsv` file which contains an additional column "Nextstrain_clade", containing clade deduced using Nextclade CLI tool.

Thus, the "nextmeta" published on GISAID will contain not only Pangolin lineages and GISAID clades, but also Nextstrain clades.

---

<sub>
Work in progress: Currently only has implementation for local runs.
</sub>

---

<sub>
TODO: run for a few days in addition to the current vanilla ingest and compare the results. Port to GitHub-action-based script.
</sub>

---

In order to make this happen, this PR introduces 2 distinct tasks:

##### One-time preparation task ("full run"):

Ensures that all currently existing GISAID sequences (~100k) are processed and have a clade assigned. It's too compute-intensive for GitHub actions and must be done locally, on premises or on a high-performance cloud instance. Runs in a few hours on a laptop, in background or overnight.

The same task can be used for periodic mass-synchronization of clades, to account for metadata and sequence changes on GISAID.

Implementation details:
 - have fresh `sequences.fasta`, `metadata.tsv`
 - split `sequences.fasta` into batches: `sequences.fasta.batch-xxxxx.tsv`
 - for each batch (in parallel): run `nextclade.js`, get `sequences.fasta.batch-xxxxx.tsv`, containing clades
 - concatenate results into `nextclade.clades.tsv`
 - join `metadata.tsv` and `nextclade.clades.tsv`, such that `metadata2.tsv` has an additional column "Nextstrain_clade"

##### Daily task:

Runs as a part of daily ingest. Takes only sequences that have no clades, assigns clades, merges them into the new metadata. Should take no more than a minute or a few for a usual daily batch of sequences.

Implementation details:

 - during ingest, have fresh `metadata.tsv`, `sequences.fasta`, and old `nextclade.clades.tsv`
 - filter `sequences.fasta`: get a list of sequences that are in `sequences.fasta`, but are not in `nextclade.clades.tsv`, save them to `nextclade.sequences.fasta`
 - run nextclade on `nextclade.sequences.fasta`, get `nextclade.clades.new.tsv`
 - add `nextclade.clades.new.tsv` to `nextclade.clades.tsv`
 - join `metadata.tsv` and `nextclade.clades.tsv`
 - TODO: the final implementation for GitHub actions should persist `nextclade.clades.tsv` on S3

##### Additional considerations and decisions:

 - Name of the new column. Currently "Nextstrain_clade".
 
 - How and how often to run "full runs".

 - Whether we need to implement matching by GISAID ID. Currently matching of clade to metadata is done by sequence name. These names might change on GISAID. A more reliable mapping can be implemented by using GISAID-specific stable identifier. Nextclade has no access to metadata and only operates on sequence names. So this might not be currently possible to do reliably without processing sequences one by one, which might be very slow.
  Additionally, matching by name may produce inaccurate results if names are not unique and sequences with the same name are supposed to be in different clades.

 - Clades is only a tiny portion of Nextclade's output. We may consider more things to be saved, either for internal use or for publishing.

### Related issue(s)  
N/A

### Testing

This is based on local ingest branch (see #89)

 - setup local ingest environment as described in #89 

 - the script currently assumes that the `master` version of `nextclade.js` is available globally. That's the only version that has the new `--output-tsv-clades-only` flag. You can build it as described in [README](https://github.com/nextstrain/nextclade/blob/master/packages/cli/README.md#build) and it appears in packages/cli/dist. Then I just symlinked it to one of my directories in `PATH` (`~/bin`).
   It will be available on npm after we finalize this business here, so that simple `npm install -g @neherlab/nextclade` will do then.

 - for "full run":

    ```bash
    nice -19 ionice -c3 ./bin/gisaid-get-all-clades
    ```
    (scales with the number of available threads)

 - observe `data/gisaid/outputs/metadata2.tsv` which contains clades for all sequences (excluding the ones that failed to align during nextclade step)

 - observe `data/gisaid/outputs/nextclade.clades.tsv` with clade mapping (and no other metadata)

 - move `data/gisaid/outputs/nextclade.clades.tsv` to `data/gisaid/inputs/nextclade.clades.tsv`
   (simulating the transition to the next day - inputs of the today are the outputs of yesterday)

 - for daily ingest:
   run as described in #89 (the new steps are added to the `--ingest` part)

 - observe `data/gisaid/outputs/metadata2.tsv` which contains clades for all sequences, including today's

🙏

@rneher @emmahodcroft Could you please expand on motivation and other considerations for this feature?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ ehthumbs.db
 Thumbs.db
 
 .env
+tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ ncov/
 Icon?
 ehthumbs.db
 Thumbs.db
+
+.env
--- a/bin/filter-fasta
+++ b/bin/filter-fasta
@@ -31,16 +31,13 @@ def main():
     args = parse_args()
 
     tsv = pd.read_csv(args.input_tsv, sep="\t")
-    tsv_ids = list(tsv['seqName'])
+    tsv_ids = set(tsv['seqName'])
 
-    result = []
     with open(args.input_fasta) as f_input:
-        for seq in SeqIO.parse(f_input, "fasta"):
-            if seq.id not in tsv_ids:
-                result.append(seq)
-
-    with open(args.output_fasta, "w") as f_output:
-        SeqIO.write(result, f_output, "fasta")
+        with open(args.output_fasta, "w") as f_output:
+            for seq in SeqIO.parse(f_input, "fasta"):
+                if seq.id not in tsv_ids:
+                    SeqIO.write(seq, f_output, "fasta")
 
 
 if __name__ == '__main__':
--- a/bin/filter-fasta
+++ b/bin/filter-fasta
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+import argparse
+
+import pandas as pd
+from Bio import SeqIO
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="\
+Produces a FASTA FILE containing sequences present in the input FASTA file, \
+but missing the input TSV file. This is useful in order to find which sequences require additional treatment. \
+")
+    parser.add_argument(
+        "--input_fasta",
+        required=True,
+        help="Path to the input FASTA file.")
+    parser.add_argument(
+        "--input_tsv",
+        required=True,
+        help="Path to the input TSV file."
+    )
+    parser.add_argument(
+        "--output_fasta",
+        required=True,
+        help="Path to the output FASTA file."
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    tsv = pd.read_csv(args.input_tsv, sep="\t")
+    tsv_ids = list(tsv['seqName'])
+
+    result = []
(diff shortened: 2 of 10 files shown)


rneher:
Looks great! 

As for the basic motivation: Many users use the nextmeta data file we provide via gisaid for their analysis. To make it easier for them to use our clades nomenclature, we want our clades to be part of this file. 

 * the linking with sequence names vs EPI_ISL_IDs would not be an issue if they isolate IDs were used as sequence idenitifiers in the fasta files. This would require having the linking of isolate ids and names at hand when splitting the fasta. 
 * `Nextstrain_clade` is a good name and mirrors `GISAID_clade` and `pangolin_lineage`

ivan-aksamentov:
There was a bug - I forgot to wait for the last few processes to finish before proceeding with joining the results. This is a race condition, which may lead to incomplete results (last few batches were still running, so they don't end up in the joined results), or tmp directory might have been deleted by the time the batches complete, which would cause crash. Fixed now.

PR #100, opened 2020-09-30
Add nextstrain clades to metadata

ivan-aksamentov:
This is a port of #99, adapted for the current master branch.
Please refer to #99 for the detailed description.

Now that we can run the new functionality locally, we can try it on GitHub Actions.

Main differences from #99 are:
 - the old clade file is fetched from S3
 - the new clade file is uploaded to S3

For the very first run, I am going to manually upload the clades file with clades precalculated for all existing sequences as of today. We ran the "full run" workflow on our cluster in order to obtain it.

Proposed course of action:

- [x] review
- [x] see if it works on a branch, run in parallel to the daily routine for a day or two
- [x] merge and see if something breaks

Questions:
 - How `metadata.tsv` ends up on GISAID (becoming "nextmeta")?

### Related issue(s)  
N/A

### Testing

It should've triggered the GitHub Action. And we will shortly see if it works. If it does I will just check the output metadata file manually.

:+1:


rneher:
Thanks, @ivan-aksamentov! @kairstenfay do you have time to take a look at it? This should make our clade definitions available to many more people and paves the way for using other useful output of nextclade.

emmahodcroft:
I'm just catching up on this, so apologies! I think this is a great and really important move to make it easier to use Nextstrain clades!

One thing I wanted to flag though, is that while I know it'll take an extra step, I think we do need to be long-term storing the clade definitions by EPI_ISL rather than strain. Strain name changes are not unusual (85 changed today), so we will lose clade definitions if we store in this way, as they won't be able to map onto the 'new' strain name that comes through.

I think we could implement this pretty easily by adding one extra step - in either of two places - either after generating the 'strain,clade' file, we create a copy of this, but using the `metadata.tsv` file to replace `strain` with `gisaid_epi_isl` - or after appending to `metadata.tsv` we then pull this data back out again - taking `gisaid_epi_isl` and `nextstrain_clade` columns to store.

I hope that makes sense, happy to chat more if needed - but we will otherwise lose clade definitions we've made in the past due to strain name changes!

PR #109, opened 2020-12-05
Updates to genbank_accession lines (1540 new, seqs downloaded 2020-12-04)

AngieHinrichs:

This time, in addition to adding new mappings, I've also done a little cleanup of some cases where the same sequence was apparently submitted multiple times to GISAID and/or GenBank:

- In some cases, the same GenBank accession was assigned to more than one EPI ID with identical sequence and similar name (apparent duplicate submissions to GISAID).  I removed 9 duplicate mappings to make the mapping 1-to-1.

- In other cases, there are multiple GenBank accessions that could be assigned to one EPI ID (apparent duplicate submissions to GenBank).  Different versions of my pipeline have arbitrarily chosen different GenBank accessions over time.  I changed 219 GenBank accessions to match the latest output of my pipeline.

- There's also one case of duplicate submission to both GenBank and GISAID, and I straightened that out to make a 1-to-1 mapping.

- 2 apparent duplicate submissions to GISAID had their newer EPI IDs retracted (the epicov.org search for those accessions says "EPI_ISL_... is currently not available at the Submitter's request"), I removed those as well.

Thanks!   :)


AngieHinrichs:
I also just found that 4 sequences have been removed from GenBank by the submitter's request, so I removed the corresponding 4 genbank_accession lines.

emmahodcroft:
Thank you once again Angie for this great work and for the detailed explanations! It looks great and I'll merge now! Thanks for keeping up with this, I think it's so incredibly useful!

AngieHinrichs:
Thank you Emma, all the tools that you and the rest of the Nextstrain team are providing are absolutely vital to SARS-CoV-2 genomic research, so I'm thrilled I can help in this small way.  :)

Issue #103, opened 2020-10-20
Slightly different strain names in sequences.fasta vs metadata


AngieHinrichs:
**Current Behavior**  
While the vast majority of today's 150k strain names match between sequences.fasta and metadata downloaded from GISAID, there are 16 sequences with slightly different names in sequences.fasta compared to metadata.

**Expected behavior**  
Ideally 100% of the names could match, so that the metadata for every sequence in sequences.fasta could be automatically associated with the sequence.

**How to reproduce**  

1. Download nextmeta and nextfasta from GISAID (currently metadata_2020-10-20_07-16.tsv.gz and sequences_2020-10-20_07-17.fasta.gz)

2.   Extract and compare strain names in a command shell:

```
gunzip -c metadata_2020-10-20_07-16.tsv.gz | tail -n +2 | cut -f 1 | sort > metadata.names
gunzip -c sequences_2020-10-20_07-17.fasta.gz | grep ^\> | sed -e 's/^>//' | sort > sequences.names
comm -3 metadata.names sequences.names
```

3. Differences (first column is name that appears in metadata_2020-10-20_07-16.tsv.gz, second column is name that appears in sequences_2020-10-20_07-17.fasta.gz:

```
	/Italy/SIC-IZSBM169-57/2020
	/Italy/SIC-IZSBM169-76/2020
	Bangladesh/BCSIR-NILMRC-125/2020
	Bangladesh/BCSIR-NILMRC-131/2020
	Bangladesh/BCSIR-NILMRC-139/2020
	Bangladesh/BCSIR-NILMRC-145/2020
	Bangladesh/BCSIR-NILMRC-146/2020
	Bangladesh/BCSIR-NILMRC-150/2020
	Bangladesh/BCSIR-NILMRC-151/2020
	Bangladesh/BCSIR-NILMRC-152/2020
	Bangladesh/BCSIR-NILMRC-153/2020
Bangladesh/BCSIR-NILMRC_125/2020
Bangladesh/BCSIR-NILMRC_131/2020
Bangladesh/BCSIR-NILMRC_139/2020
Bangladesh/BCSIR-NILMRC_145/2020
Bangladesh/BCSIR-NILMRC_146/2020
Bangladesh/BCSIR-NILMRC_150/2020
Bangladesh/BCSIR-NILMRC_151/2020
Bangladesh/BCSIR-NILMRC_152/2020
Bangladesh/BCSIR-NILMRC_153/2020
[...]

PR #110, opened 2020-12-10
bin/transform-gisaid address #103 strain name bug

eharkins:
this is an attempt to fix a #103 
where the strain names output
by bin/transform were not consistent
across metadata and fasta output files.
Since this was only an issue when
--sorted-fasta was not passed, this
addresses the issue by adding logic
to use strain names from the transformed
metadata list (to match the metadata output)
when writing out an unsorted fasta.

### Testing
Tested locally with bin/local-ingest-gisaid --transform which resulted in a clean diff (instructions for diff to illustrate the bug are in #103).
Also tested memory usage (only once) using `/usr/bin/time -l pipenv run bin/local-ingest-gisaid --transform` which showed  only a ~28MB increase when running transform (as in bin/local-ingest-gisaid) with this commit compared to running without. This seems to avoid the larger scale (~5.5GB) increase of using --sorted-fasta.

### DIFF
--- a/bin/transform-gisaid
+++ b/bin/transform-gisaid
@@ -174,6 +174,7 @@ if __name__ == '__main__':
             )
             metadata_csv.writeheader()
 
+            updated_strain_names_by_line_no = {}
             for entry in sorted_metadata:
                 if entry[LINE_NUMBER_KEY] in line_numbers:
                     additional_info_csv.writerow(entry)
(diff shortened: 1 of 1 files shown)


eharkins:
The check here fails because it is trying to access a file on aws according to the name of this branch and that file only exists without the branch name prefix. It should not fail in that way when we merge.

kairstenfay (approved):

PR #112, opened 2020-12-18
Updates to genbank_accession lines (529 new, seqs downloaded 2020-12-18)

AngieHinrichs:

Last update in 2020!  Just 529 new genbank_accessions this time.

Thanks & happy holidays!


AngieHinrichs:
529 seemed awfully small and it was -- I missed some new sequences.  Closing this for now, should be able to fix it soon.

AngieHinrichs:
I didn't have quite all of the latest sequences the first time around -- now there are 1418 new mappings which is more in line with what I would expect after two weeks.  Thanks again!

emmahodcroft:
Thank you Angie!! As always your work on this is always so appreciated!

AngieHinrichs:
Thank you Emma!  :)

PR #115, opened 2021-01-06
Upgrade Nextclade to 0.11.1 with new clades

ivan-aksamentov:

This brings the new upcoming clades, as described in 
https://virological.org/t/updated-nextstain-sars-cov-2-clade-naming-strategy/581

The Nextclade default tree was updated here: https://github.com/nextstrain/nextclade/pull/293

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,6 @@
   "license": "MIT",
   "private": true,
   "dependencies": {
-    "@neherlab/nextclade": "0.10.0"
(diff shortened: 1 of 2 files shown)


ivan-aksamentov:
I cancelled the triggered workflow, as the usual incremental run will produce incorrect results.

`nextclade.tsv` should be recomputed before proceeding.

PR #118, opened 2021-01-17
DRAFT: Snakefile

rneher:
This adds a `Snakefile` to the ncov-ingest repo and makes some minor changes to a few scripts.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -5,6 +5,8 @@ ruleorder: unpack_gisaid>fetch_gisaid_from_S3
 ruleorder: concat_clades>nextclade_all
 localrules: download_inputs, download_gisaid, push_gisaid_to_S3, post_to_slack, fetch_gisaid_from_S3, rerun_clades, upload_gisaid_to_S3
 
+wildcard_constraints:
+    data_source = "gisaid|genbank"
 
 rule download_inputs:
     output:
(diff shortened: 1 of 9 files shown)


Snakefile
```diff
@@ -0,0 +1,260 @@
...
+        '''
+
+
+##########################################################################
+onerror:
+    shell("./bin/notify-slack 'ncov-ingest failed'")
+
+
+##########################################################################
+## target rules
```
eharkins, line 224: @rneher If remove these rules (locally) and this one:
https://github.com/nextstrain/ncov-ingest/pull/118/files#diff-47959dfd378b3cd1d39b5515418ee8e4444ab7a6036d5197a5bea82814f928a3R41

does that seem like a good way to be able to test running this on AWS since it won't mess with anything official by not uploading anything new to s3?

Seems like (?) we just need to make sure it has access to inputs and dependencies and then we can just do `nextstrain build --aws-batch`

rneher: yes. or you wrap all pushed to S3 in an `echo "..."`. Then it prints them but doesn't do anything. These would be the rules `upload_gisaid_to_S3` and `push_gisaid_to_S3` and maybe `post_to_slack`. 

These rule names are a little convoluted, mostly since I didn't really know what would come further down the file...

better defining what needs updating after annotations get updated is also necessary. For example by splitting the `transform` into something that turns the gisaid json into metadata and sequences and another step that applies the annotations.

joverlee521:

PR #121, opened 2021-01-21
Support for gzipped fasta files throughout workflow

ttung:
1. Upload and delta change scripts understand .gz files.  Checksums are still performed on uncompressed data.  This works out well, because we compress at most once, and uncompressing is cheap.
2. `bin/transform-gisaid` will inspect the input to see if it's gzipped.  If so, it unzips it.
3. `bin/transform-gisaid` will use the output filename to decide whether or not to compress it on the fly.

### Testing
Launched a build where we keep most files compressed.  We had an earlier version of the workflow without the nextclade integration, but when rebased, it'll likely be problematic.  However, this workflow can still operate on uncompressed data, so it is backwards compatible.

### DIFF
--- a/bin/transform-gisaid
+++ b/bin/transform-gisaid
@@ -2,11 +2,16 @@
 """
 Parse the GISAID NDJSON load into a metadata tsv and a FASTA file.
 """
-import os
 import argparse
+import contextlib
 import csv
+import gzip
+import io
+import os
 import sys
 from pathlib import Path
+
+from magic import from_buffer
 
 sys.path.insert(0, str(Path(__file__).parent.parent / "lib"))
 from utils.transform import (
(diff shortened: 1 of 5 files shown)


tsibley:
Thanks, @ttung! Is this working well for you folks in your own ingests? Did you notice a meaningful time increase due to compression?

Background for the rest of the team/posterity: Tony's change was born out of disk-space issues on GitHub Actions. The Nextstrain team has since moved our ingest to run on AWS Batch (via `nextstrain build` infrastructure) after being triggered by GitHub Actions.

On AWS Batch we're not meaningfully constrained on disk anymore, so I'm not sure trade of disk usage for CPU/wall clock time that compression makes is worth it? That said, this PR doesn't require use of compression, just adds support for it, so we could still accept the PR but not make the compression trade.

ttung:
It's not meaningfully different because we are compressing during upload-to-s3 anyway.  Decompressing the .gz file takes ~4s on my laptop.

Yes, this does not require you to use compression.

rneher:
@WandrilleD Do you see any reason why this shouldn't be merged?

bin/transform-gisaid
```diff
@@ -50,7 +76,7 @@ if __name__ == '__main__':
         formatter_class=argparse.RawTextHelpFormatter
     )
     parser.add_argument("gisaid_data",
-        default="s3://nextstrain-ncov-private/gisaid.ndjson.gz",
+        default="data/gisaid.ndjson.gz",
```
WandrilleD, line 79: this looks like a local adaptation. Any reason why we should not use the s3 file there ?

WandrilleD (changes requested):
I have doubt about the new suggested default value for the gisaid json file name in transform-gisaid. Aside from that everything seems file, especially since it also support non-gzipped files.

PR #124, opened 2021-01-29
Updates to genbank_accession lines (5974 new, 1760 removed, seqs downloaded 2021-01-29)

AngieHinrichs:

 * removed 1760 for which sequences were either updated in GISAID but not GenBank or were withdrawn from GISAID
 * added 5794 from sequences downloaded 2021-01-29.

Thanks so much!  :)

Angie


emmahodcroft:
Thank you as always for your tireless work on this @AngieHinrichs ! 👍

tsibley:
@AngieHinrichs Thanks again for this curation work!

> removed 1760 for which sequences were either updated in GISAID but not GenBank or were withdrawn from GISAID

Thoughts on ways to maintain some linkage here, even if the sequences are no longer identical?  For example, do you think it'd be reasonable in cases like this to revert to the non-versioned GenBank accession?

AngieHinrichs:
Good question @tsibley.  Since GISAID sequences aren't versioned, I think it would be best if the submitters of those sequences could be persuaded to update the sequences in GenBank too.  I will attempt to contact them.

In other cases, where GenBank was updated and not GISAID, the GenBank accession version numbers were incremented, and I have kept the mappings to older GenBank versions.  However, in this case, the submitters updated GISAID and not GenBank.  The sequences are no longer identical, and GISAID has a claim on whatever base differences are unique to the updated sequence in GISAID.

Version numbers can cause their own minor headaches, but for issues like this, and for reproduceability of analyses, it would be nice if GISAID versioned their sequences as GenBank.  I don't think that's going to happen, though.

@dmaccannell maintains https://github.com/CDCgov/SARS-CoV-2_Sequencing/tree/master/files/ with all versions of my mappings between EPI IDs and GenBank (and COG-UK sequences, which are separately available from COG-UK), so the removed mappings can still be found in older files -- e.g. the 1760 removed mappings are still in the file epiToPublicAndDate.2021-01-15.gz.

tsibley:
@AngieHinrichs Nod, that makes sense. Thanks for the explanation!

The pointer to the full mapping files are useful. We may want to use them as part of this repo in the future. :-)

Issue #125, opened 2021-02-04
Add GISAID accessions to the GenBank metadata file


tsibley:
The GISAID metadata references GenBank accessions, but the GenBank metadata doesn't have any GISAID accessions. Refactor how these accessions are tracked and cross-referenced in `transform-gisaid` and `transform-genbank` so they share a _source-data/_ file separate from annotations.

See also https://github.com/nextstrain/ncov-ingest/pull/124#issuecomment-772940020 re: versioned vs. non-versioned GenBank accessions.

tsibley:
@AngieHinrichs I'm going to lift over your wonderful annotations to our GenBank metadata too so it can point back to GISAID. There will be some upcoming code/data changes for that in this repo, which may impact your update process? I wanted to flag this for you now so you're aware. Happy to discuss, and I can flag again when there's code to show.

AngieHinrichs:
@tsibley delighted to hear that you're working on the GenBank metadata side!  I have my own [cobbled-together flow](https://github.com/ucscGenomeBrowser/kent/blob/master/src/hg/utils/otto/sarscov2phylo/getNcbi.sh) to integrate NCBI Virus API download files with a couple BioSample attributes from eutils, and locally computed Nextclade and Pangolin, but I'll bet ncov-ingest results will be better!  :)  Are the GenBank metadata results going to be available for public download?  I would like to take a look & compare notes.  Thanks!

And no worries about my update process, I imagine it will be no trouble to swizzle columns to whatever new form is required.  :)

tsibley:
@AngieHinrichs

> but I'll bet ncov-ingest results will be better! :)

Well, we'll see! Right now it's not getting as much usage, but we do want to improve it.

> Are the GenBank metadata results going to be available for public download?

Yes! You can take a peek at the current outputs here:

* https://data.nextstrain.org/ncov-ingest/genbank.ndjson.gz
* https://data.nextstrain.org/ncov-ingest/genbank_metadata.tsv.gz
* https://data.nextstrain.org/ncov-ingest/genbank_sequences.fasta.gz

but these specific URLs will soon vanish when we move the files around before documenting/distributing more widely. So please don't rely on these quite yet.

> And no worries about my update process, I imagine it will be no trouble to swizzle columns to whatever new form is required. :)

:grin:

tsibley:
[...]

PR #127, opened 2021-02-04
Shared accession mapping

tsibley:
See commit messages for details.

Resolves https://github.com/nextstrain/ncov-ingest/issues/125.


tsibley:
This worked flawlessly on the branch CI, so I've merged it to avoid data drift.

PR #130, opened 2021-02-12
Updates to accessions.tsv (5619 new, 1 removed, seqs downloaded 2021-02-12)

AngieHinrichs:
Another bi-weekly update of GenBank <-> GISAID mappings -- now in accessions.tsv.  😃  One sequence was withdrawn from GenBank, so I removed that mapping.  5619 new mappings added.

Thanks again and I'm very glad you're working on GenBank metadata too!


emmahodcroft:
Thank you so much for this @AngieHinrichs  - as always! We're so lucky to have your hard work on this, and we appreciate it always! 😁

PR #132, opened 2021-02-16
update dependency setup script for gh actions;

eharkins:
according to release of nextstrain-cli v3.0.1


tsibley (approved):

eharkins:
Thanks @tsibley. Merging since I'm running the ingest tomorrow we can work together if for some reason this breaks it.

PR #135, opened 2021-02-26
Geo location rules

WandrilleD:

We propose to change geographic location substitutions from a manual annotation mode which is base of strain id, to a more generalist substitution based rule system.

We have implemented the rule substitutions, and aggregated rules from the manual annotation file as well as some of the files used by the parse_metadata.py file in the ncov repository.

We have also added a couple of utility scripts to help maintain the rule sets.


bin/filter-gisaid-annot-using-transform-log
```diff
@@ -0,0 +1,79 @@
...
+            if not filterOut:
+                linesToKeep.append(l)
+
+    with open(args.output,'w') as OUT:
+        for l in linesToKeep:
+            print( l , end = '' , file = OUT )
+    print('removed',removed,'annotation lines')
+
+
+    #for a in redundantAnnot:
```
eharkins, line 76: Can this commented code be removed?

bin/transform-gisaid
```diff
@@ -153,8 +213,8 @@ if __name__ == '__main__':
             )
         )
 
-    for unused_gisaid_epi_isl in annotations.get_unused_annotations():
-        print(f"WARNING: annotation for {unused_gisaid_epi_isl} was not used.")
+    #for unused_gisaid_epi_isl in annotations.get_unused_annotations():
```
eharkins, line 216: Can this commented code be removed if it's replaced by your added functionality in this PR?

bin/transform-gisaid
```diff
@@ -129,15 +172,32 @@ if __name__ == '__main__':
...
+        pipeline = ( pipeline  | WriteCSV(RAW_METADATA_FILENAME, 
+                                            METADATA_COLUMNS , 
+                                            restval = '?' , 
+                                            extrasaction ='ignore' , 
+                                            delimiter  = '\t', 
+                                            dict_writer_kwargs  = {'lineterminator': args.newline} ) )
+
+
+        # applying the substitution rules (temporary : writing the intermediary data to verify effect )
+        dict_writer_kwargs = {'lineterminator': args.newline}
```
eharkins, line 196: Do we need to assign this here if it is reassigned before being used below https://github.com/nextstrain/ncov-ingest/pull/135/files#diff-5ffbfeffdca16cf94d3822dca1129f8644b0430390a76df2551eed4db94cfa33R232 ?

eharkins (commented):
Thanks @WandrilleD and @MoiraZuber! Added a few comments about small things but this looks good overall and it sounds like the testing was successful based on messages in slack.

PR #136, opened 2021-02-26
Updates to accessions.tsv (13715 new, seqs downloaded 2021-02-26)

AngieHinrichs:
Happy two-weeks-flew-by-again.  Here are 13715 new GISAID <--> Genbank ID mappings.  🙂 

No promises, but if I were to automate this and run it more frequently than every two weeks, how often would you like to get these pull requests?

[If it can be completely automated someday then I will want to import the process into ncov-ingest, but at this point, duplicate submissions (to both GISAID and GenBank) make the output a little unstable and I end up doing some manual fixups to make sure the mapping stays 1-to-1 and as stable as possible.]

Thanks!
Angie


emmahodcroft:
Angie - from my end I now completely trust your work and no longer check it, and we've been absolutely happy to have your tireless contributions! So I'm happy to merge these PRs as fast as I get them (and, if on a busy day or weekend, as soon as I have a minute/am on a real computer!). If these started coming in very regularly then I think we'd be happy to discuss making you able to push straight to master for `ncov-ingest`. Let me know if you think you'd get to a point where that would make things smoother on your end, and I'll bring it up with the team :)

emmahodcroft:
I suppose worth adding one minor note - if these came in every day (as an example) I'd time merging them to some standardized time just so that they don't interfere with the build shepherds builds - but that would be very easy to do (but might mean a couple hours delay in merges, which I think is completely ok, but might be worth explaining up front!)

PR #137, opened 2021-03-04
Update to use pango_lineage 

emmahodcroft:
What comes in from GISAID is `covv_lineage`. We previously renamed this to `pangolin_lineage`, but in `ncov` PR [568](https://github.com/nextstrain/ncov/pull/568) we decided to rename this to `pango_lineage` to reflect the updated naming of this nomenclature scheme.

However, this currently isn't working, because in the rest of the pipeline we are still calling it `pangolin_lineage` starting with this renaming in `ncov-ingest`.

This PR changes the rename to `pango_lineage` and hopefully catches all other instances where this is referred to by name. (I found all instances through simple `grep` so if that may not have been sufficient please double-check I caught them all!)

Would appreciate extra eyes on this as I haven't tested it.

### Testing
UNTESTED

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -6,7 +6,7 @@ import pandas as pd
 
 INPUT_CLADE_COLUMN = "clade"
 OUTPUT_CLADE_COLUMN = "Nextstrain_clade"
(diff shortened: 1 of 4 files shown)


trvrb (approved):

trvrb:
Tested locally and everything seems well in order. I'm going to merge this now.

PR #140, opened 2021-03-05
Updates to accessions.tsv (8396 new, 9 removed, seqs downloaded 2021-03-05)

AngieHinrichs:
The sequences are pouring into GenBank and GISAID (yay!) so let's try for a weekly update.  There are still manual steps for newly duplicated sequences and diagnosing lost or changed mappings as name changes, removed sequences (sometimes fixing duplications) or sequences updated in GISAID but not GenBank.  (E.g. this time 2 sequences were updated in GISAID but not GB, and 7 sequences were removed from GB as contam or dups, so I removed those 9 mappings.)  Eventually I might be able to automate those a bit more but... not this week.  😆  Great news: we got some funding to hire someone to do this even better, as a general set of tools that we can share on github.

Thanks again!


emmahodcroft:
Amazing as always Angie! Thank you for your tireless work on this!

PR #147, opened 2021-03-17
rules for us cities to counties

eharkins:
Used this one time script to add rules correcting US cities to counties:
```python
#!/usr/bin/env python
import csv

def add_exception(exceptions, entry, conflicting_counties):
    conflicting_counties_s = '\n'.join([f"{c['city']}, {c['county_name']} County, {c['state_name']}" for c in conflicting_counties])
    exceptions.append(f"{entry['city']}, {entry['county_name']} County, {entry['state_name']}\n{conflicting_counties_s}\n\n")

def add_rule(rules, entry, counties_to_states_map):
    county_name = f"{entry['county_name']} County"
    if len(counties_to_states_map[entry['county_name']]) > 1:
        county_name += f" {entry['state_id']}"
    rules.append(f"North America/USA/{entry['state_name']}/{entry['city']}\tNorth America/USA/{entry['state_name']}/{county_name}\n")

if __name__ == '__main__':
    rules = []
    exceptions = []
    entries = []
    counties_to_states_map = {}
    with open('scripts/developer_scripts/config_files_parse_metadata/uscities.csv', 'r') as uscitiesfile:
        entries = [r for r in csv.DictReader(row for row in uscitiesfile if not row.startswith('#'))]
    for entry in entries:
        if counties_to_states_map.get(entry['county_name']) is None:
            counties_to_states_map[entry['county_name']] = set()
        counties_to_states_map[entry['county_name']].add(entry['state_id'])
    for entry in entries:
        if entry['city'] == entry['county_name']:
            add_rule(rules, entry, counties_to_states_map)
        else:
            conflicting_counties = [e for e in entries if e['county_name'] == entry['city'] and e['state_name'] == entry['state_name']]
            if len(conflicting_counties) > 0:
                add_exception(exceptions, entry, conflicting_counties)
            else:
                add_rule(rules, entry, counties_to_states_map)
    with open('us_cities_rules.tsv', 'w') as outfile:
        for l in rules:
            outfile.write(l)
    with open('exceptions.tsv', 'w') as outfile:
        for l in exceptions:
            outfile.write(l)
```
with exceptions where city names conflicted with a different county of the same name in the same state, as discussed in https://github.com/nextstrain/ncov/pull/577. In those cases there's not a good way for us to know if the submitter intended the city or the county with that name.

Some examples of such exceptions:
```
Houston, Harris County, Texas
Crockett, Houston County, Texas

Austin, Travis County, Texas
Sealy, Austin County, Texas

Wichita, Sedgwick County, Kansas
Leoti, Wichita County, Kansas
```
This adds A LOT of lines to this file which might be of concern if it affects performance of scripts in this repo that loop over the rules or just store them all in memory. In general it might be a concern to store a large file in git without using lfs. You have to check this out locally to be able to see the diff of the commit.


rneher:
HI @eharkins,

I don't see a reason not to add this to git. Did you notice any performance hits with this large increase in rule number? Roughly 4-times more, right? I suppose this comes with a large decrease in some other file? Did you check how many of these rules can be abstracted in some way?

eharkins:
@rneher 

> Did you notice any performance hits with this large increase in rule number? 

Not when running `bin/transform-gisaid` locally.

> Roughly 4-times more, right?

Right.

> I suppose this comes with a large decrease in some other file?

No, this is mostly anticipating future USA locations and correcting some existing ones.

> Did you check how many of these rules can be abstracted in some way?

Yes, I ran @WandrilleD's script to check this and wasn't able to remove any.

There's definitely a chance that I didn't consider certain exceptions and so some of these rules are wrong. For example I just remembered counties in Louisiana are called "Parish"es. This is an easy fix to the script, but in general there could be other mistakes here due to edge cases. I suppose we could merge and add these to the end of the rule list, and not sort the list as we do regularly, so that they are easy to remove if they begin causing more trouble than good.

MoiraZuber:
I'm in favor of adding these rules. In case this increases the duration of ingest significantly, we can always revert back to the original plan of suggesting corrective rules when they first appear and adding them manually one by one. @eharkins does the current script consider duplicates (existing and/or new)? This and the "Parish" naming convention are the only potential issues I can think of.

eharkins:
@emmahodcroft @MoiraZuber I updated this so that it can be merged. I am almost certain it might cause some issues with edge cases where https://simplemaps.com/data/us-cities happens to differ with some existing geo location rules we had. But we can smooth that out if we think this will save us time in the long run. Also not sure exactly where is best to cite https://simplemaps.com/data/us-cities.

emmahodcroft:
I think this is great Eli - thanks! Yes, I think we might just have to keep an eye out for the edge-cases you mention. Unfortunately I think with the continued ingests there'll be some conflict resolution - I think it's fine to merge this whenever you next get a chance to make a pass.

For the credit, I would hope that putting this somewhere in the ReadMe of `ncov-ingest`, though perhaps this isn't quite within the specifications. Alternatively, we could add a link, or a link to the `ncov-ingest` Readme, to the description of the North America build. 

Here's the attribution details:
```
If I use the free database, what type of attribution is required?

If you use the Basic (free) database, you must link back to this page:
https://simplemaps.com/data/us-cities
from a public webpage where you are using the data. If you want to use the data internally, you must link back to this page from your firm's website on a page that can be easily found though links on the root domain. The link must be clearly visible to the human eye.
```

@rneher @trvrb I'd appreciate your thoughts on this!

eharkins:
Merged in https://github.com/nextstrain/ncov-ingest/commit/fb6b35a2f94fe6ba5e12b35e85ec6568c02e022d

PR #159, opened 2021-04-15
Snakemake

WandrilleD:

This pull request replace the bin/ingest-{database} scripts with a unified snakemake pipeline.

The rules have been written to keep the same process as before, putting emphasis on the commonalities between the treatment of gisaid and genbank data (currently the only 2 implemented database).

The slack notifications have also been pushed outside of bash scripts to make them more explicit and easy to handle from the snakefile (although for the sake of code clarity, they still employ the `bin/notify-slack`).

The rules DAG can be printed in image form with a command such as :
 ```
snakemake --rulegraph all_then_clean | dot -Tsvg > dag.svg
```
and then visualized using your favorite image viewer.

### Testing

Testing through the full ingest-gisaid and ingest-genbank pipeline is advised.
Provided these change replace the actual scripts, the git actions will not do and
one should utilize the following instead to realize the pipelines :
```
snakemake gisaid_then_clean
```
```
snakemake genbank_then_clean
```

For the script to work, the environment should contain the following variables:
 *   "S3_SRC" 
 *   "GITHUB_REF" # for instance "refs/heads/master"
 *   "GISAID_API_ENDPOINT"
 *   "GISAID_USERNAME_AND_PASSWORD" 
 *   "SLACK_TOKEN" # token to post on slack channel
 *   "FETCH"  # set to 1 to trigger download of data from the database

### CODE REVIEW

**Comment by eharkins** [comment]
Thanks for this work @WandrilleD!

> Provided these change replace the actual scripts, the git actions will not do and
one should utilize the following instead to realize the pipelines :

I think we can stick with github actions to trigger this workflow to run on s3. The difference here is that now it's in snakemake, so we'll have to change this bit https://github.com/nextstrain/ncov-ingest/blob/master/.github/workflows/ingest-gisaid-master.yml#L37-L45 to run the snakemake workflow, something similar to:
```
        nextstrain build \
          --aws-batch \
          --no-download \
          --image nextstrain/ncov-ingest \
          --cpus 8 \
          --memory 15GiB \
          --exec env \
          . \
            envdir env.d snakemake gisaid_then_clean
```

File: snakefile [hunk 1]
```diff
@@ -0,0 +1,362 @@
...
+                if [[ $added_records -gt 0 ]]; then
+                    msg="📈 New nCoV records (n=$added_records) found on {params.database}."
+                elif [[ $added_records -lt 0 ]]; then
+                    msg="WARNING: the new version of {params.database} has fewer records‽"
+                
+                else
+                    msg="📈 New nCoV records (n=$added_records) found on {params.database}."
+                fi
+
+                ./bin/notify-slack $msg $SLACK_TOKEN {params.slack_channel}
```
eharkins, line 105: If we are running this somewhere public where the logs can be viewed, like we do now with ingest (https://github.com/nextstrain/ncov-ingest/runs/2353020060?check_suite_focus=true#step:4:271), will people be able to see the value of the `SLACK_TOKEN` variable as it is passed as an argument to `bin/notify-slack`? 

If so, can we avoid this by using it directly from the environment variable in any bash scripts that need it like this https://github.com/nextstrain/ncov-ingest/pull/159/files#diff-3d879f93bb87b9c04449da332db37bd189cdc4b38d7703b659d126dc9e6dec03L4 ?

emmahodcroft:
Thanks for this work @WandrilleD  - this seems immensely useful. If we run this pipeline through locally at the same time as running the 'normal'/current ingest, would we expect all outputs to be identical? (Would that be a good way of testing?)

snakefile
```diff
@@ -0,0 +1,362 @@
...
+            --output-fasta {output.sequences} \
+            --problem-data {output.problem} > {output.flagged_annotation}
+
+        touch {output.additional_info}
+        '''
+
+
+
+
+rule dowload_old_clades :
```
rneher, line 154: spelling...

snakefile
```diff
@@ -0,0 +1,362 @@
...
+        idcolumn=lambda wildcards : config['idcolumn'][wildcards.database]
+    output :
+        "data/{database}/location_hierarchy.tsv"
+    shell:
+        "./bin/check-locations {input} {output} {params.idcolumn}"
+
+
+rule notify_and_upload:
+    input :
+        sequences = lambda wildcards : "data/"+wildcards.database+"/sequences.fasta",
```
rneher, line 221: we can use wildcards directly here
`sequences = "data/{database}/sequences.fasta",`

rneher:
I think `FETCH` would be better as a `config` variable. This could be passed in a command line argument to snakemake as

`snakemake gisaid_then_clean --config fetch=source` 

or 

`snakemake gisaid_then_clean --config fetch=S3` 

or similar.

rneher:
I think I'd prefer to have a rule `ingest` that produces all the input for the notify and upload rules such that we can separate running the pipeline. The output of that rule could then be in the input to the notify rule. 

I am not too familiar with the `envar` directives in snakemake. can one make them rule specific? because currently it wants you to specify a GISAID_URL if you want to do the genbank ingest.

WandrilleD:
Thank you for your reviews.
I think that with these commits I have addressed most of your comments.

@eharkins from the test that I have run the content of snakemake tokens (such as {params.slack_token} for instance) are revealed in the logs (especially with option -p) whereas the content of an environment variable would not be revealed. This is one of the reason I use a environment variable for the slack token.

regarding the github action, I defer to your judgment as I far no practical experience with these.
Taking into account the new configuration of the script, it would now look like : 

```

nextstrain build \
          --aws-batch \
          --no-download \
          --image nextstrain/ncov-ingest \
          --cpus 8 \
          --memory 15GiB \
          --exec env \
          . \
            envdir env.d snakemake gisaid_then_clean \
               --config fetch=yes gisaid_login='${{ secrets. ... }}' gisaid_endpoint='${{ secrets. ... }}' slack_token='${{ secrets.SLACK_TOKEN }}'

```



@emmahodcroft you are correct! The only change would be that the standard output of transform-gisaid previously went to a temporary file names named : "flagged-annotations-XXXXXX" whereas now it goes to "data/gisaid/transform-log.txt" (the name is more indicative of what this file means, in particular across the different data-sources).

@rneher  `envar` cannot, AFAIK, be made rule specific. I tried something by having the used set them using `--config`, with an error popping up when a rule is called without the proper elements set.
Also, there is now an `ingest_gisaid` and `ingest_genbank` rule.

rneher:
Thanks, Wandrille. I pushed this branch to nextstrain/ncov-ingest such that CI runs and we can do trial runs more easily:
https://github.com/nextstrain/ncov-ingest/tree/snakemake

joverlee521:

PR #160, opened 2021-04-16
Updates to accessions.tsv (22340 new, 1742 removed, seqs downloaded 2021-04-16)

AngieHinrichs:
22k new mappings!  :)  1742 sequences from Switzerland have been withdrawn from GenBank (submitter on the few that I checked: Swiss Pathogen Surveillance Platform).

Thanks & have a great weekend.  :)


emmahodcroft:
Thank you Angie for this fantastic update! And thanks also for the flag about the Swiss Pathogen Surveillance Platform. We have some connections to this and can think about whether this is worth chasing up. They're just getting set up at the moment so it may have been a mistaken double-upload, error in data, or similar.

@rneher is it worth looking into what might have happened here with SPSP?

emmahodcroft:
@AngieHinrichs - seems like these were messed up sequences, but hopefully we'll be getting some more up on Genbank from Switzerland soon! They're still fine-tuning the pipeline, but your linking will be absolutely fantastic to have here!

AngieHinrichs:
Thank you @emmahodcroft for looking into what happened with those!  I checked a few in GISAID and they seemed to still be there (e.g. EPI_ISL_418271, EPI_ISL_830785, EPI_ISL_931450) - should they be withdrawn from GISAID as well?  Glad to hear more are on the way.  :)

emmahodcroft:
Hmmm, that's curious. We'll see if we can look into it a bit more!

PR #163, opened 2021-05-06
Schedule GenBank ingests

tsibley:
Timing is on Monday night (Seattle) / Tuesday morning (Basel).  We want
to start making regular builds from this data, so keep it up to date
automatically.


rneher (approved):
looks good to me. It looks like CI failed bc of bad zip file from gisaid, possibly being written at the time?

PR #164, opened 2021-05-06
Upload GenBank files to s3://nextstrain-data/files/ncov/open/…

tsibley:
For downloading by the ncov "open data" build.  We'll be consolidating and uploading additional files from the ncov build to the same location as well.

Removes "genbank_" prefix from filenames because in time we'll start ingesting other Open Data sources (e.g. COG-UK) into these same files.

## Todo

* [x] Figure out where we want `branch/` and `tmp/` prefixes to live, if not under `files/ncov/open/{branch,tmp}/…`.
* [x] Update [NextstrainNcovIngestUploader](https://console.aws.amazon.com/iam/home?region=us-east-1#/policies/arn:aws:iam::827581582529:policy/NextstrainNcovIngestUploader$jsonEditor) IAM policy for paths
* [x] Configure S3 lifecycle policies for new branch/tmp data, if necessary.

## Cleanup (after merge)

* [ ] Delete [NextstrainNcovPrivateUploader](https://console.aws.amazon.com/iam/home?region=us-west-2#/policies/arn:aws:iam::827581582529:policy/NextstrainNcovPrivateUploader$jsonEditor) and [NextstrainDataUploader](https://console.aws.amazon.com/iam/home?region=us-west-2#/policies/arn:aws:iam::827581582529:policy/NextstrainDataUploader$jsonEditor) IAM policies. They're only attached to the `nextstrain-ncov-ingest-uploader` user, which already now has a new combined policy attached covering what it needs for this PR.

### DIFF
--- a/bin/ingest-genbank
+++ b/bin/ingest-genbank
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-: "${S3_SRC:=s3://nextstrain-data/ncov-ingest}"
+: "${S3_SRC:=s3://nextstrain-data/files/ncov/open}"
 : "${S3_DST:=$S3_SRC}"
 
 # Determine where to save data files based on if we're running as a result of a
@@ -21,12 +21,12 @@ case "${GITHUB_REF:-}" in
         # Save data files under a per-branch prefix
         silent=yes
         branch="${GITHUB_REF##refs/heads/}"
-        S3_DST="$S3_DST/branch/$branch"
+        S3_DST="${S3_DST/nextstrain-data/nextstrain-staging}/branch/$branch"
         ;;
     "")
         # Save data files under a tmp prefix
         silent=yes
-        S3_DST="$S3_DST/tmp"
+        S3_DST="${S3_DST/nextstrain-data/nextstrain-staging}/tmp"
         ;;
     *)
         echo "Skipping ingest for ref $GITHUB_REF"
@@ -88,13 +88,13 @@ fi
 
 
 if [[ "$branch" == master ]]; then
-    ./bin/notify-on-metadata-change data/genbank/metadata.tsv "$S3_SRC/genbank_metadata.tsv.gz" genbank_accession
+    ./bin/notify-on-metadata-change data/genbank/metadata.tsv "$S3_SRC/metadata.tsv.gz" genbank_accession
     ./bin/notify-on-problem-data data/genbank/problem_data.tsv
     ./bin/notify-on-location-hierarchy-addition data/genbank/location_hierarchy.tsv source-data/location_hierarchy.tsv
 fi
 
-./bin/upload-to-s3 ${silent:+--quiet} data/genbank/metadata.tsv "$S3_DST/genbank_metadata.tsv.gz"
-./bin/upload-to-s3 ${silent:+--quiet} data/genbank/sequences.fasta "$S3_DST/genbank_sequences.fasta.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/genbank/metadata.tsv "$S3_DST/metadata.tsv.gz"
+./bin/upload-to-s3 ${silent:+--quiet} data/genbank/sequences.fasta "$S3_DST/sequences.fasta.gz"
 ./bin/upload-to-s3 ${silent:+--quiet} data/genbank/nextclade.tsv "$S3_DST/nextclade.tsv.gz"
 
 ./bin/clean
--- a/bin/transform-genbank
+++ b/bin/transform-genbank
@@ -303,7 +303,7 @@ if __name__ == '__main__':
         description=__doc__,
         formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument("genbank_data",
-        default="s3://nextstrain-data/ncov-ingest/genbank.ndjson.gz",
+        default="s3://nextstrain-data/files/ncov/open/genbank.ndjson.gz",
         nargs="?",
         help="Newline-delimited GenBank JSON data")
     parser.add_argument("--annotations",


tsibley:
Marking as draft. Let's wait to merge this until we have [a migration plan](https://github.com/nextstrain/ncov/pull/626#issuecomment-833718070).

rneher (approved):

rneher:
this seems to need some modifications to S3 permissions though:
```
[batch] upload failed: - to s3://nextstrain-data/files/ncov/open/branch/genbank-upload-location/genbank.ndjson.gz An error occurred (AccessDenied) when calling the CreateMultipartUpload operation: Access Denied
```

tsibley:
@rneher Thanks, I hadn't looked at the CI results yet. Fixed IAM permissions.

tsibley:
CI passed; GISAID ingest failed because gisaid.org is down.

trvrb:
Hmm... this is tricky and I don't have strong opinions here. Trying to think through slightly, I had been considering `files/ncov/` as files provisioned by us for use with the `ncov` pipeline. I'm okay thinking of these branch-specific `ncov-ingest` provisioning as targeting `files/ncov/open/branch/test/asia/aligned.fasta.xz` rather than `files/ncov/open/asia/aligned.fasta.xz`. This makes sense to me in continuing the hierarchy.

Hmm... also complicated is that these files are provisioned between `ncov-ingest` and `ncov`. You might have branches from either repo. I really liked the setup with `/staging/trial/` for `ncov` "trial" builds, where you can name trials as you see fit (and might commonly use branch name, but it's not enforced). `files/ncov/open/trial`?

rneher:
Maybe we could have something like `nextstrain-data/scratch/ncov/open/branch/{branch_name}/*`?

tsibley:
Thanks, both. @trvrb's vote seems to be for 1. @rneher's for 3. I don't have a strong opinion.

With Richard's suggestion of `/files/` vs. `/scratch/`, I do like the parallelism of using an entirely different prefix and the connotation (correctly) imparted by "scratch". This would also avoid needing exclusion rules if at some point we generate directory indexes of `/files/`.

Agreed it's important to think about how this interacts with the trial builds. Note that currently trial builds only run with the GISAID data, and they always download the latest inputs (e.g. they don't currently allow downloading from a branch/tmp ncov-ingest run). Trial results are uploaded to a separate location, currently `nextstrain-ncov-private/trial/${trial_name}/`. If we want to extend trial builds to the "open" builds and publicly expose the trial build intermediates, I'd recommend putting them under `nextstrain-data/files/staging/trial/${trial_name}/ncov/open/`, so that the path continues to parallel the `nextstrain.org/staging/trial/${trial_name}/ncov/open/` paths.

The ncov-ingest branch/tmp runs are slightly different from the ncov trial runs in that the former are automatic and the latter are manually triggered. This means they have different constraints on how they name output locations, particularly with regard to systematically avoiding naming conflicts vs. assuming the human will take care of it.

jameshadfield:
> I'd recommend putting them under nextstrain-data/files/staging/trial/${trial_name}/ncov/open/, so that the path continues to parallel the nextstrain.org/staging/trial/${trial_name}/ncov/open/ paths.

While we haven't ever formalised "nextstrain-staging" vs "nextstrain-data", my vote would be to store data from all trial runs, as well as automatic runs which need human vetting, in the "nextstrain-staging" bucket, and thus access them via nextstrain.org/staging URLs.

tsibley:
Ah, yes, good point. The Auspice JSONs for trial builds are currently uploaded to `nextstrain-staging`, but we could also put the intermediates for trials in that bucket, e.g. `nextstrain-staging/files/trial/${trial_name}/ncov/open`.

tsibley:
Ok. Here's a proposal that tries to harmonize everything a bit more.

    ncov-ingest outputs uploaded to…
      master branch  → nextstrain-data/files/ncov/open/
      other branches → nextstrain-staging/files/ncov/open/branch/${branch_name}/
      ad-hoc         → nextstrain-staging/files/ncov/open/tmp/

    ncov intermediates uploaded to…
      master branch  → nextstrain-data/files/ncov/open/
      other branches → [nowhere; nothing runs for these automatically]
      trial builds   → nextstrain-staging/files/ncov/open/trial/${trial_name}/

Notably:

- Non-production data/builds always end up in s3://nextstrain-staging, never right alongside the production data in s3://nextstrain-data.  This keeps `nextstrain-data/files/ncov/open/` less cluttered and reduces potential for confusion with branch, tmp, and trial runs, while not introducing new top-level hierarchy like `scratch`.

- Trial builds switch order of `ncov/open` and `trial/${trial_name}` in order to a) be consistent and b) namespace the trial under the ncov build instead of at the top-level.  The nextstrain.org URLs to view the trial datasets would also change to match, e.g.

      from nextstrain.org/staging/trial/${trial_name}/ncov/open/global
        to nextstrain.org/staging/ncov/open/trial/${trial_name}/global

  This will be easy with the soon-to-be-merged `auspice_json_prefix` config option in #643.

There's maybe more we could do to converge the automatic branch runs of ncov-ingest with the manual trial runs of ncov, but I'm not sure it's worth worrying about right now.

jameshadfield:
This looks pretty good to me. @tsibley would you mind adding the gisaid information to that table for easy comparison?

> The nextstrain.org URLs to view the trial datasets would also change to match ... `nextstrain.org/staging/ncov/open/trial/${trial_name}/global`

Just to clarify, we're going to (or already are?) creating files such as `s3://nextstrain-staging/ncov_open_trial_${trial_name}_global.json`?

> while not introducing new top-level hierarchy like scratch.

We do have the `s3://nextstrain-scratch` bucket ;)

tsibley:
> This looks pretty good to me. @tsibley would you mind adding the gisaid information to that table for easy comparison?

Will do.

> Just to clarify, we're going to (or already are?) creating files such as `s3://nextstrain-staging/ncov_open_trial_${trial_name}_global.json`?

Yep. Trial builds currently create files like:

    s3://nextstrain-staging/trial_${trial_name}_ncov_open_global.json 

but with the proposal above, they'd switch to:

    s3://nextstrain-staging/ncov_open_trial_${trial_name}_global.json

> We do have the `s3://nextstrain-scratch` bucket ;)

Ha, so we do! But we don't use it anywhere for nextstrain.org, right?

tsibley:
A similar table for outputs from our GISAID runs. (Seemed easier to separate out than figure out how to legibly combine in one.)

    ncov-ingest outputs uploaded to…
      master branch  → nextstrain-ncov-private/
      other branches → nextstrain-ncov-private/branch/${branch_name}/
      ad-hoc         → nextstrain-ncov-private/tmp/

    ncov intermediates uploaded to…
      master branch  → nextstrain-ncov-private/
      other branches → [nowhere; nothing runs for these automatically]
      trial builds   → nextstrain-ncov-private/trial/${trial_name}/

This is the current behaviour, though the exact files uploaded back by ncov runs would change a little with https://github.com/nextstrain/ncov/pull/628.  Compared to the table above for "open" runs, it's basically substituting:

    nextstrain-{data,staging}/files/ncov/open/ → nextstrain-ncov-private/

emmahodcroft:
This is a very naive question, but what's a trial run/build vs a 'non-master branch' run? Sorry!

tsibley:
@emmahodcroft Not naive!

Best description of trial ncov builds is [in Slack](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1612203970011100).

"non-master branch" means a build started from any branch other than `master`, e.g. the [automatic runs of ncov-ingest for this PR](https://github.com/nextstrain/ncov-ingest/pull/164/checks) or a trial build of an ncov branch ([recent example](https://github.com/nextstrain/ncov/runs/2677037937?check_suite_focus=true)).

tsibley:
Created two S3 lifecycle rules for nextstrain-staging, named:

- Cleanup dev files under files/ncov/open/branch/
- Cleanup dev files under files/ncov/open/tmp/

tsibley:
Rebased onto latest `master`.

emmahodcroft:
Thank you @tsibley ! I have that post bookmarked and everything... 😆  But clearly forgot about it!

PR #166, opened 2021-05-10
Auto-deploy our core builds to production instead of staging

tsibley:
Discussion in Slack¹ led to consensus to switch. The reasoning is that
nearly all the time the build shepherd's are just copying files between
staging and production. Any occasional bad build can be fixed by rolling
back to yesterday's build (from the dated JSON files).

Mirrors [corresponding change](https://github.com/nextstrain/ncov/pull/629) in ncov workflow.

¹ https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1620422859124500

### DIFF
--- a/.github/workflows/rebuild.yml
+++ b/.github/workflows/rebuild.yml
@@ -1,4 +1,4 @@
-name: Rebuild nextstrain.org/staging/ncov
+name: Rebuild nextstrain.org/ncov
 
 on:
   # Manually triggered using bin/trigger rebuild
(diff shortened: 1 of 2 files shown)


rneher (approved):

tsibley:
Repushed to remove some references to "staging" in filenames and output messages.

tsibley:
Updated the (user-specific) `NextstrainNcovIngestUploader` IAM policy to additionally allow writing of `s3://nextstrain-data/ncov_*` objects, since the ncov builds run on Batch with the credentials of the `nextstrain-ncov-ingest-uploader` user.

PR #167, opened 2021-05-14
Updates to accessions.tsv (32634 new, 337 updated, 79 removed) seqs downloaded 2021-05-14

AngieHinrichs:
In addition to the usual additions and deletions, this time 337 mappings updated GenBank accession version .1 to .2 because sequences were updated in both GISAID and GenBank (yay!).

Thanks as always and enjoy your weekend!  :)

Angie


emmahodcroft:
Fantastic! Thank you as always Angie! I heard that GISIAD was introducing versioning (a similar .1 .2 etc) - are you seeing any signs of this yet by chance, or are you just seeing the updates in sequence comparisons?

AngieHinrichs:
I have noticed some .2's on GISAID's web site, just when clicking to view a specific sequence's info, but I figured you guys would know before I would about whether those would be available for download.  :)  I am only able to download fasta sequences from queries of 10,000 or fewer sequences at a time, no metadata besides what's in the fasta headers (e.g. ">hCoV-19/USA/VA_NIDDL_20357/2020|EPI_ISL_1624267|2020-08-24").  I have not noticed any .2's in there yet.

emmahodcroft:
Ah ok - no, I don't think we've seen it either. At least, no EPI_ISL numbers in the metadata files we generate from what we get from GISAID seem to have any `.2` or similar after them. But, I didn't know if this meant they just haven't been pushed to the info we get, but might be present somewhere else...! Strange if they are appearing in the "clicky" metadata but nowhere else yet, but perhaps they'll appear in other forms soon.

PR #168, opened 2021-05-20
feat: add substition columns from nextclade to metadata

rneher:
This branch adds two additional columns from the nextclade output to the metadata.


huddlej:
@rneher This seems potentially really useful for some features we've discussed internally. Would you be ok with merging this as it is or are there concerns that this could break our downstream processes?

PR #169, opened 2021-05-20
feat: change compression of sequences to xz

rneher:
compress sequences with `xz -2 -c` instead of `gz` due to superior performance on fasta files (with nearly identical genomes)


bin/upload-to-s3
```diff
@@ -34,6 +34,8 @@ main() {
         echo "Uploading $src → $dst"
         if [[ "$dst" == *.gz ]]; then
             gzip -c "$src"
+        elif  [[ "$dst" == *.xz ]]; then
+            xz -2 -c "$src"
         else
             cat "$src"
         fi | aws s3 cp --no-progress - "$dst" --metadata sha256sum="$src_hash" "$(content-type "$dst")" "$(content-encoding "$dst")"
```
joverlee521, line 41: I believe the GitHub action is failing at the this `aws s3 cp` step. 
Looks like `content-type` and `content-encoding` also need to be updated to expect `.xz` files.

ivan-aksamentov: I attempted to address this in 3ef9082ee9f1070c974bbf9e1c5c655f09d1d804

https://github.com/nextstrain/ncov-ingest/blob/3ef9082ee9f1070c974bbf9e1c5c655f09d1d804/bin/upload-to-s3#L56-L74

Considerations:
 - Do we even need to set these in the first place?
 - The content-type function now remove both `.gz` and `.xz` suffixes before trying to deduce the file extension. This would not fork for `.gz.xz` files, if we ever need those.
 - No browser actually supports xz, lzma or lzma2 encoding, so the content-encoding part is pretty much useless in this case. Might be better to just set content-type to xz.

One place where I noticed content-type and content-encoding make a difference is when manually downloading files from S3 web interface. In this case `.gz` files get automatically decompressed by the browser and the downloaded file is actually the text file, but with `.gz` extension. This is half-convenient, half-confusing. And the AWS CLI ignores these fields entirely, I think. Not sure we even need this whole content type and encoding business.

ivan-aksamentov:
I added a new script called `./bin/download-from-s3` which does the opposite of `./bin/upload-to-s3`, that is downloads and decompresses a file, whether it's `.gz` or `.xz`. We now need to carefully replace all the `aws cp ... | gunzip ...` commands with it.

rneher:
Looks like this work now! we need to time this merge with a change in `ncov` since it is currently expecting a `gz` file.

emmahodcroft:
Yes, It would be good to know when we make this switch so I can adjust my download rules, as I imagine the old files will continue existing and I could continue pulling old files without realising (this has happened before 🙃 ).

bin/upload-to-s3
```diff
@@ -52,7 +54,11 @@ main() {
 }
 
 content-type() {
-    case "${1%.gz}" in
+    BASENAME="${1}"
+    BASENAME="${BASENAME%.gz}"  # Remove trailing ".gz"
+    BASENAME="${BASENAME%.xz}"  # Remove trailing ".xz"
+
+    case "${BASENAME}" in
```
tsibley, line 61: The simpler change here is only a few characters:

```diff
-    case "${1%.gz}" in
+    case "${1%.[gx]z}" in
```

and avoids the double-removal edge case issue [you noted](https://github.com/nextstrain/ncov-ingest/pull/169/files#r638212162) in another comment.

bin/upload-to-s3
tsibley: Setting Content-Type and Content-Encoding (but see caveat below) is a bit friendlier, though maybe not strictly necessary for any of our own code using these files. We do intend these files for the "open" data to be publicly consumed, so friendliness is important to consider there.

Agreed that serially removing the extensions is no good (even though I don't think `.gz.xz` would ever be a concern). See my comment below for a better alternative.

You're right that `Content-Encoding: xz` won't work because clients won't understand it, and this will generally be a fatal error. We should avoid that! Instead, returning `Content-Type: application/x-xz` makes sense I think, though it will mask the contained data type.

You're also right that the Content-Encoding + extension is a bit weird, but it makes it possible to transparently decompress the response, e.g. `curl --compressed … > output` will Just Work, as will fetches from many HTTP client libraries.

tsibley: I meant to add that for downloading, I think we actually want to move away from `aws s3` API operations and instead access via our CDN'd HTTP endpoints (e.g. data.nextstrain.org), which is why HTTP headers matter a bit.

bin/download-from-s3
```diff
@@ -0,0 +1,77 @@
...
+        if [[ $quiet == 1 ]]; then
+            echo "Quiet mode. No Slack notification sent."
+            exit 0
+        fi
+
+        if ! "$bin"/notify-slack "Updated $dst available."; then
+            echo "Notifying Slack failed, but exiting with success anyway."
+        fi
+    else
+        echo "Files are identical, skipping upload"
```
tsibley, line 53: `s/upload/download/`

bin/download-from-s3
```diff
@@ -0,0 +1,77 @@
...
+
+        if [[ $quiet == 1 ]]; then
+            echo "Quiet mode. No Slack notification sent."
+            exit 0
+        fi
+
+        if ! "$bin"/notify-slack "Updated $dst available."; then
+            echo "Notifying Slack failed, but exiting with success anyway."
+        fi
+    else
```
tsibley, line 52: Why would we want to notify Slack on download?

bin/upload-to-s3
ivan-aksamentov: Applied in 71fd4a1eaa497c555924033ff25dca8412858697

bin/download-from-s3
ivan-aksamentov: > returning Content-Type: application/x-xz makes sense I think
Addressed in b9690e8c850d56bc2d83141a22b63e12549a98e2

> s/upload/download/

Addressed in bf7bf8b8ba98bbc7119c45c2b776039bdf3f8980

bin/download-from-s3
ivan-aksamentov: Because "we" copy-paste a lot :)

Removed in ee73938fa6e67c00ce20a721b3b3d39259737c61

jameshadfield:
Closing this PR as a rebased version (#194) has been merged

PR #170, opened 2021-05-21
Replace Nextclade 0.x with Nextclade 1.x

ivan-aksamentov:

This makes ingest to use the new Nextclade C++ implementation. There is a few differences in this version.

#### Differences compared to 0.x:

 - It should be much faster
 - It does not require input sequences batching or external parallelization, so batching scripts are removed.
 - Nodejs is no longer required
 - Added `time` command to docker container to measure Nextclade's performance.
 - Nextclade 1.x requires additional external [data](https://github.com/nextstrain/nextclade/tree/master/data/sars-cov-2) to be downloaded. This is handled inside `./bin/run-nextclade` script. This data is placed to `data/gisaid/nextclade-inputs`. Perhaps should be moved to `tmp/` or similar?
 - There are now some additional outputs you cannot opt out from, such as aligned sequences, peptides and insertions. They currently go to `data/gisaid/nextclade`. We should probably make these optional in Nextclade itself.
 - If input fasta is empty, Nextclade currently fails with error (bug or feature)? So there is an external check in `./bin/ingest-gisaid` to account for that and skip if there are no sequences. However, when it is the case, the script `./bin/flag-metadata` fails with:
    ```
    AttributeError: 'DataFrame' object has no attribute 'divergence
    ```
    Calculation of divergence and other metrics should probably be moved out from `./bin/join-metadata-and-clades`.

#### Other considerations:
 - Full recompute on entire dataset will be needed when going live with this
 - The flag `--in-order` is used to preserve the order of outputs in TSV. I think this is needed for `./bin/join-metadata-and-clades` to function properly.
 - The new flag `--include-reference` is currently not used, as we seem to not care about the ref sequence. It will also break the order.

#### Future improvements:

 - The newly detected sequences can be placed on auspice tree if needed with `--output-tree`. This has some performance hit: both memory consumption and and runtime.
 - The JSON output can be written. It contains more info compared to TSV and easier to parse. This also has some performance hit: both memory consumption and and runtime.

Files of interest:

https://github.com/nextstrain/ncov-ingest/blob/0f5716348571cbe3b77bd168e5b3c57fa5bfebd7/bin/ingest-gisaid#L96-L115

https://github.com/nextstrain/ncov-ingest/blob/0f5716348571cbe3b77bd168e5b3c57fa5bfebd7/bin/run-nextclade#L25-L54

### Related issue(s)  
-

### Testing
-


rneher:
This has passed. I updated the clades on S3 such that they contain 21A and will merge this now. It will now use the new version of nextclade which is a lot faster.

PR #172, opened 2021-05-24
fetch-from-genbank: Increase CSV field size limit

tsibley:
Runs were failing because a GenBank-provided data field started
exceeding the default limit of 128 KiB.¹  5 MiB is an arbitrarily larger
amount.

¹ https://github.com/python/cpython/blob/7148293d/Modules/_csv.c#L1686-L1687


tsibley:
5 MiB is not large enough. I'll do some actual investigation into what size will work before repushing again, rather than just guessing.

PR #174, opened 2021-05-27
Snakemake

WandrilleD:
# Description of proposed changes

This pull request replace the bin/ingest-{database} scripts with a unified snakemake pipeline.

The rules have been written to keep the same process as before, putting emphasis on the commonalities between the treatment of gisaid and genbank data (currently the only 2 implemented database).

The slack notifications have also been pushed outside of bash scripts to make them more explicit and easy to handle from the snakefile (although for the sake of code clarity, they still employ the bin/notify-slack).

The rules DAG can be printed in image form with a command such as :

snakemake --rulegraph all_then_clean | dot -Tsvg > dag.svg

and then visualized using your favorite image viewer.

# Testing

The git workflows have been adapted to perform similarly as before with the new system.
This morning :
* Shellcheck and gisaid-fetch pass
* Genbank fails for reasons similar to the problem that occurred yesterday (related to reading a csv file AFAIK)


bin/setup-github-workflow
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 # Adds $HOME/.local/bin to PATH for subsequent workflow steps (but not this one!).
 echo "$HOME/.local/bin" | tee -a "$GITHUB_PATH"
 
-echo "::group::Installing/upgrading Pip, setuptools, and wheel"
+echo "::group::Installing/upgrading Pip, setuptools, amd wheel"
```
tsibley, line 7: Typo addition slipped in here.

snake_config.yaml
```diff
@@ -0,0 +1,15 @@
+
+idcolumn:
+    gisaid:  gisaid_epi_isl
+    genbank: genbank_accession
+
+
+slack_channel:
+    genbank: ncov-genbank-updates
+    gisaid: ncov-gisaid-updates
```
tsibley, line 9: The use of lookup keys like this in config jumps out to me as a sort of anti-pattern. A less brittle approach is to get rid of the lookup keys and instead split one config into multiple, for example:

**genbank.yaml**

```yaml
idcolumn: gisaid_epi_isl
slack_channel: ncov-gisaid-updates
```

**gisaid.yaml**

```yaml
idcolumn: genbank_accession
slack_channel: ncov-genbank-updates
```

snakefile
```diff
@@ -0,0 +1,495 @@
...
+            # checking additions
+            if [[ -s "$additions" ]]; then
+                # "Notifying Slack about metadata additions."
+                ./bin/notify-slack --upload "metadata-additions.tsv" $SLACK_TOKEN {params.slack_channel} < "$additions"
+
+                if [[ "{params.idcolumn}" == "gisaid_epi_isl" ]]; then
+                    ./bin/notify-users-on-new-locations "$additions" --slack-token $SLACK_TOKEN --slack-channel {params.slack_channel}
+                fi
+            fi
+            touch {output}
```
tsibley, line 325: It's not clear to me why it's beneficial to pull out the notification logic from the computation in this rule and the others like it. That is, why change `notify-on-metadata-change` to `compute-metadata-change` + the embedded shell in this rule? It's particularly odd to me because the computation is solely for the sake of the notification (right?), so this change ends up separating the two pieces of tightly-coupled code.

tsibley (changes requested):
Thanks for all this work, @WandrilleD! I read through the changes as a whole (e.g. `git diff master..snakemake`), mostly without looking at individual commits, as I found the commits themselves pretty disjointed and full of dead ends. I haven't tried running any of the changed code yet, and I didn't review the Snakefile for correctness.

I've left some line-specific comments, but also have some general thoughts:

* Not all of the GitHub Actions workflows are updated to use the Snakemake workflow, e.g. some of them still reference `ingest-genbank`, which no longer exists.
* Large `shell` blocks in Snakefiles are a red flag or code smell to me, and here I see many of them. This implies to me that either some Snakefile rules are either doing too much, or that what they're doing should be wrapped up into a single external shell program, or both.
* I think that differentiating GISAID vs. GenBank via a `database` wildcard, as opposed to an entirely separate workflow file, will make it harder for their respective workflows to diverge much in the future. This might be seen as a feature, but I think we're actually likely to want to diverge them quite a bit in the future, particularly as we roll up other data sources into the combined "Open" dataset, of which GenBank is only a start.

bin/notify-slack
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-: "${SLACK_TOKEN:?The SLACK_TOKEN environment variable is required.}"
-: "${SLACK_CHANNELS:?The SLACK_CHANNELS environment variable is required.}"
```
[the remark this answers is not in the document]
WandrilleD: The idea of this change is to provide more flexibility to the pipeline, eventually allowing notification of multiple slack channels, as well as making the dependency of certain parts of the code on particular environment variables more explicit (from the standpoint of the snakefile).

snake_config.yaml
WandrilleD: Could you elaborate on why the 2 files idea is the less brittle approach here ? 

My feeling is that having to maintain multiple  files in parallel, sometimes with redundancies (for all common parameters) is more prone to errors. Provided at the moment the only commonality is the fetch argument, so the difference would not be huge, but still.

snakefile
WandrilleD: The idea underlying this choice was to decouple creation of the message from sending the message in order to be more explicit. This being said, I see how this could be "beautified" by wrapping this in more bash scripts.

joverlee521:

PR #175, opened 2021-05-30
Updates to accessions.tsv (57201 new, 635 updated, 1814 removed) seqs downloaded 2021-05-29

AngieHinrichs:
Sorry I didn't get it together last week, but oh well, better late than never I hope.  :)  Lots of changes to slog through since I re-downloaded many 10,000-sequence chunks from GISAID that had grown a little stale.  Still no indication of dot-version numbers in those files.  Thanks!


emmahodcroft:
Thank you Angie!! No worries, whenever you are able to do it, we always appreciate it :) I absolutely know it's not easy to stay on top of these repetitive things especially during busy times!

PR #182, opened 2021-06-16
Updates to accessions.tsv (24026 new, 2 removed) seqs downloaded 2021-06-15

AngieHinrichs:
A couple sequences were withdrawn from GenBank but a couple dozen thousand sequences were added.  :)  Thanks!


emmahodcroft:
Thanks Angie, as always! I hope you guys are enjoying a nice summer over there - summer feels like it has finally arrived in Europe, after a damp and slow spring!

PR #183, opened 2021-06-16
Fix typo in comment

corneliusroemer:
Needs no testing, but should be checked by someone else just for sanity and not to break anything


emmahodcroft:
I imagine this really is a comment and there's no reason not to merge in, if nobody else weighs in differently.

corneliusroemer:
That's what I thought, but I just don't want to make sure I don't break the
internet with my new powers 😉

PR #186, opened 2021-06-22
Fix fetch from GenBank

tsibley:
See commit messages for details.

### DIFF
--- a/bin/csv-to-ndjson
+++ b/bin/csv-to-ndjson
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+"""
+Convert CSV on stdin to NDJSON on stdout.
+"""
+import csv
+import json
+from sys import stdin, stdout
+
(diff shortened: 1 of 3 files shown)


tsibley:
GitHub Action workflows were automatically cancelled when they exceeded the 6h maximum run time of a job. This is strange, because the jobs historically took 1-2h. The AWS Batch job (`a2c3b9cc-ad19-44d0-b988-f1bb0a81258c`) kept running after the GitHub Action workflow detached, and it's timeline (all Pacific) was illuminating:

```
submitted: 2021-06-21 17:14:37
started:   2021-06-21 22:31:27 
finished:  2021-06-21 23:37:24 
```

So the Batch job took only ~1h to run, but waited for over 5h before starting!

Indeed, it uploaded the expected files to `s3://nextstrain-staging/files/ncov/open/branch/fix-fetch-from-genbank/`.

PR #187, opened 2021-06-24
Update rebuild script for GISAID and GenBank builds

jameshadfield:
See commit message for details.

@trs could you review this? I'm sure there are other ways it could be accomplished.

I presume we also wish to remove `trigger rebuild` and replace it with `trigger rebuild-gisaid` and `trigger rebuild-genbank`, which I can attempt if this is indeed the direction to go in.

### DIFF
--- a/.github/workflows/rebuild.yml
+++ b/.github/workflows/rebuild.yml
@@ -17,11 +17,20 @@ jobs:
     - name: setup
       run: ./bin/setup-github-workflow
 
-    - name: rebuild
-      run: ./bin/rebuild
+    - name: rebuild-gisaid-data
+      run: ./bin/rebuild gisaid
       env:
         AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
         SLACK_CHANNELS: ncov-gisaid-updates
+
+    - name: rebuild-genbank-data
+      run: ./bin/rebuild genbank
+      env:
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+        SLACK_CHANNELS: ncov-genbank-updates
--- a/bin/rebuild
+++ b/bin/rebuild
@@ -24,7 +24,7 @@ main() {
       data_source="GenBank (open)"
       profile_name="nextstrain_profiles/nextstrain-open"
     else
-      echo "Invalid source name. Options: 'gisaid', 'genbank' or 'open'."
+      echo "Invalid source name. Options: 'gisaid', 'genbank' or 'open'. 'genbank' and 'open' both run using 'nextstrain-open' profile."
       exit 1
     fi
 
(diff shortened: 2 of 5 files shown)


.github/workflows/rebuild.yml
```diff
@@ -17,8 +17,17 @@ jobs:
     - name: setup
       run: ./bin/setup-github-workflow
 
-    - name: rebuild
-      run: ./bin/rebuild
+    - name: rebuild-gisiad-data
```
huddlej, line 20: `rebuild-gisiad-data` -> `rebuild-gisaid-data`

tsibley:
@jameshadfield Thanks for catching this. I wonder if it'd be better to handle this more like the ncov trial build workflow and split each `nextstrain build` submission into a separate GH Action job within the workflow instead of separate steps within a single job.

As for launch both with `rebuild` vs. launching separately with `rebuild-gisaid` and `rebuild-genbank`, I think separately might be better here. The ingests are separate and on different schedules, and it still seems to be that updates and corrections are made to each ingest separately. This is different from trial builds in ncov, where we decided to always launch both builds together because they share the same workflow code (so if there's a workflow change to trial, we want to trial both).

There's already some precedent for separate launches in ncov-ingest with `gisaid/ingest` and `genbank/ingest`. I can imagine corresponding `gisaid/rebuild` and `genbank/rebuild` as well.

Separately, I'd suggest making the rebuild action "open" instead of "genbank", to match ncov and anticipate that GenBank will be just one source of open data even though it's the only one currently. That is, in the future we anticipate multiple ingests feeding into one "open" build.

.github/workflows/rebuild.yml
```diff
@@ -17,8 +17,17 @@ jobs:
...
+      run: ./bin/rebuild gisaid
+      env:
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+        SLACK_CHANNELS: ncov-gisaid-updates
+
+    - name: rebuild-genbank-data
+      run: ./bin/rebuild genbank
```
tsibley, line 30: GH won't let me comment on the line directly, but the `SLACK_CHANNELS` env for this job step is still `ncov-gisaid-updates` instead of `ncov-genbank-updates`. (And really, we'll want to update it to `ncov-open-updates` in the future, but that requires some additional coordination.)

jameshadfield: Ahh - I noticed this but didn't realise there was a `ncov-genbank-updates` channel. Will update the PR.

jameshadfield:
Thanks @tsibley - I've updated this to make separate actions so that we can trigger each independently (as well as both when needed). These are now the relevant trigger commands:

* `./bin/trigger rebuild` remains and will trigger both rebuilds
* `./bin/trigger gisaid/rebuild`
* `./bin/trigger open/rebuild` (`./bin/trigger genbank/rebuild` will also work, but this may change in the future)

(Note that `./bin/trigger` will use the scripts from the `master` branch, so you can't test this PR this way. Furthermore, if you do run these new workflows, they won't work until https://github.com/nextstrain/ncov/pull/628 is merged.)

cc @eharkins @MoiraZuber

bin/rebuild
```diff
@@ -11,26 +11,25 @@ fi
...
+    local data_source profile_name
+    if [[ $1 == "gisaid" || $1 == "GISAID" ]]; then
+      data_source="GISAID"
+      profile_name="nextstrain_profiles/nextstrain-gisaid"
+    elif [[ $1 == "genbank" || $1 == "GenBank" || $1 == "open" ]]; then
+      data_source="GenBank (open)"
+      profile_name="nextstrain_profiles/nextstrain-open"
     else
-        echo "Rebuilding using remote data files."
+      echo "Invalid source name. Options: 'gisaid', 'genbank' or 'open'."
```
eharkins, line 27:
```suggestion
      echo "Invalid source name. Options: 'gisaid', 'genbank' or 'open'. 'genbank' and 'open' both run using 'nextstrain-open' profile."
```

bin/rebuild
```diff
@@ -39,46 +38,23 @@ main() {
...
-# Returns 1 if both files match (have identical hashes). Else returns 0.
-hashes-match() {
-    local dst_s3path="${2#s3://}"
-    local dst_bucket="${dst_s3path%%/*}"
-    local dst_key="${dst_s3path#*/}"
-
-    local src_hash dst_hash no_hash=0000000000000000000000000000000000000000000000000000000000000000
-    src_hash="$("$bin/sha256sum" < "$1")"
-    dst_hash="$(aws s3api head-object --bucket "$dst_bucket" --key "$dst_key" --query Metadata.sha256sum --output text || echo "$no_hash")"
+    aws_batch_job_id=$(grep "AWS Batch Job ID" <<<"$output" | cut -d ' ' -f 5) || 1
```
eharkins, line 54: What does ` || 1` do here?

eharkins (approved):
Looks good. Can't think of a way to test other than coordinate to have folks on standby if it doesn't work after we merge along with the open profile. I agree with the decision to make the GH actions separate.

bin/rebuild
jameshadfield: Thanks for noticing this! I must have added this unintentionally & have removed it.

eharkins: Can never tell with bash code if it is some clever syntax I don't recognize!

tsibley (approved):
@jameshadfield Looks good. Thanks!

PR #189, opened 2021-07-01
fix: adjust metadata types

rneher:


emmahodcroft:
This is to cut down on the significant digits, right - just converts to ints?

PR #194, opened 2021-07-05
use xz to compress sequences instead of gz

rneher:
Rebase of #169


jameshadfield:
Merging after review in #169 and discussion in slack

PR #195, opened 2021-07-07
genbank ingest automation and workflows;

eharkins:
we want genbank to work like how we have
GISAID in that it fetches new records on
a schedule and otherwise only runs ingest
according to pushes and explicit dispatches.
This should achieve that by mirroring the
implementation of the analogous GISAID workflows
and bash script for ingest

### DIFF
--- a/.github/workflows/fetch-and-ingest-genbank-branch.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-branch.yml
@@ -1,4 +1,4 @@
-name: GenBank ingest (on branch)
+name: GenBank fetch and ingest (on branch)
 
 on:
   push:
@@ -31,7 +31,7 @@ jobs:
           --memory 30GiB \
           --exec env \
           . \
-            envdir env.d ingest-genbank
+            envdir env.d ingest-genbank --fetch
       env:
         AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -1,26 +1,23 @@
-name: GenBank ingest
+name: GenBank fetch and ingest
 
 on:
-  push:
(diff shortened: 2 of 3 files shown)


eharkins:
Canceled workflows though we may want to rerun them to test later. Still need to update the special `-branch` workflows for when the ingest is run on a branch other than `master`

eharkins:
4afa708 mirrors the only GISAID branch workflow with an analogous GenBank one. Do we still want to automatically be doing fetch and ingest rather than just ingest on pushes to branches other than `master`?

emmahodcroft:
How does it work now? (Sorry) I thought fetch was always followed by ingest, but ingests could also be triggered by pushing to files or manually? I imagine mirroring how it works for GISAID would be easiest to keep up with?

emmahodcroft:
Ok sorry - from the convo in Slack I think I understand now. Yes, I think we want to have the same options as for GISAID - fetch+ingest or ingest-only (that's how I understand it at the moment, anyway - please to correct me!)

eharkins:
Thanks @emmahodcroft. https://github.com/nextstrain/ncov-ingest/pull/195#issuecomment-875987619 was referring specifically to the behavior/workflows that happen on non-`master` branches. Current for GISAID (and for GenBank in this PR since I tried to match GISAID behavior) is that when you push any file to a non-`master` branch such as this one, it auto-triggers a `fetch-and-ingest` workflow. I was just wondering if we wanted to keep that specific behavior or update it somehow. 

I guess the intention behind this behavior is to test that ingest still works on branches/PRs, but I wasn't sure if the fetching part of that was something we still thought necessary as part of that testing. I suppose it depends on the PR; I could imagine you'd want to test the full `fetch-and-ingest` workflow especially if your changes affected the fetching part.

emmahodcroft:
Thank you for clarifying Eli! That makes sense to me now. I would agree with you that you might indeed want to test the entire workflow, including fetching - particularly as with `open` we've had some challenges surrounding that already, and there may be new routes to access this data in future, so we'd want to test before switching.

It's a bit odd to me that it does a `fetch` for any file pushed (maybe overkill?), but I've no strong feelings here, and if that's what GISAID does it might be easier to match it! (And maybe there are also reasons for doing this that I don't understand!)

PR #196, opened 2021-07-09
Updates to accessions.tsv (33301 new, 5 removed) seqs downloaded 2021-07-09)

AngieHinrichs:
Congrats on the release of ncov/open!  I see my pipeline is missing some GenBank accessions for COG-UK sequences that you guys are catching -- it looks like my scripts are favoring direct COG-UK download mappings over GenBank, need to fix that.


emmahodcroft:
Thank you as always Angie - we are excited about `open` too!
If it's every helpful to have a dedicated Slack convo or Zoom meeting to chat about matchings, just let us know! We really appreciate your matching, especially now!

PR #198, opened 2021-07-12
feat: use xz for ndjson compression

rneher:
`xz -2 ` is substantially faster and more efficient than gzip on sequence data. this uses `xz` for the ndjson files which contain sequence data.


eharkins:
@rneher should we also change the default args for the two transform scripts?

https://github.com/nextstrain/ncov-ingest/blob/master/bin/transform-genbank#L52
https://github.com/nextstrain/ncov-ingest/blob/master/bin/transform-gisaid#L56

PR #204, opened 2021-08-13
Automatically retry fetches from GISAID and GenBank

tsibley:
Try to recover from transient network/connection failures we sometimes
see.  Most of these are protocol errors or connections unexpected closed
by the remote peer.  We can't use curl's --retry option because it
doesn't apply to the kinds of errors we see and its --retry-all-errors
option would produce duplicate/corrupted data since we process it as a
stream and write to disk.

This retry logic is rather crude but should be fine for now.  At some
point when this pipeline uses a workflow engine like Snakemake,
presumably we can make use of builtin retry logic.

~Try to recover from transient network/connection failures we sometimes
see. Most of these are protocol errors or connections unexpected closed
by the remote peer. The documention is unclear and only mentions
timeouts and HTTP 5xx, but I _think_ curl will treat protocol issues as
transient failures for the purposes of --retry.~

~This will make up to 10 total requests (1 original + up to 9 retries).
curl's default backoff delay time for retries starts at 1s and doubles
for every subsequent attempt: 1 2 4 8 16 32 64 128 256. If this proves
to be unsuitable for our purposes, we can further tune retry behaviour
with curl's other --retry-* options.~

### DIFF
--- a/bin/ingest-genbank
+++ b/bin/ingest-genbank
@@ -68,7 +68,22 @@ main() {
     set -x
 
     if [[ "$fetch" == 1 ]]; then
-        ./bin/fetch-from-genbank > data/genbank.ndjson
+        local attempt=0 max_attempts=5
+
+        while [[ $((++attempt)) -le $max_attempts ]]; do
+            echo "Fetch attempt $attempt"
+            if ./bin/fetch-from-genbank > data/genbank.ndjson; then
+                break
+            else
+                echo "...FAILED"
+                rm data/genbank.ndjson
+                sleep 10
+            fi
+        done
+        if [[ ! -f data/genbank.ndjson ]]; then
+            echo "Failed to fetch"
+            exit 1
+        fi
         if [[ "$branch" == master ]]; then
             ./bin/notify-on-record-change data/genbank.ndjson "$S3_SRC/genbank.ndjson.xz" "GenBank"
         fi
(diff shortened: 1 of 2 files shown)


tsibley:
Ok, so my guess was wrong: curl won't retry with connection closed (and its documentation is precise). I tested this by creating a small HTTP "server" as `/tmp/server`:

```bash
cat <<.
HTTP/1.1 200 Ok
Content-Type: text/plain
Content-Length: 31

.

n=10
while [[ $n -gt 0 ]]; do
    echo $n
    sleep 1
    n=$((n - 1))
done
```

running it with:

```
socat tcp4-listen:4444,reuseaddr,fork,crlf exec:"bash /tmp/server"
```

connecting with curl:

```
curl -fsSv localhost:4444 --retry 5
```

and then Ctrl-Cing the `socat` server to break the connection unexpectedly. curl doesn't retry.

Newer curl versions have a [`--retry-all-errors` option](https://curl.se/docs/manpage.html#--retry-all-errors), but the documentation helpfully explains why it's often not a good idea: data that's already been received and output to stdout will be refetched and re-output, so you'll often end up with duplicated or corrupted data.

Instead, I'll implement the retry outside of curl instead.

emmahodcroft:
This is a great idea Tom, and will be really handy on days where we seem to overlap with  GISAID file update (or other temporary problem).

joverlee521 (approved):

AngieHinrichs:
Thanks so much, I'm going to use this in my scripts that fetch from NCBI and COG-UK!  😃

PR #205, opened 2021-08-21
Updates to accessions.tsv (57911 new, 88 removed) seqs downloaded 2021-08-20

AngieHinrichs:
Many of the changed mappings were caused by newly duplicated sequences in GISAID, from multiple submitting organizations.  Curiously, the sequences already in GISAID had names that were slightly different from their names in GenBank (e.g. CA-CHLA-PLM43021777 vs CA-CHLA-P-PLM43021777, PR-HF101 vs PR-FIOCRUZ-HF101, IN-GD-071321-C21194875 vs IN-GD-SID-C21194875), while the newly duplicated sequences in GISAID have names that exactly match the GenBank names.  I will try to contact the submitters to inquire about the duplicates.

Hope you have a great weekend.  :)


emmahodcroft:
Thank you for digging into this @AngieHinrichs ! Really disappointing to hear that duplications are such an issue. Is there a way to keep track of these? Ideally we would want to exclude duplicates. But this might be getting into a whole nest of thorns.... 😬

AngieHinrichs:
Hi Emma,

I have been excluding duplicates in order to keep the mapping one-to-one.  So far, duplicate submissions have used the same names, so when my pipeline complains about a GenBank sequence matching two GISAID sequences with the same name but different IDs, I usually just exclude the newer IDs.  I've been accumulating a file of those newer IDs (currently 2169 of those, really a small number compared to 3 million), and could share that.  The best thing would be to contact all the submitters to ask them to ask GISAID to remove the duplicated sequences, but... too many things to do.  

This new category is a little more annoying because the names are changing in GISAID, so they're not as easy to detect and work around.  Earlier submissions have names that are similar to GenBank names but not exactly the same.  The new duplicate submissions, which don't seem to be coming from the submitters, use the GenBank names.  Now we have a new GISAID sequence whose name exactly matches the GenBank sequence, so it is favored over the original submission with a name that doesn't exactly match, and not detected as a plain old duplicate.  

Before I send an update of the mapping, I look for previous mappings that have disappeared or changed, just in case there's a bug (or a need to re-download a chunk of GISAID sequences for a name change in both GenBank and GISAID, etc).  That's how I end up finding duplicates.  The numbers are small in the grand scheme of things, but they take time to investigate.  

Thanks,
Angie

emmahodcroft:
Hi Angie, 
Sorry for the delayed reply, but thank you so much for the detailed information and breakdown! Good to know that in the grand scheme it's a small number, but agree with you that it's good to keep a finger on, and we appreciate it!

PR #206, opened 2021-08-29
Updates to accessions.tsv (50367 new, 50 removed) seqs downloaded 2021-08-28

AngieHinrichs:
The 50 removed mappings were again caused by duplicates in GISAID, looking like a glitch on GISAID's end.  I contacted the submitter so they can ask GISAID to fix it.  

More later in response to your earlier question about the duplications in the other PR...

Thanks!


emmahodcroft:
Thanks so much Angie!

PR #207, opened 2021-08-31
Use Nextclade datasets

ivan-aksamentov:

Nextclade data from GitHub is deprecated in favor of datasets since [1.3.0](https://github.com/nextstrain/nextclade/releases/tag/1.3.0).

In  this PR:

 - [x] replace data fetching from GitHub with `dataset get`

 - [x] always fetch the latest release of Nextclade (previously a random version from the ingest container was used. Container releases are not currently synchronized with Nextclade releases)

### Related issue(s)  

None

### Testing

I hope it runs well on CI. Let's see.


rneher (approved):

tsibley (approved):
:+1: for the upgrade.

The switch to always downloading the latest copy seems ok to me, with the caveat that new releases of Nextclade may break ncov-ingest. That is, the switch trades predictability/stability for not having to consciously choose to bump the version in the container. (I don't think it's a trade I'd make in this instance, but I'm fine if you prefer it that way and are willing to fix things when it breaks. :-)

Since the version baked into the container isn't used, would you include a commit removing the relevant lines from the Dockerfile? Good practice to remove code that's no longer used rather than let it accumulate into cruft.

rneher:
re updates: the current version has the same behavior and pulls data from `master`.

I think that is fine. a clade update anyway requires a rerun on the entire file which we haven't automated yet in a satisfactory manner. But now that `ncov-ingest` is running on `aws-batch`, we could have an action that 

 * triggers a rerun on all of gisaid with a specified version
 * modifies the variable `NEXTCLADE_DATA_VERSION` to the new version

this way, all subsequent incremental runs would use the same version automatically. But this is probably better done in a separate PR.

tsibley:
> re updates: the current version has the same behavior and pulls data from `master`.

Ah, I meant updates of the code, not the data. I agree it's probably fine, though.

PR #209, opened 2021-09-06
Updates to accessions.tsv (74308 new, 3787 removed, 1 changed) seqs downloaded 2021-09-06

AngieHinrichs:
Most of the lost sequences were caused by getting IMS-... IDs in BioSample from RKI in Germany that don't bear any resemblance to the GISAID names.  Previously the name matching was a little too loose (accepting just Germany/2021 as close enough) and the new IMS- names cause a name mismatch.  I'll see if I can get a better name mapping from RKI.

I have found some cases of too-loose name matching with identical WA-UW-# sequences and will be changing some of those in a future update (hopefully the next one).


emmahodcroft:
Thank you as always Angie! 
Always strange when the names start to deviate a lot - seems like more work on the side of the submitter. But RKI is sending quite a lot, so we could try to reach out, if you think it would help at all? Either to find out what their pattern is, or to see if they might be able to keep it more consistent? Let us know!

AngieHinrichs:
Sadly, AFAICT there is no pattern with the two sets of RKI IDs -- they are independent sets (IMS-<long hex/text strings> in ENA/INSDC/BioSample and https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland vs. XX-RKI-Y-NNNN in GISAID, where NNNN is a serial number and I don't see any relationship between the NNNN serial number and the long IMS ID strings, although maybe I just haven't looked hard enough 😆).  Tantalizingly, [SARS-CoV-2-Sequenzdaten_Deutschland.csv.xz](https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland/blob/master/SARS-CoV-2-Sequenzdaten_Deutschland.csv.xz) has an alternate ID column, but when it's populated, it turns out to have 64-character hex strings.  🙃  Just what we needed, another set of IDs that are even longer and less human-readable!

I heard back from someone at RKI that they do plan to add the RKI IDs to BioSample, but due to staffing constraints, they won't have time for that until November.  In the meantime, I may just decide that when there is a 1-to-1 sequence identity with IMS- and -RKI- (as opposed to a set of identical sequences with multiple IMS-'s and multiple RKI's), they are the same, because there are ~150k RKI's and ~130k IMS's, so probably ~100k of those unambiguous matches.

PR #212, opened 2021-09-24
Trigger preprocessing pipelines after fetch-and-ingest

jameshadfield:
Our intention is to have the preprocessing (alignment) pipeline
run following fetch-and-ingest, which currently happens daily.

The API call is similar to that in `./bin/trigger` and follows [these docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api)

### Testing
???

### DIFF
--- a/bin/ingest-gisaid
+++ b/bin/ingest-gisaid
@@ -96,6 +96,9 @@ main() {
         --output-fasta data/gisaid/sequences.fasta \
         --output-unix-newline > "$flagged_annotations"
 
+    # Remove gisaid.ndjson to save disk space.
+    rm data/gisaid.ndjson
+
     # Download old clades
     (   aws s3 cp --no-progress "$S3_DST/nextclade.tsv.gz" - \
      || aws s3 cp --no-progress "$S3_SRC/nextclade.tsv.gz" -) \
--- a/bin/ingest-gisaid
+++ b/bin/ingest-gisaid
@@ -157,6 +157,17 @@ main() {
     ./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/sequences.fasta "$S3_DST/sequences.fasta.xz"
 
     ./bin/clean
+
+    if [[ "$fetch" == 1 && "$branch" == master ]]; then
+        echo "Triggering ncov preprocessing GitHub action via repository dispatch."
+        curl -fsS https://api.github.com/repos/nextstrain/ncov/dispatches \
+          -H 'Accept: application/vnd.github.v3+json' \
+          -H 'Content-Type: application/json' \
+          -H "authorization: Bearer ${PAT_GITHUB_DISPATCH}" \
+          -d '{"event_type":"preprocess-gisaid"}'
+    else
+        echo "Skipping running of ncov preprocessing workflow as the current branch is not master or we are running an ingest only (not a fetch-and-ingest)."
+    fi
 }
 
 print-help() {
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -31,7 +31,8 @@ jobs:
           AWS_DEFAULT_REGION \
           GITHUB_REF \
           SLACK_TOKEN \
-          SLACK_CHANNELS
+          SLACK_CHANNELS \
+          PAT_GITHUB_DISPATCH
 
         nextstrain build \
           --aws-batch \
@@ -48,3 +49,4 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
         SLACK_CHANNELS: ncov-genbank-updates
+        PAT_GITHUB_DISPATCH: ${{ secrets.PAT_GITHUB_DISPATCH }}
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -32,7 +32,8 @@ jobs:
           GISAID_USERNAME_AND_PASSWORD \
           GITHUB_REF \
           SLACK_TOKEN \
-          SLACK_CHANNELS
+          SLACK_CHANNELS \
+          PAT_GITHUB_DISPATCH
 
         nextstrain build \
           --aws-batch \
@@ -51,3 +52,4 @@ jobs:
         GISAID_USERNAME_AND_PASSWORD: ${{ secrets.GISAID_USERNAME_AND_PASSWORD }}
         SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
         SLACK_CHANNELS: ncov-gisaid-updates
+        PAT_GITHUB_DISPATCH: ${{ secrets.PAT_GITHUB_DISPATCH }}
--- a/bin/ingest-genbank
+++ b/bin/ingest-genbank
@@ -170,6 +170,17 @@ main() {
     ./bin/upload-to-s3 ${silent:+--quiet} data/genbank/duplicate_biosample.txt "$S3_DST/duplicate_biosample.txt.gz"
 
     ./bin/clean
+
+    if [[ "$fetch" == 1 && "$branch" == master ]]; then
+        echo "Triggering ncov preprocessing GitHub action via repository dispatch."
+        curl -fsS https://api.github.com/repos/nextstrain/ncov/dispatches \
+          -H 'Accept: application/vnd.github.v3+json' \
+          -H 'Content-Type: application/json' \
+          -H "authorization: Bearer ${PAT_GITHUB_DISPATCH}" \
+          -d '{"event_type":"preprocess-open"}'
+    else
+        echo "Skipping running of ncov preprocessing workflow as the current branch is not master or we are running an ingest only (not a fetch-and-ingest)."
+    fi
 }
 
 print-help() {


eharkins (commented):
> The API call is similar to that in ./bin/trigger and follows these docs

Looking at both of those examples, it seems like this will work to trigger the workflows in https://github.com/nextstrain/ncov/pull/730. 👍 

>Testing
> ???

Ways I can think of to test:
- [ ] run the `curl` command from the command line to test that it works as expected
- [ ] run the whole `fetch-and-ingest` workflow by pushing to this branch to test that the curl triggering of the workflow in ncov works from the context of the workflow in this PR. Seems like https://github.com/nextstrain/ncov-ingest/runs/3694985696?check_suite_focus=true might have already done that? Hard to tell if the `trigger-preprocessing-pipeline` job step was successful without going over to the ncov repo because curl is running in silent mode. Maybe we want to change this to have it print out whether the request succeeded or not?

jameshadfield:
Thanks @eharkins. The actions in github/ncov will only respond to repository dispatch events once they are merged into master, so we won't be able to test those until that PR gets merged, but I agree that as soon as that happens we can run a local `curl` command before merging this PR. 

> Seems like https://github.com/nextstrain/ncov-ingest/runs/3694985696?check_suite_focus=true might have already done that? 

Even though that action is running on this branch ("trigger-preprocess") my understanding is that the action itself is the one from the master branch, not this PR branch. Note that that action didn't have a "trigger-preprocessing-pipeline" step in it. (I admit to finding GitHub actions confusing, so I'm not entirely confident about this!)

eharkins:
Ah that makes sense, thanks @jameshadfield. I think you can get it to run on this branch by 

A. Making the same changes to the `fetch-and-ingest-{gisaid,genbank}-branch.yml` workflows e.g. https://github.com/nextstrain/ncov-ingest/blob/master/.github/workflows/fetch-and-ingest-gisaid-branch.yml. These are the ones that ran automatically with pushes to this PR since the `fetch-and-ingest-{gisaid,genbank}-master.yml` ones only run when you push to master.

B. Alternatively, you could change the master workflows you are working on in this PR to have just:
```
on: push
```
which will run on any branch. Then change it back when done testing before you merge.

I think A makes more sense and needs to happen anyway since I think one purpose of the `fetch-and-ingest-{gisaid,genbank}-branch.yml` workflows is this exact case where we want to use them as a proxy for the ones that run on `master` without interfering with live data on s3. It does end up meaning we have a lot of duplicate code between the `fetch-and-ingest-{gisaid,genbank}-branch.yml` and `fetch-and-ingest-{gisaid,genbank}-master.yml` workflows, but having them do (almost) the exact same thing is the precedent, so I think we should keep it for now at least.

tsibley (commented):
Since the ingest jobs run on AWS, I'd think about putting these API calls into `ingest-gisaid` and `ingest-genbank` instead as optional/conditional final steps, probably both calling out to the same small program that wraps the `curl` invocation (similar to `bin/trigger`). This ensures that they happen if the AWS job succeeds even if the GitHub job fails for some reason.

Also, if it's your personal access token in the new `PAT_GITHUB_DISPATCH` secret, I'd suggest making it a token associated with the @nextstrain-bot account instead.

eharkins:
> Since the ingest jobs run on AWS, I'd think about putting these API calls into ingest-gisaid and ingest-genbank instead as optional/conditional final steps, probably both calling out to the same small program that wraps the curl invocation (similar to bin/trigger). This ensures that they happen if the AWS job succeeds even if the GitHub job fails for some reason.

Good point. The `fetch-and-ingest` process has been taking increasingly long so this helps too if it exceeds the GH actions job time limit.

jameshadfield:
> Since the ingest jobs run on AWS, I'd think about putting these API calls into ingest-gisaid and ingest-genbank instead as optional/conditional final steps

Done ✅ 

> I'd suggest making it a token associated with the @nextstrain-bot account instead.

Done ✅ 

I've tested the repository dispatch calls (and the token) from my machine using [these test actions](https://github.com/nextstrain/test-github-actions/actions/workflows/preprocess-open.yml).

On a call with @eharkins and @corneliusroemer we considered allowing `*-branch.yml` actions to trigger a preprocessing action which would use the branch-specific sequences/metadata but decided this was not worth the increase in complexity for a rare use case.

jameshadfield:
It looks like the [GenBank fetch and ingest (on branch)](https://github.com/nextstrain/ncov-ingest/actions/runs/1294100742) succeeded, but the [GISAID fetch and ingest (on branch)](https://github.com/nextstrain/ncov-ingest/actions/runs/1294100743) failed & it's not clear to me why exactly; [here's](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=%2Faws%2Fbatch%2Fjob;stream=nextstrain-job_21_nextstrain-ncov-ingest_latest%2Fdefault%2Fa73ab2cbea6c4590a90aa9f200e0394b) the logs.

eharkins:
> It looks like the GenBank fetch and ingest (on branch) succeeded, but the GISAID fetch and ingest (on branch) failed & it's not clear to me why exactly; here's the logs.

I'm guessing it's unrelated to your changes (hence the fact it didn't also fail genbank) - sometimes the GISAID ingest process has been running out of memory and the error code (137) is the same for the failure you linked to:
https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1632762577171800?thread_ts=1632734001.170500&cid=CTZKJC7PZ 

I'll create an issue for that in this repo.

eharkins:
Rebased again to get https://github.com/nextstrain/ncov-ingest/commit/92fb9e414de567f3080eed6908d10c93ce7ab98d which should address the cause of the issue with the most recent run of the GISAID fetch-and-ingest on this branch (the issue was also in `master`, so I rebased to get the fix from `master`).

eharkins:
Do we want to trigger preprocessing workflows every time we push changes like https://github.com/nextstrain/ncov-ingest/commit/e618256f20f9756c8bbd058536927d480bed8a29 [as we do now](https://github.com/nextstrain/ncov-ingest/pull/212/files#diff-a451bc1b08f1b99861ea1baff2673bdabaae9159fb49fe99db6024cc4c2f8aa7R161) / as we would if this were merged in its current state?

If not, we could add some logic e.g. here https://github.com/nextstrain/ncov-ingest/pull/212/files#diff-a451bc1b08f1b99861ea1baff2673bdabaae9159fb49fe99db6024cc4c2f8aa7R161 to only trigger preprocessing workflows when fetching from GISAD/Genbank. The [`fetch` option that is true if we are fetching from GISAID/Genbank](https://github.com/nextstrain/ncov-ingest/blob/master/bin/ingest-gisaid#L67) could be used to only trigger preprocessing in that scenario.

jameshadfield:
> Do we want to trigger preprocessing workflows every time we push changes like e618256 as we do now / as we would if this were merged in its current state?

The intention was to only run preprocessing on a fetch-and-ingest, however when I moved them from the specific GitHub action into the bash script (as suggested) I forgot to add such a conditional. This should be changed.

bin/ingest-gisaid
```diff
@@ -157,6 +157,17 @@ main() {
     ./bin/upload-to-s3 ${silent:+--quiet} data/gisaid/sequences.fasta "$S3_DST/sequences.fasta.xz"
 
     ./bin/clean
+
+    if [[ $fetch == 1 && "$branch" == master ]]; then
```
eharkins, line 161: Same question here if you maybe need to add quotes to match other logic referencing variables?
```suggestion
    if [[ "$fetch" == 1 && "$branch" == master ]]; then
```

eharkins (commented):
Looks good, might need quotes around the fetch variable in the last change. Though perhaps not according to [this answer on SO](https://stackoverflow.com/questions/10067266/when-should-i-wrap-quotes-around-a-shell-variable#:~:text=General%20rule%3A%20quote%20it%20if,a%20single%20argument%20into%20many.). Can't hurt I don't think.

jameshadfield:
> Looks good, might need quotes around the fetch variable in the last change. 

They're not technically needed, as we know the value of `$fetch` doesn't have whitespace in it and isn't empty. I've added them here since we quote it earlier in the script so it's nicer to be consistent (and safer).

jameshadfield:
Unfortunately the GISAID trial ingest, when running on this branch, is [still failing](https://github.com/nextstrain/ncov-ingest/runs/3786281991?check_suite_focus=true) with exit code 14. This is despite being rebased and therefore including the patch introduced in 92fb9e414de567f3080eed6908d10c93ce7ab98d which [seems to have fixed this on master.](https://github.com/nextstrain/ncov-ingest/actions/runs/1304580258). @joverlee521 @eharkins any ideas?

joverlee521:
> Unfortunately the GISAID trial ingest, when running on this branch, is [still failing](https://github.com/nextstrain/ncov-ingest/runs/3786281991?check_suite_focus=true) with exit code 14. This is despite being rebased and therefore including the patch introduced in [92fb9e4](https://github.com/nextstrain/ncov-ingest/commit/92fb9e414de567f3080eed6908d10c93ce7ab98d) which [seems to have fixed this on master.](https://github.com/nextstrain/ncov-ingest/actions/runs/1304580258). @joverlee521 @eharkins any ideas?

The latest run exited with a [different error in the pipeline](https://github.com/nextstrain/ncov-ingest/runs/3797878162?check_suite_focus=true#step:4:276) that's a memory issue unrelated to the patch. It's unclear to me why there is a memory issue on this branch but not on master since @eharkins [bumped the memory for both](https://github.com/nextstrain/ncov-ingest/commit/5db5d2574210d60b4ae6434248af64dd3187a781).

PR #214, opened 2021-10-01
Full Nextclade runs

ivan-aksamentov:

This PR adds `--nextclade_full_run` flag for `ingest-gisaid` and `ingest-genbank` scripts as well as a pair of GitHub actions. This flag is needed to trigger full run of Nextclade on all available sequences.

In current implementation, clade results (and other metrics) are cached in `nextclade.tsv` file. Upon rerun, the sequences in `sequences.fasta` and  `nextclade.tsv` files are compared and only sequences that are missing in `nextclade.tsv` undergo analysis with Nextclade. The results are then appended to `nextclade.tsv`. Then `nextclade.tsv` gets merged with `metadata.tsv` and that's how clades and some of the other Nextclade metrics appear in the medatata.

With the new flag `--nextclade_full_run` provided, the `nextclade.tsv` (cached clades from the previous run) does not get downloaded. So the comparison always gives a list of all sequences and Nextclade processes them all again, and metadata table gets refreshed clades and metrics for all rows.

This refresh is necessary every time when new clades are defined or when upgrading to the new version of Nextclade with breaking changes.

Previously we used to perform these runs manually on external resources, but now we are thinking about automating this.

The trigger should probably still be manual, because not every Nextclade release needs the full run.

Depending on the manual data curation flow, these runs can be performed on master or on a separate branch. If on separate branch, the resulting `metadata.tsv` and `nextclade.tsv` will end up in the `branch/` subdirectory and should be copied manually to the location when the next ingest from master can find them.

Another challenge arises when we want to perform this full run before the required Nextclade version is released. Nextclade currently does not publishes pre-releases of CLI, but that can be worked out. Although automated C++ builds from master branch might be too costly in terms of CI credits. I currently experiment with the pre-released binaries in #215, branch [nextclade-full-run-2021-09-30](https://github.com/nextstrain/ncov-ingest/commits/nextclade-full-run-2021-09-30)

### Related issue(s)  

 - #215 

### Testing

Let's experiment a little in [#215] and see how well it goes :) 

We need to make sure the workflow for full reruns is convenient and the results are correct (that is that the full run actually happens and clades get recomputed).

### DIFF
--- a/.github/workflows/ingest-full-genbank.yml
+++ b/.github/workflows/ingest-full-genbank.yml
@@ -1,0 +1,47 @@
+name: GenBank ingest with full Nextclade run
+
+on:
+  # Manually triggered using `./bin/trigger genbank/ingest-full`
+  # (or `ingest-full`, which includes GISAID)
+  repository_dispatch:
+    types:
+      - genbank/ingest-full
+      - ingest-full
+
+  # Manually triggered using GitHub's UI
+  workflow_dispatch:
+
+jobs:
+  ingest-genbank-with-full-nextclade-run:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: 'nextclade-full-run-2021-09-30'
+
+    - name: setup
+      run: ./bin/setup-github-workflow
+
+    - name: ingest-genbank-with-full-nextclade-run
+      run: |
+        ./bin/write-envdir env.d \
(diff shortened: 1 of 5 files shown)


ivan-aksamentov:
I realized that we don't need to mix the usual ingest flow with Nextclade full run. It can be made separate. The new version is here: https://github.com/nextstrain/ncov-ingest/pull/218/files

PR #215, opened 2021-10-01
Experiment: Automation of full Nextclade runs

ivan-aksamentov:

I am experimenting with possible implementations of automation of full Nextclade reruns in ingest.

This branch combines #214 as well as the pre-release Nextclade binary ([`f5a1ef8` (#215)](https://github.com/nextstrain/ncov-ingest/pull/215/commits/f5a1ef814aa81903528dcb834506e01adc3844da), this bit still needs some thought).

And while we are on it, I also added:
 - #168 
 - #213 

to the mix. These features require full reruns, so if this experiment succeeds, we can add them in one go.

### Related issue(s)  

 - #214 

### Testing

Let's run it and see if something breaks :boom:


ivan-aksamentov:
I realized that we don't need to mix the usual ingest flow with Nextclade full run. It can be made separate. The new version is here: https://github.com/nextstrain/ncov-ingest/pull/218/files

PR #218, opened 2021-10-02
feat: add script to make full nextclade parallel run

ivan-aksamentov:
 
This PR adds script `./bin/run-nextclade-full` and a set of helper scripts, to trigger full run of Nextclade: all available sequences will be processed again. The goal here is to automate, at least partially, the tedious task of performing these full runs.

The changes in this PR should not affect the normal daily flow and are just helpers for the full reruns that are performed occasionally.

#### Previously

In the routine daily ingests only the new sequences get processed by Nextclade. Clade results (and other metrics) from previous runs are cached in `nextclade.tsv` file. Upon rerun, the sequences in `sequences.fasta` and `nextclade.tsv` files are compared and only sequences that are missing in `nextclade.tsv` undergo analysis with Nextclade. The results are then appended to `nextclade.tsv`. Then nextclade.tsv gets merged with metadata.tsv and that's how clades and some of the other Nextclade metrics appear in the medatata file.

However, the cache needs to be invalidated when
 - clades names or definitions change
 - we want to upgrade Nextclade to a new version that has breaking changes 
 - we want to change what columns we add to metadata
a full rerun on all sequences is needed in order to refresh the metadata table.

It takes a while to run it, and back in the day, when ingest ran purely on GitHub Actions, @rneher was running these full runs on an external cluster. It's a tedious task, so my goal here is to automate the flow, at least partially. Especially now that ingest is on AWS.

#### In this PR

The following scripts are introduced:

 - `./bin/run-nextclade-full` - runs locally (not feasible computationally for most people). For development and debugging. The script accepts `--database=gisaid|genbank` flag to set the input and output sources on S3. It downloads currently available `sequence.fasta` (from daily workflow). It splits this file into batches and runs several instances of Nextclade in parallel on a single batch to achieve better resource utilization. The batch size and number of threads per Nextclade instance is configurable. Nextclade instances are allocated a certain number of threads and they wait until previous instances finish to not cause too much oversubscription. A small overlap is given to oversubscribe slightly for better resource use.

 - `./bin/run-nextclade-full-docker` - runs `./bin/run-nextclade-full` in a Docker container (still not feasible computationally for most people). For more convenient development and debugging.

 - `./bin/run-nextclade-full-aws` - schedules `./bin/run-nextclade-full` on AWS Batch. There are additional flags to set number of requested CPUs and memory from AWS Batch.

 - `./bin/split-fasta` - brought back from the olden days: 9bce646534ac436a414d31ba7126135a446bf0c4. It was removed in 94b0372d0e9afbe2c7668c5af92f034f541d98a5 when Nextclade became multithreaded. This scrip is what splits sequences into batches.

##### Intended usage

When it's decided that a full run is needed, a dedicated person should run:

```bash
cd ncov-ingest
./bin/run-nextclade-full-aws --database=gisaid
./bin/run-nextclade-full-aws --database=genbank
```
(making sure AWS credentials are available for the script)

The resulting `nextclade.tsv.gz` should be then available in the subdirectory `nextclade-full-run` in the usual S3 location for the particular database:

```bash
s3://nextstrain-ncov-private/nextclade-full-run/nextclade.tsv.gz
s3://nextstrain-data/files/ncov/open/nextclade-full-run/nextclade.tsv.gz
```

The `nextclade.tsv.gz` should be manually inspected for scientific correctness, comparing to the old one, if necessary. If all went well, after agreeing with ingest team, the `nextclade.tsv.gz` should then be copied to the location where the daily ingest can find it, overwriting the old one. These locations are:

```
s3://nextstrain-ncov-private/nextclade.tsv.gz
s3://nextstrain-data/files/ncov/open/nextclade.tsv.gz
```
(just omit the subdirectory `nextclade-full-run`)

The next ingest will then include the freshly computed clades. We might try to automate this step as well. However, given that manual inspection is needed, the automation options are limited.

### Related issue(s)  

None

### Testing

The examples of AWS Batch jobs:

GISAID:
```
nextstrain build --aws-batch --no-download --attach 0591961b-fb72-4f18-b55f-38e8f8c9f20d .
```

GenBank:
```
nextstrain build --aws-batch --no-download --attach 20698f2c-69f1-4302-8df0-3b300ea69411 .
```

We need to make sure the workflow for full reruns is convenient and the results are correct (that is that the full run actually happens and clades get recomputed).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,10 @@
 FROM nextstrain/base:branch-python-base
 
 # Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN apt-get --allow-releaseinfo-change-suite update && apt-get install -y --no-install-recommends \
         python3-netifaces \
         time\
+        csvkit \
         xz-utils
 
 # Install Python deps
@@ -41,15 +42,15 @@ RUN python3 -m pip install pipenv
 COPY Pipfile Pipfile.lock /nextstrain/ncov-ingest/
 RUN PIPENV_PIPFILE=/nextstrain/ncov-ingest/Pipfile pipenv sync --system
 
-# Install Nextclade
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
-         -o /usr/local/bin/nextclade \
- && chmod a+rx /usr/local/bin/nextclade
-
-# Install Nextclade C++
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
-         -o /usr/local/bin/nextclade \
- && chmod a+rx /usr/local/bin/nextclade
+## Install Nextclade
+#RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
+#         -o /usr/local/bin/nextclade \
+# && chmod a+rx /usr/local/bin/nextclade
+#
+## Install Nextclade C++
+#RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
+#         -o /usr/local/bin/nextclade \
+# && chmod a+rx /usr/local/bin/nextclade
 
 # Put any bin/ dir in the cwd on the path for more convenient invocation of
 # ncov-ingest's programs.
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,9 @@
 FROM nextstrain/base:branch-python-base
 
 # Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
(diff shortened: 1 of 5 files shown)


ivan-aksamentov:
Richard and I checked the outputs and all looks good. This does not affect the main workflow, so it's low rick. I am going to merge.

corneliusroemer:
Quick note for what to do after full run is triggered:

The `nextclade.tsv.gz` should be manually inspected for scientific correctness, comparing to the old one, if necessary.

If all went well, after agreeing with ingest team, the **entire directory** (`nextclade.tsv.gz` and also `aligned.fasta.xz`) should be copied to the location where the daily ingest can find it, overwriting the old one. These locations are:

```
s3://nextstrain-ncov-private/nextclade.tsv.gz
s3://nextstrain-data/files/ncov/open/nextclade.tsv.gz
```
(just omit the subdirectory `nextclade-full-run`)

It's important to copy `aligned.fasta.xz` as well because otherwise there will be duplicated sequences (see #281)

Make sure you copy when ncov-ingest is **not** running (you can check in Slack). Good time to copy is 1-5pm Basel time.

PR #219, opened 2021-10-02
perf: use multithreaded mode of xz

ivan-aksamentov:

Should make things a bit faster. Not sure if there are any disadvantages, but nothing comes to mind.

Threaded decompression has not been implemented yet, but compression should work. We could use [pixz](https://github.com/vasi/pixz) for threaded decompression as well.

From `man xz`:

```
-T threads, --threads=threads
       Specify the number of worker threads to use.  Setting threads to a special value 0 makes xz use as many
       threads as there are CPU cores on the system.  The actual number of threads can be less than threads if
       the input file is not big enough for threading with the given settings or if using more  threads  would
       exceed the memory usage limit.

       Currently  the  only threading method is to split the input into blocks and compress them independently
       from each other.  The default block size depends on the compression level and can be overriden with the
       --block-size=size option.

       Threaded  decompression  hasn't been implemented yet.  It will only work on files that contain multiple
       blocks with size information in block headers.  All files compressed in multi-threaded mode  meet  this
       condition, but files compressed in single-threaded mode don't even if --block-size=size is used.

```

Also:

 - https://stackoverflow.com/a/33441796
 - https://superuser.com/a/897325
 - https://git.tukaani.org/?p=xz.git;a=blob_plain;f=NEWS;hb=HEAD

### Related issue(s)  

 - #194 
 - #169 
 - #198 

### Testing

:boom:

### DIFF
--- a/bin/download-from-s3
+++ b/bin/download-from-s3
@@ -22,7 +22,7 @@ main() {
         if [[ "$src" == *.gz ]]; then
             gunzip -cfq
         elif  [[ "$src" == *.xz ]]; then
-            xz -dcq
+            xz -T0 -dcq
         else
             cat
         fi > "$dst"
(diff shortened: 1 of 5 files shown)


tsibley:
Seems like a no-brainer to enable this. I do wonder what the impact is on time, though. Did you compare average job runtimes before/after?

ivan-aksamentov:
Nope, did not have a chance yet. It's not very straightforward to do with the current setup: it takes more time to zip the working directory at the end than to run the actual work.

PR #225, opened 2021-10-27
check-locations: remove `validate_metadata()`

joverlee521:
Empty string location fields should not break the ncov build according
to the [SARS-CoV-2 workflow docs](https://docs.nextstrain.org/projects/ncov/en/latest/analysis/data-prep.html#missing-metadata):
"Missing data is to be expected for certain fields. In general, missing
data is represented by an empty string or a question mark character."

Since we are expecting the ncov-ingest/ncov build pipeline to be
automated, remove this assertion that requires manual curation to fix.
The empty string locations should still get flagged as new location
that can be manually curated at a later time.

### DIFF
--- a/bin/check-locations
+++ b/bin/check-locations
@@ -5,7 +5,6 @@ scripts. If the validations succeed, saves a location hierarchy from the
 given metadata to a local file.
 """
 import sys
-import copy
 import argparse
 import pandas as pd
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


joverlee521:
> > Would be good to keep an eye on the resulting ncov builds after this change lands so we can check that the final phylogenomic outputs handled the missing location data (particularly region, country, division) reasonably. (Could also proactively test this, of course.)
> 
> Yup, I purposely excluded [3646792](https://github.com/nextstrain/ncov-ingest/commit/3646792f21c0e723c6a8ecefcbe8cb76755e09f7) from this PR to ensure the test run will have missing location data and then run a trial build with the result of the test run.

Well, that didn't work as I had planned. The test build ran with a new fetch from GISAID that included the full location data, so there were no missing data. 

Instead, I locally ran a CI build after manually removing the location data for `Australia/VIC05/2020` in `data/example_metadata.tsv`. The CI build was able to run to completion without any errors and I pushed the results to [staging](https://nextstrain.org/staging/ncov/gisaid/trial/remove/location/assertion/europe). `Australia/VIC05/2020` shows up as a default grey color when using any of the location color-bys and the legend does not include any extra "blank" locations.

PR #227, opened 2021-11-03
Require aiobotocore<2 when installing nextstrain-cli

tsibley:
Works around an upstream dependency issue where Pip ends up installing
the latest aiobotocore with an older s3fs, which aren't compatible.¹
I'm not sure if the fault lies with Pip's resolution of dependencies or
with the dependency declarations in aiobotocore/s3fs/fsspec/boto3/…
themselves.

Jover's diagnosis of the issue, which I've confirmed and reproduced
locally², is:

> For some reason multiple versions of s3fs were downloaded in the setup
> step. It finally installs version 2021.7.0 of s3fs, which does not have
> the fix to support aiobotocore >= 1.4.0.

Hopefully this workaround can be temporary and removed soon: either
upstream will sort things out or we'll bake the workaround into a new
nextstrain-cli release (since currently all new installs of
nextstrain-cli will not work).

¹ Throws an «AttributeError: module 'aiobotocore' has no attribute
  'AioSession'» during upload of the workdir to S3, causing cascading
  failures.

² mkdir /tmp/venv
  python3 -m venv /tmp/venv
  source /tmp/venv/bin/activate
  pip install --upgrade pip setuptools wheel
  pip install nextstrain-cli
  nextstrain build --aws-batch zika-tutorial

### DIFF
--- a/bin/setup-github-workflow
+++ b/bin/setup-github-workflow
@@ -9,5 +9,5 @@ python3 -m pip install --upgrade pip setuptools wheel
 echo "::endgroup::"
 
 echo "::group::Installing nextstrain-cli"
-python3 -m pip install 'nextstrain-cli >=3.0.1'
+python3 -m pip install 'nextstrain-cli >=3.0.1' 'aiobotocore<2'
 echo "::endgroup::"


tsibley:
[Changelog for aiobotocore](https://github.com/aio-libs/aiobotocore/blob/d55cc528fafe0f506b10e74fd945c17007774984/CHANGES.rst#200-2021-11-02) suggests we could set `AIOBOTOCORE_DEPRECATED_1_4_0_APIS=1` and it'd continue to work. s3fs version resolution would still be weird.

joverlee521:
> [Changelog for aiobotocore](https://github.com/aio-libs/aiobotocore/blob/d55cc528fafe0f506b10e74fd945c17007774984/CHANGES.rst#200-2021-11-02) suggests we could set `AIOBOTOCORE_DEPRECATED_1_4_0_APIS=1` and it'd continue to work without the workaround in this PR. s3fs version resolution would still be weird.
> 
> Update: Confirmed this works. Not sure it's a better workaround, though?

I have a slight preference for the PR workaround since you've made the nice commit message to document this is a temporary workaround 😄

joverlee521 (approved):

tsibley:
I dug into the dep declarations of the various packages.

nextstrain-cli declares deps on fsspec, s3fs, and aiobotocore.

Initially, pip plans ([full gory log](https://github.com/nextstrain/ncov-ingest/files/7469720/log.txt)) to fulfill those with the latest versions of each, currently fsspec 2021.10.1, s3fs 2021.10.1, and aiobotocore 2.0.0.

s3fs 2021.10.1 [declares `aiobotocore~=1.4.0`](https://github.com/dask/s3fs/commit/90bd7bd6c56af89a93c1f0692463bf67e776104e)[^1], however, which precludes aiobotocore 2.0.0. At this point, it seems pip starts methodically trying previous versions of s3fs to see if it can resolve the dep conflict. It tries 2021.10.0, 2021.9.0, 2021.8.1, and 2021.8.0, all of which declare `aiobotocore~=1.4.0`, until it gets to 2021.7.0, which declares `aiobotocore>=1.0.1`. This is satisfied by aiobotocore 2.0.0, so pip is happy and updates the plan to use s3fs 2021.7.0.

Since s3fs pins exact fsspec versions ([for example](https://github.com/dask/s3fs/commit/9a0ea4bcc41dbf92df242a6c75363005240e50d2)) and releases in lockstep with fsspec, pip then updates the plan to use fsspec 2021.7.0 as well.

The dep resolution plan is satisfied and pip installs everything. Unfortunately, the runtime AttributeError for AioSession happens because s3fs 2021.7.0 isn't actually compatible with `aiobotocore>=1.0.1` as declared, because aiobotocore 2.0.0 legitimately introduces breaking changes.

Pip's dependency resolution seems prone to stumble into this sort of issue, as dependency declarations which are only lower-bounded (e.g. `>=`) can only really be considered valid at the time they're released. After that point, new releases of the dependencies may render the declaration invalid. At the same time, one could say that s3fs should have declared `aiobotocore>=1.0.1,<2` from the start, to exclude legitimate breaking changes in a new major version.

Dependencies are hard.

[^1]: This declaration was [a reaction](https://github.com/dask/s3fs/pull/510) to aiobotocore's mistake of [releasing 1.4.0](https://github.com/aio-libs/aiobotocore/blob/d55cc528fafe0f506b10e74fd945c17007774984/CHANGES.rst#140-2021-08-20) with breaking changes. aiobotocore addressed the mistake 4 days later by [releasing 1.4.1](https://github.com/aio-libs/aiobotocore/blob/d55cc528fafe0f506b10e74fd945c17007774984/CHANGES.rst#141-2021-08-24), but s3fs has already reacted by then.

PR #231, opened 2021-11-22
Snakemake pipeline

jameshadfield:
The aim of this PR is to make an unchanged version of the pipeline which we can then iterate to, for instance, improve slack notifications, implement cached alignments or combine with nextstrain/ncov. A key aim was to do this with minimal (if any) changes to the files in `./bin/`; these can come in subsequent PRs, as desired. Similarly, the existing GitHub actions should behave as before. 

There is a large degree of overlap between the steps we run on each database. Placing both pipelines into the same snakefile allows understanding of the common code paths. Where the steps have diverged significantly, I’ve used different rules instead of overloading one rule with conditional logic. See the DAGs in the figure below.

The bash script contained a lot of conditional logic - for instance, if the GitHub branch was “master” then certain steps happened. I’ve chosen to make this explicit in the snakemake config and set by the calling GitHub actions. 

If you wish to run this locally, the easiest may be to look at the commands used in the GitHub actions. I haven’t changed the “Running locally” section of the README as this calls the scripts themselves, rather than the runner scripts which the Snakefile replaces. For what it’s worth, I think these are out of date and need to be updated, but that’s for another PR.

**Changes in behavior**

I’ve tried to minimise these, but made a few changes during implementation:

* The pipeline previously couldn’t work when both uploading to a non-master location and downloading the ndjson file (i.e. not fetching from the database). This was due to the download command always using `S3_DST`. This is changed here to fallback to `S3_SRC` if needed, which  mirrors the behaviour for downloading nextclade assets. Since we don’t have an action of `ingest-from-branch` this fallback is not used in practice.
* `rule annotate_via_nextclade` uses conditional logic to check if there are new sequences for Nextclade to process. This was previously only used for GISAID, but is used for both databases here.
* The GitHub actions now detach immediately (see below).

**Review items to focus on / questions:**

* The snakemake pipeline is prone to logging environment variables (e.g. a Snakemake `CalledProcessError` may log all the env variables which are prepended to each command). For this reason I always run the job on AWS and detach from it in GitHub actions.
* Snakemake allows `nextclade` 8 threads, which is the same as we use for `nextalign` in the nextstrain/ncov pipeline. The nextclade docs state that “the algorithm will use all the available threads” so we don’t need to specify this via `--jobs` however I haven’t profiled this.
* There are lots of hardcoded assumptions — for instance, we only send slack notifications when running on master, and thus we can use `S3_SRC` rather than `S3_DST` in the slack messages. This is in line with previous behaviour,  but should be changed as we update this pipeline in further PRs. 
* We use Snakemake’s `temp()` to manage removal of large files rather than explicit `rm ...` calls. This may need to be improved if we run out of disk space.

**Testing**

Tested most steps locally (on subsets of the data), and currently running `fetch-and-ingest-{database}-branch` actions here: [GISAID](https://console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/0f11c607-3c4d-4794-8f96-68bb17043fc0) and [GenBank](https://console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/b39e979a-7c68-45d8-b0cb-76f07409eb49).

image


tsibley:
I haven't reviewed this yet, but at a quick glance, the first thing that jumps out at me is env var and config handling. I think there's maybe a misunderstanding here about env vars work? For example, why does the Snakemake workflow explicitly prepend lots of commands with env vars instead of letting them ambiently pass through? Why switch away from `write-envdir` (which doesn't reveal secrets) to passing config options (which does, and thus necessitates other changes)? I saw the comment that:

> By default, Snakemake doesn't pass environment variables through to shell()

but I don't think this is true (and have tested it locally). Let's talk more about what's going on here?

(Both of these changes also introduce a number of places which appear to have quoting/escaping issues. I haven't inspected closely yet, though, as I think they can all go away given the above.)

jameshadfield:
Hmmm. Yes, local testing indicates that env variables are indeed available to `shell()`. The misunderstanding came from my reading of [the docs](https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#environment-variables) which state:

> note that this does not mean that Snakemake makes [environment variables] available e.g. in the jobs shell command. Instead, for data provenance and reproducibility reasons, you are required to pass them explicitly to your job via the params directive

I'll update this PR to use env variables.

tsibley:
Ah, yeah, those docs are weirdly worded and confusing. I _think_ they mean that, using their example, you can't do this:

```
    shell:
        "echo {os.environ['SOME_VARIABLE']} > {output}"
```

or 

```
    shell:
        "echo {SOME_VARIABLE} > {output}"
```

and so if you wanted the same effect, you have to do the equivalent of:

```
    params:
        x=os.environ["SOME_VARIABLE"]
    shell:
        "echo {params.x} > {output}"
```

which is weird to want to do, because this works just fine:

```
    shell:
        "echo $SOME_VARIABLE > {output}"
```

jameshadfield:
Updated to use environment variables (similar to as before). The combination of `nextstrain build` + `envdir` + `snakemake` is a bit complex so I'll note here for others:
* The env variables `$AWS_ACCESS_KEY_ID` and `$AWS_SECRET_ACCESS_KEY` are passed through to Snakemake automatically by `nextstrain build` ([code here](https://github.com/nextstrain/cli/blob/1f7c233b9357f2488cf003a06c6c80358f3783e9/nextstrain/cli/hostenv.py#L24-L25)). I thought about adding them to the `envdir` calls (along the lines of "explicit is better than implicit") but opted to reuse the existing implementation. 
* I opted for a slightly conservative memory ceiling for Snakemake

Some documentation, which I'll happily add to a markdown if desired:

- To send slack notifications: set environment variables `SLACK_TOKEN` and `SLACK_CHANNELS`
- To trigger a subsequent preprocessing job: use `--config trigger_preprocessing=True` and set environment variable `PAT_GITHUB_DISPATCH` 
- To fetch sequences & metadata from the database: use `--config fetch_from_database=True` and, if GISIAD, set environment variables `GISAID_API_ENDPOINT` and `GISAID_USERNAME_AND_PASSWORD`.
- To store resulting files in specific S3 location: use `--config s3_dst=s3://....`

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+
+rule all:
+    input: _compute_targets
+
+#################################################################
+###################### rule definitions #########################
+#################################################################
+
+
+def run_shell_command_n_times(cmd, msg, cleanup_failed_cmd, retry_num=5):
```
tsibley, line 42: It would be neat to replace this with [built-in rule-specific retry behaviour](https://github.com/snakemake/snakemake/issues/367), if that ever lands in Snakemake.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+        s3_bucket = config["s3_src"]
+    output:
+        touch("data/gisaid/notify.done")
+    run:
+        shell("./bin/notify-slack --upload flagged-annotations < {input.flagged_annotations}")
+        # notify-on-metadata-change disabled as csv-diff runs out of memory
+        # shell("./bin/notify-on-metadata-change {input.metadata} {params.s3_bucket}/metadata.tsv.gz gisaid_epi_isl")
+        shell("./bin/notify-on-additional-info-change {input.additional_info} {params.s3_bucket}/additional_info.tsv.gz")
+        shell("./bin/notify-on-flagged-metadata-change {input.flagged_metadata}  {params.s3_bucket}/flagged_metadata.txt.gz")
+        shell("./bin/notify-on-location-hierarchy-addition {input.location_hierarchy} source-data/location_hierarchy.tsv")
```
tsibley, line 304: Why not make this one `shell:` block instead of a `run:` block with lots of `shell()`s?

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+    params:
+        s3_bucket = config["s3_src"]
+    output:
+        touch("data/genbank/notify.done")
+    run:
+        shell("./bin/notify-on-metadata-change {input.metadata} {params.s3_bucket}/metadata.tsv.gz genbank_accession")
+        # TODO - which rule produces data/genbank/problem_data.tsv? (was not explicit in `ingest-genbank` bash script)
+        shell("./bin/notify-on-problem-data data/genbank/problem_data.tsv")
+        shell("./bin/notify-on-location-hierarchy-addition {input.location_hierarchy} source-data/location_hierarchy.tsv")
+        shell("./bin/notify-on-duplicate-biosample-change {input.duplicate_biosample} {params.s3_bucket}/duplicate_biosample.txt.gz")
```
tsibley, line 320: Ditto?

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+
+onstart:
+    print(f"Pipeline starting.")
+    print(f"Source s3 bucket: {config['s3_src']}, destination: {config['s3_dst']}")
+    print("Environment variables present:")
+    for var, description in env_variables.items():
+        print(f"\t${{{var}}}: " + ("YES" if os.environ.get(var, "") else "NO") + f"({description})")
+    if send_notifications:
+        message="🥗 GISAID ingest" if config["database_name"]=="gisaid" else "🥣 GenBank ingest"
+        shell(f"./bin/notify-on-job-start \"{message}\"")
```
tsibley, line 392: This is great! :sparkles:

.github/workflows/fetch-and-ingest-gisaid-master.yml
```diff
@@ -24,18 +24,24 @@ jobs:
...
           GITHUB_RUN_ID \
           SLACK_TOKEN \
           SLACK_CHANNELS \
           PAT_GITHUB_DISPATCH
 
+        GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
+        declare -a config
+        config+=(
+          fetch_from_database=True
+          trigger_preprocessing=True
```
huddlej, line 42: Pre-merge change: Is there a reason these config arguments aren't hardcoded in the Snakemake command below? I'd think it would be more readable to keep these config flags as close as possible to where they are used.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+
+# certain files can be fetched from the database or downloaded from our S3 bucket.
+# config["fetch_from_database"] dictates this
+# We could implement this via remote snakemake inputs (such as in nextstrain/ncov),
+# conditional rules (a bit of an anti-pattern) or ruleorders.
+ruleorder: download_previous_biosample > fetch_biosample_from_database
+ruleorder: download_previous_raw_ndjson > fetch_from_database
+if config.get("fetch_from_database", False):
+    ruleorder: fetch_biosample_from_database > download_previous_biosample
+    ruleorder: fetch_from_database > download_previous_raw_ndjson
```
tsibley, line 15: Hmm. I agree that special input functions as conditionals are best avoided if possible, but not sure I find ruleorders much better. We could instead conditionalize the rule _definitions_ themselves (either at the whole `rule` level or at the `shell/run` block level) with normal Python `if/elif/else` or standard function calls instead of using ruleorder to mask some rules. Not a blocking concern, just a thought.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+        try:
+            subprocess.run(cmd, shell=True, check=True)
+            break
+        except subprocess.CalledProcessError:
+            print("...FAILED")
+            attempt+=1
+            shell("{cleanup_failed_cmd} && sleep 10")
+    else:
+        print(msg + f" has FAILED {retry_num} times. Exiting.")
+        raise Exception("function run_shell_command_n_times has failed")
```
huddlej, line 55: Post-merge change: I see that this approach existed in the original shell script, but now that we are using Snakemake, we can use [its `--restart-times` argument](https://snakemake.readthedocs.io/en/stable/executing/cli.html) to specify how many times we want to retry any failing rule. This logic will apply to all rules in the workflow, but that should be OK. Then we can drop this kind of custom function that tries to do the same thing inside `run` blocks and just make the shell call in a normal `shell` block.

tsibley (commented):
The invocation with env vars seems better to me, thanks for the rework!

I didn't go through the Snakefile with a fine-toothed comb, but I left a few minor comments here and there.

One overall question is that I'm not sure why `{database}` is used as a wildcard throughout the workflow instead of avoiding wildcards all together (and the input functions they necessitate here) by directly interpolating the value of `config["database"]` into values using f-strings? I think this would clean up a bit of unnecessary indirection, but maybe I'm missing a reason it's useful!

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+
+rule fetch_biosample_from_database:
+    message:
+        """Fetching Biosample from the database"""
+    params:
+        s3_src_bucket = config["s3_src"]
+    output:
+        biosample = "data/biosample.ndjson"
+    run:
+        run_shell_command_n_times(
```
huddlej, line 85: This is an example where we could use `--restart-times`.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+    input:
+        unpack(_raw_files_to_upload)
+    output:
+        touch("data/{database}/raw.upload.done")
+    params:
+        quiet = "" if send_notifications else "--quiet",
+        s3_bucket = config["s3_dst"]
+    run:
+        for remote, local in input.items():
+            shell("./bin/upload-to-s3 {params.quiet} {local:q} {params.s3_bucket:q}/{remote:q}")
```
huddlej, line 107: Post-merge change: This for-loop approach may be historical, but we might consider splitting this out to separate Snakemake rules via wildcards (e.g., `data/{database}/raw.upload.{local}.done`).

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+        clades and calculate other useful metrics.
+        """
+    input:
+        sequences = "data/{database}/nextclade.sequences.fasta",
+        nextclade_info = "data/{database}/nextclade_old.tsv"
+    params:
+        new_info = temp("data/{database}/nextclade_new.tsv"),
+        nextclade_input_dir = temp(directory("data/{database}/nextclade_inputs")),
+        nextclade_output_dir = temp(directory("data/{database}/nextclade"))
+    threads: 8
```
huddlej, line 223: Pre-merge change: We should pass this `threads` variable directly to Nextclade to prevent overcommitting resources when we run Snakemake rules in parallel (something we might not have been doing before with the bash scripts). We could do this by modifying the `run-nextclade` script to take a threads argument and then pass this with `run-nextclade --threads {threads}`.

huddlej (changes requested):
This generally looks good to me. I haven't tested it (because testing is tricky), but I left some notes about changes I'd recommend pre-merge and those that can be left to post-merge.

.github/workflows/fetch-and-ingest-genbank-branch.yml
```diff
@@ -16,13 +16,18 @@ jobs:
...
-          GITHUB_RUN_ID \
-          SLACK_TOKEN
+          GITHUB_RUN_ID
+  
+        GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
+        declare -a config
+        config+=(
+          fetch_from_database=True
+          s3_dst=s3://nextstrain-staging/files/ncov/open/branch/${GITHUB_BRANCH}
+        )
```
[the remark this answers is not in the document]
tsibley: …and [this new PR](https://github.com/nextstrain/ncov-ingest/pull/238) extends Shellchecking to our GitHub Actions workflows.

Snakefile
tsibley: > This logic will apply to all rules in the workflow, but that should be OK. 

I'm not so sure about this. There are a few very long-running rules in this workflow, which means if they fail late in a deterministic way, we'd spend hours retrying them only to see the same late failures again, before reporting failure of the whole workflow.

huddlej: Yeah, that's a fair point. I wrote that comment just before asking about the NDJSON download times, etc. Probably better to wait for Snakemake to support this as a rule-specific option, as you mention.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+                {input.sequences:q} \
+                {params.new_info} \
+                {params.nextclade_input_dir} \
+                {params.nextclade_output_dir}
+            # Join new and old clades, so that next run won't need to process sequences that are already processed
+            ./bin/join-rows \
+                {input.nextclade_info:q} \
+                {params.new_info} \
+                -o {output.nextclade_info:q}
+        fi
```
huddlej, line 245: Just making a note that we should consider moving this conditional logic either a) _into_ a single script that handles the logic or b) _out of_ a single Snakemake rule and up to a workflow-level conditional statement. I've personally had more success debugging logic with the first approach where the I can test a single script manually outside of the workflow and inside the workflow.

Snakefile
huddlej: This seems like a case where asking ourselves how we would implement this logic in a different workflow language might lead to a clearer solution. In a less expressive workflow language, we might be forced to move this kind of conditional logic into a single script that takes `--fetch-from-database` as an argument and produces the output(s) supported by either downloading from S3 or GISAID, for example.

This makes the workflow simpler at the expense of more complicated scripts.

Snakefile
```diff
@@ -0,0 +1,402 @@
...
+###################### rule definitions #########################
+#################################################################
+
+
+def run_shell_command_n_times(cmd, msg, cleanup_failed_cmd, retry_num=5):
+    attempt = 0
+    while attempt < retry_num:
+        print(f"{msg} attempt number {attempt}")
+        try:
+            subprocess.run(cmd, shell=True, check=True)
```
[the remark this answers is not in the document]
huddlej: Agree. This should probably happen before merge, so we don't get bitten sooner or later by this issue.

Snakefile
jameshadfield: 👍  resolving as this todo in #239 (and there's also a todo comment in the code for this rule)

Snakefile
jameshadfield: Agreed! This is now documented for posterity in #239

Snakefile
jameshadfield: Agreed! This is a great example of what I hope can be improved now that we use a workflow :) Documented in #239 so resolving here

Snakefile
jameshadfield: I think they're much of a muchness. I chose this form as it makes it easier to break those scripts out into python commands if desired down the track. This is also noted in #239 so I'll resolve this for now.

Snakefile
jameshadfield: Good call - resolved in 8204f9b

Snakefile
jameshadfield: Switched to python conditionals within the rules, which I agree is nicer. Commit ab59826

Snakefile
jameshadfield: Thanks!

Snakefile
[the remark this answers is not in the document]
jameshadfield: Yes it can — it turns out that when `shell()` fails it throws a `subprocess.CalledProcessError` so we can even use  the same try/except logic here! Implemented in ab59826

.github/workflows/fetch-and-ingest-gisaid-master.yml
```diff
@@ -24,18 +24,24 @@ jobs:
...
           AWS_DEFAULT_REGION \
           GISAID_API_ENDPOINT \
           GISAID_USERNAME_AND_PASSWORD \
-          GITHUB_REF \
           GITHUB_RUN_ID \
           SLACK_TOKEN \
           SLACK_CHANNELS \
           PAT_GITHUB_DISPATCH
 
+        GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
```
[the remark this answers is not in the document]
jameshadfield: removed in ab59826

.github/workflows/fetch-and-ingest-gisaid-branch.yml
```diff
@@ -16,15 +16,20 @@ jobs:
...
-          GITHUB_REF \
           GITHUB_RUN_ID \
-          SLACK_TOKEN
+
+        GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
+        declare -a config
+        config+=(
+          fetch_from_database=True
+          s3_dst=s3://nextstrain-ncov-private/branch/${GITHUB_BRANCH}
+        )
```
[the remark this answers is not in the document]
jameshadfield: 👍 done in ab59826

.github/workflows/fetch-and-ingest-genbank-branch.yml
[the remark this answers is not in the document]
jameshadfield: 👍 done in ab59826

Snakefile
```diff
@@ -0,0 +1,381 @@
...
+        If there are sequences without clades, then run nextclade to assign
+        clades and calculate other useful metrics.
+        """
+    input:
+        sequences = f"data/{database}/nextclade.sequences.fasta",
+        nextclade_info = f"data/{database}/nextclade_old.tsv"
+    params:
+        new_info = temp(f"data/{database}/nextclade_new.tsv"),
+        nextclade_input_dir = temp(directory(f"data/{database}/nextclade_inputs")),
+        nextclade_output_dir = temp(directory(f"data/{database}/nextclade"))
```
joverlee521, line 201: I was confused by the use of `temp` here. Since these are not consumed as inputs in another rule, I don't think these ever get removed by snakemake.

.github/workflows/fetch-and-ingest-gisaid-master.yml
jameshadfield: We [recently shifted to this form in nextstrain/ncov](https://github.com/nextstrain/ncov/commit/c91ad66fa22cedc7484c72c8b67336a35ef562f3) which I actually find easier to read, especially if we increase the complexity of the actions in the future and start modifying which parameters are in `config`. (I don't feel super strongly about this however.)

jameshadfield:
Thanks for the comments @tsibley @huddlej - I've resolved them all (I think). I plan to merge tomorrow assuming the [GISAID](https://github.com/nextstrain/ncov-ingest/actions/runs/1518794768) and [GenBank](https://github.com/nextstrain/ncov-ingest/actions/runs/1518794769) per-branch actions work.

cc @ivan-aksamentov for commit 8204f9b which (optionally) passes through a `--jobs` parameter to nextclade.

joverlee521 (approved):
Thank you James for working through this. It was very helpful that there were minimal changes to the actual scripts so that we could focus on the conversion to snakemake 🙏 

I left a couple comments below but nothing blocking.

Snakefile
```diff
@@ -0,0 +1,381 @@
...
+
+onstart:
+    print(f"Pipeline starting.")
+    print(f"Source s3 bucket: {config['s3_src']}, destination: {config['s3_dst']}")
+    print("Environment variables present:")
+    for var, description in env_variables.items():
+        print(f"\t${{{var}}}: " + ("YES" if os.environ.get(var, "") else "NO") + f"({description})")
+    if send_notifications:
+        message="🥗 GISAID ingest" if database=="gisaid" else "🥣 GenBank ingest"
+        shell(f"./bin/notify-on-job-start \"{message}\"")
```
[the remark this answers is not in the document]
jameshadfield: I think this is a "feature" of GitHub actions -- when I run locally they are printed before any rule is run. Would be interested to know what exactly is happening tho.

Snakefile
jameshadfield: My understanding is that because they aren't used by another rule they will be cleaned up after this rule (`annotate_via_nextclade`) finishes.

ivan-aksamentov:
@jameshadfield 

> cc @ivan-aksamentov for commit 8204f9b which (optionally) passes through a --jobs parameter to nextclade.

Hmm.. looks a bit wonky. And I am not familiar with this bash syntax. I don't think limiting Nextclade's concurrency is needed.
Will something run concurrently with Nextclade step?
Where does the `threads` variable come from and what it will be by default?

Current daily ingest runs on 16 cores, I think
https://github.com/nextstrain/ncov-ingest/blob/9e257d85a9012c1d4fca6c0a53dfecdfc4ec29c6/.github/workflows/fetch-and-ingest-gisaid-master.yml#L42-L50

If we mean to speed things up, I think all these limitations should be removed.

My current suggestion is to remove the `threads` and the `--jobs` passthrough for now and unleash the parallelism (unless there are Snakemake-releated considerations). And later, separately, when we inline that bash script into the Snakemake, we can pass args directly to Nextclade, including the `--jobs`, if someone is into it. But I think by default we should not use this flag.


Regarding inlining the bash script. There are 4 steps in this script:
 - Download latest release of Nextclade - probably can be made prior to running Snakemake
 - Download Nextclade dataset into a directory - can be a step in Snakemake?
 - Run Nextclade
 - Join outputs with the cache - can be a separate step in Snakemake?

Each step depends on previous and none of this should run if there are no new sequences after filter-fasta step.

Snakefile
jameshadfield: Hey @ivan-aksamentov - those changes were in response to this review comment so i'll inline responses here to keep things ordered.

> Will something run concurrently with Nextclade step?

Probably not at the moment, but I think @huddlej's comment was to future proof ourselves for the situation when a job does run concurrently.

> Where does the threads variable come from and what it will be by default?

Snakemake will choose the minumum of (i) the user-provided cores (via snakemake's `--cores` arg), (ii) system cores available and (iii) 8, which is hardcoded in the rule. I can easily change (iii) to 16. Relatedly, my understanding was that nextclade didn't really scale well beyond 8 cores, or was specific to nextalign?

P.S. I specifically modified the bash script in such a way that it's backwards compatible -- if you call it with 4 arguments as before, then the behavior is unchanged.

Snakefile
joverlee521: Hmm, the [docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=temp#protected-and-temporary-files) say 
>an output file marked as temp is deleted after all rules that use it as an input are completed

So I wonder if snakemake will delete them if they are `output` rather than `params`?
(These were all included in the [final output](https://github.com/nextstrain/ncov-ingest/runs/4329428003?check_suite_focus=true#step:4:458) from `./bin/clean` so snakemake did not remove them.)

jameshadfield: Good find! I'll try to fix but might be pushed to a subsequent PR

Snakefile
ivan-aksamentov: Okay then. But we need to take care to not put unnecessary slowdowns/limits.


> Relatedly, my understanding was that nextclade didn't really scale well beyond 8 cores, or was specific to nextalign?

The scaling is not great currently, for both tools, but I don't think it should be additionally limited artificially. Especially considering that it is the slowest and most resource-intensive part of the pipeline. I think the hard limit of 8 or 16 threads should be removed. At least this is what I would do if I wanted things to go faster. I plan to invest some time on profiling and optimizing the C++ code, so that it will scale better hopefully soon.

Regarding the bash script, I think it's meant to go away anyways. It's really mostly a glue, with only 3 statements that do anything useful. So we should probably not add any more things into it. Taking care of backwards compat is also not necessary, I don't think anyone is using this script directly. But if keeping the current latest commit is the simplest for you, I'd just go with it.

jameshadfield: Sweet. I'll increase it to 16 cores (which is what the AWS job is provisioned with) before merge (but will first wait to see if the jobs succeed).

Agree with the statement that the bash script will go away eventually - my desire for this PR was to modify the status quo as little as possible.

tsibley:
@jameshadfield I pushed one commit to the tip of this branch to remove the unnecessary input functions.

Snakefile
jameshadfield: Resolved in 
416c56f

Snakefile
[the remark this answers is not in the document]
joverlee521: Huh, I also see the messages print out before any rules when I run it locally.

The latest [AWS logs](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/nextstrain-job_21_nextstrain-ncov-ingest_latest$252Fdefault$252F21cbcf2a84df4c309bd977f497a6180e) show the timestamp of the onstart messages at the end of the pipeline, so it's not a GitHub issue 🤔

Snakefile
```diff
@@ -0,0 +1,377 @@
...
+
+onsuccess:
+    message = "✅ This pipeline has successfully finished 🎉"
+    print(message)
+    shell("./bin/clean")
+
+onerror:
+    message = "❌ This pipeline has FAILED 😞. Please see linked thread for more information."
+    print(message)
+    print("P.S. Because this failed, we did not run './bin/clean' but you may wish to run this now.")
```
tsibley, line 377: Does not running clean here on failure mean that remote jobs may try to upload several large files when they fail?

If so, would it be better to only skip clean on local runs?

jameshadfield: Are you here referring to the `nextstrain build` behavior on AWS where it'll upload the entire working directory as a (zipped) asset to S3?

> If so, would it be better to only skip clean on local runs

Sure thing. My inclination would be to have an explicit opt-in config param here, such as `--config keep_all_files=True`, do you have an alternative method you'd prefer?

huddlej:
> ...unleash the parallelism (unless there are Snakemake-releated considerations)

@ivan-aksamentov You're probably right that it doesn't matter much here. One standard I've tried to follow with Snakemake workflows is to always document rules that run multithreaded commands by adding the `threads` parameter and explicitly passing the value of that parameter to the command. Without this information, Snakemake can schedule multiple multithreaded jobs to run simultaneously without realizing it has overcommitted resources. When everything is a shell script, I don't usually worry about this scenario as much, but when Snakemake allows you to type `-j 96` or `--cores 96` to run many rules at once, it's more likely to happen.

Snakefile
tsibley: Yes, that `nextstrain build` behaviour.

Opt-in config flag is fine by me! (Though it may mean someone debugging a local run will forget too set it and then have to re-run with it set to inspect failed outputs.)

jameshadfield: Actually, I'm going to merge as-is for expediency as the current behavior preserves the previous (bash script) behavior where `./bin/clean` was skipped if there was an error in any previous step. (The one exception being if the GitHub action trigger were to fail.) I do agree that clean-up of files is important and we should audit this as part of #240.

jameshadfield: Implemented in 45dcea6 and i'll also note this on #240.

jameshadfield:
Merging now - note that I cancelled the two CI runs triggered by commit 45dcea6 as I'm confident they don't materially change the behavior (famous last words!)

Snakefile
jameshadfield: Didn't get to this, added as a to-do in #239

Snakefile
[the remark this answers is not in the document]
jameshadfield: Huh! I've just triggered a run on AWS and can confirm that the slack message (from `onstart`) comes through at the start of the run but the `print()` output has not appeared in the AWS logs.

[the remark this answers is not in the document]
joverlee521: OH! This isn't specific to the `onstart` handler. Looking at the [failed run logs](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/nextstrain-job_21_nextstrain-ncov-ingest_latest$252Fdefault$252F21cbcf2a84df4c309bd977f497a6180e) again, _all_ of the `print()` statements are printed at the end with the exact same timestamp.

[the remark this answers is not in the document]
jameshadfield: Long ago I remember having to flush print calls, e.g. `print(..., flush=True)`. It may resolve this issue?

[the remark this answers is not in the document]
joverlee521: Ah, that may resolve it. I see `flush` suggested in this [random carpentries tutorial](https://carpentries-incubator.github.io/workflows-snakemake/05-snakemake-python/index.html) as well.

PR #232, opened 2021-11-23
feat: allow for custom Nextclade binary and dataset for full runs

ivan-aksamentov:

This adds 2 checks in the beginning of "Nextclade full run" script:

1. Whether `nextclade` binary is present in the root of the project.

   By default the latest release of Nextclade CLI is downloaded. But if the `nextclade` file is detected, then the download is skipped and this file is used instead.

2. Whether `tmp/dataset/` directory contains `tree.json` and `reference.fasta` files.

   By default the latest dataset is downloaded. Hoverver, if a dataset is detected in the `tmp/dataset/`, the download is skipped and this dataset is used instead.

This allows to customize manual Nextclade full runs, when a combination of a specific Nextclade version (perhaps unreleased) and of a specific dataset is needed.

This customization is intended for cases when putting Nextclade binary and dataset into a local working directory of ncov-ingest and running one of the `./run-nextclade-full*` scripts manually. It would be nice to somehow simplify and automate this process, but for now I don't have good ideas for this. In particular, because the binary and datasets are different depending on a particular use-case for these custom runs.

### Related issue(s)  

Extends: #218 

### Testing

I am running these 2 jobs to check this:

```
nextstrain build --aws-batch --no-download --attach 4b6bbd1d-f8ff-4f31-8f53-157e4947b046 .
nextstrain build --aws-batch --no-download --attach 191a049c-5bae-4c5f-86d2-0536a625c84a .
```
These particular jobs are using the custom unreleased Nextclade version from https://github.com/nextstrain/nextclade/pull/610

### DIFF
--- a/bin/run-nextclade-full
+++ b/bin/run-nextclade-full
@@ -123,9 +123,13 @@ main() {
 
   ./bin/notify-on-job-start "🦬 'Nextclade full run: ${DATABASE}'" "The job involves ${N_PROCESSORS} CPUs and will run ${N_CONCURRENT_JOBS_MAX} Nextclade instances in parallel, each processing ${BATCH_SIZE} sequences at a time. The results will be uploaded to \`${S3_DST}\`. Someone needs to inspect these results and then copy them over to the S3 directory where the next scheduled ingest can find them, replacing the old files. For more details, see PR #218 in ncov-ingest."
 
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
I cancelled the triggered "daily" ingest checks, because this PR intends to modify the Nextclade full run workflow, no the daily one.

ivan-aksamentov:
I've tested this in battle a few times now and it works great. I think this is safe to merge, considering it only affects full runs and not daily flow.

PR #234, opened 2021-11-27
Update gisaid ingest schedule to 1pm CET every day

emmahodcroft:
Long term: 
- Change the start of pull-and-ingest time from 3AM UTC to 12PM UTC (should be 1pm CET), so that we catch the latest download update from GISAID (should be around 11am CET) - with room for delays on their end

Short term:
- Run pull-and-ingest every day (not just M-F) for the short-term to pull in new Omicron sequences ASAP

Not tested!

### DIFF
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -3,6 +3,8 @@ name: GISAID fetch and ingest
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
+    # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule
+    # https://crontab.guru/
     - cron:  '7 13 * * *'
 
   # Manually triggered using `./bin/trigger gisaid/fetch-and-ingest`
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -3,7 +3,6 @@ name: GISAID fetch and ingest
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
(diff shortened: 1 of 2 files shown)


.github/workflows/fetch-and-ingest-gisaid-master.yml
```diff
@@ -3,8 +3,7 @@ name: GISAID fetch and ingest
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
-    # "At 03:00 (05:00 CET) on every day-of-week from Monday through Friday."
-    - cron:  '0 3 * * MON-FRI'
+    - cron:  '7 13 * * *'
```
victorlin, line 6: This means 1:07pm daily to be exact. There's different forms of CRON syntax, but you can use an online tool to decipher the one used by GitHub, from the [official docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule):

> You can use [crontab guru](https://crontab.guru/) to help generate your cron syntax and confirm what time it will run.

victorlin (approved):

.github/workflows/fetch-and-ingest-gisaid-master.yml
corneliusroemer: Yes, I got rid of the MON-SUN since `*` is simpler

I start at 1.07pm because Github doesn't like if you start it on the hour since that's what most people use and it gets crowded.

ivan-aksamentov:
Looks fine to me.




It modifies only GISAID workflow. Should GenBank be adjusted for symmetry?

ivan-aksamentov:
We chatted with Emma on Slack and agreed to also adjust the schedule for GenBank, so that they happen at the same time.

I also added the links that Victor mentioned above as comments to the yaml, for people (me) who search every time how to write a correct cron expression

PR #236, opened 2021-11-29
Move ingest trigger up 1 hr to 1pm CET

emmahodcroft:
Follow up to #234 - this just adjusts the timing to be 12 UTC which is currently 13 CET. (Previously it was 13 UTC which is 14 CET). Also added a comment that this is UTC and the relationship to CET, to help make this easier in future! (Though you still have to know if it's daylight savings or not, to know if it's 1 or 2 hours ahead of UTC....)

This is for both GISAID & Genbank


emmahodcroft:
I cancelled the test ingests as this was a very minor change, no need to run an entire ingest.

PR #237, opened 2021-11-29
feat: add incremental sequence alignment

ivan-aksamentov:

During daily ingest, extracted full fasta file is being filtered by whether sequence name is already present in the nextclade.tsv file from S3, which is an accumulation of all Nextclade outputs produced during previous ingests. The  sequences that are not present in nextclade.tsv are considered new and are passed to Nextclade for processing.

Nextclade aligns these sequences on every run. However currently alignment is ignored (not emitted).

In this PR:

I add `--output-fasta` flag to Nextclade invocation, so that the nucleotide alignment is emitted for the new sequences into `data/gisaid/nextclade.aligned.new.fasta`.
The old "cache" of aligned sequences is downloaded from S3
```
nextclade.aligned.fasta.xz
```
as 
```
data/gisaid/nextclade.aligned.old.fasta
```
Then I concatenate the 2 fasta files to get the new   
```
data/gisaid/nextclade.aligned.fasta
```
and is uploaded back to S3.

That is, this file will then accumulate aligned sequences after each daily ingest.
This behavior is similar to the concatenating nextclade.tsv for new sequences to the accumulator nextclade.tsv from S3.

For the very first run, I generated the `nextclade.aligned.fasta.xz` by running the Nextclade full run.

TODO:
 - [x] Implement downloading accumulator fasta from S3 in the beginning of the run
 - [x] Emit alignment
 - [x] Concatenate new fasta to the accumulator fasta
 - [x] Implement uploading accumulator fasta to S3 in the end of the run
 - [x] Modify `bin/run-nextclade-full` script to also emit and accumulate the alignment. Run it once to generate the initial alignment of the entire database.
 - [x] Repeat the same steps for GenBank
 - [x] Calculate mutation summary in daily ingest
 - [x] Implement uploading mutation summary to S3
 - [x] Modify `bin/run-nextclade-full` script to generate mutation summary. Run it once to generate the initial mutation summary of the full database.
 - [x] Run daily ingest to verify that incremental updates of the full alignment and mutation summary work
 - [x] Repeat the same steps for GenBank
 - [x] Port to Snakemake (in a separate PR)

### Testing

This is what I did to test alignment for GISAID (Upd: this was before mutation summary was added):

 - 2021-12-02: ran "Nextclade full run" on AWS Batch to generate alignment for the entire GISAID (appeared as `nextclade.aligned.fasta.xz` in a special directory on S3)
 - copied the alignment (`nextclade.aligned.fasta.xz`) to the root of the bucket, where the modified daily ingest can find it
 - 2021-12-03: waited until GISAID data updated
 - 2021-12-03: ran daily GISAID locally, checked that the resulting concatenated fasta file has grown. I used `seqkit stat` to count number of sequences.


jameshadfield:
This is great - thanks for starting this @ivan-aksamentov. Let me know if I can help to convert to Snakemake. 

This should allow us to get rid of the preprocessing step (from `nextstrain/ncov`) and simply start phylogenetic builds from the alignment produced here. This will be a huge improvement! The remaining steps which would be needed to achieve this (not necessarily for this PR):
* There's a mutation summary computed from the alignment and other files produced by nextalign/nextclade. I think we'd want to calculate this here in `ncov-ingest`. cc @emmahodcroft 
* Filtering (mainly QC)

Finally, on today's call we also discussed indexing this alignment here, but that may also be best left for a subsequent PR.

emmahodcroft:
>There's a mutation summary computed from the alignment and other files produced by nextalign/nextclade. I think we'd want to calculate this here in ncov-ingest.

Agreed, if we're doing other things in Nextclade here, we might as well gather up more information!
One thing I'm a little less clear about is whether there's any difference between the mutation summary that ends up back in the metadata & the `mutation_summary.tsv` file - which is what I currently use, at least. I could make the switch to using the one in metadata (if someone can let me know any key differences) - but it would be super nice if this could wait until things die down a bit from Omicron 😅

ivan-aksamentov:
I just pushed all necessary steps to produce nuc alignment and ran full on AWS + daily locally for GISAID. See the updated first message.

However, I just realized that Nextalign in ncov also produces and uses:
 - peptides: emitted into `--output-dir` as `<output-dir>/gene.<gene_name>.fasta`
 - insertions: emitted into `--output-insertions` as `insertions.csv`

So these folks also need to be packaged into the cache...

@jameshadfield @rneher Can you tell if some of these steps can be moved from ncov into ingest to reduce the amount of things to be cached? I don't think peptides are used raw, they are probably summarized into something? So maybe we can join this stuff into metadata?

Alternatively, we could declare caching of alignment in ingest a mistake and instead do the caching in ncov, whatever is simpler. 

Or what is a good solution here?

ivan-aksamentov:
I added similar logic for peptide alignment. It's on branch `feat/incremental-alignment-with-peptides`. The tricky part here is the full runs. Because of batching, joining many fasta files gets very hairy in bash. My current hope is that we don't need to cache them and can just append the required summarized data to metadata instead.

Just realized that we might not need to cache insertions, because they are already in the metadata and are exactly the same thing as in `insertions.csv`. I hope `ncov` can use Nextclade's insertions from metadata instead of from `insertions.csv`.

jameshadfield:
> Can you tell if some of these steps related to peptides and insertions can be moved from ncov into ingest to reduce the amount of things to be cached?

The insertions, translations etc for the entire alignment are only needed to produce the mutation summary. This is currently [part of the ncov preprocessing workflow](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/export_for_nextstrain.smk#L91-L121) (note that this is different from `rule build_mutation_summary`   [which is done on the subsampled alignments](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/main_workflow.smk#L827-L853), but both rules use the same python script).

The mutation summary can be shifted over to ncov-ingest as we are the only consumers of it, and this is what I recommend. This allows ncov-ingest to upload the alignment & mutation summary without needing to upload insertions / peptides etc. 

P.S. I improved the efficiency of the mutation summary script recently so it shouldn't be the end of the world to simply recompute this each time for the entire alignment; at the very least it shouldn't block this PR. 

> And what is the story with these indexes you are talking about?

The indexing of the alignment (often done behind-the-scenes by `augur filter`) is taking a long time. Ideally this could be computed here (via `augur index`) and uploaded alongside the alignment. However there was discussion whether the index is actually needed since we have most of this information in the metadata and thus I think we ignore this for the purposes of this PR.

ivan-aksamentov:
Thanks, James, @jameshadfield 

> The mutation summary can be shifted over to ncov-ingest as we are the only consumers of it

`mutation_summary.tsv` seems to be containing nuc and aa mutations per gene. `metadata.tsv` already contains it in a slightly different format. I'd say that we might not need `mutation_summary.tsv` at all. But again I don't know how it's used. I could  produce it from `metadata.tsv`.


There's still a few questions remaining:
 - The whole idea is to avoid running Nextalign in `ncov` for GISAID sequences. Let's say we precompute nuc alignment and `mutation_summary.tsv`, and we don't need peptides then. But a re there other remaining outputs of Nextalign in `ncov` that are used and need to be precomputed? For example `insertions.tsv` (these are also in the metadata by the way).
 - GISAID alignment is handled by ingest fully, okay, but how about these spike-in sequences and otherwise external non-GISAID inputs. How I imagine that: currently external unaligned fastas and metadatas are merged to the GISAID ones and and passed to Nextalign. With the cache instead we need to: pass externals to Nextalign and then merge the results. This merge might be tricky. I don't have a clear idea of how `ncov` should be changed to accommodate the GISAID cache (because as we discovered it goes beyond just nuc alignment).

emmahodcroft:
> I'd say that we might not need mutation_summary.tsv at all. But again I don't know how it's used. I could produce it from metadata.tsv.

Long term you're right, we have a lot of this info (if not the same thing exactly? I haven't had time to check!) in metadata.tsv now. The plea is mostly a personal one from me - I use `mutation_summary.tsv` as *the* key part of CoVariants, and though I can make a switch longer term, it would be an enormous help if I don't have to do this right now - I've got so many balls in the air. However, if the information is the same, happy for this to just be copied out of metadata. And my apologies!! 🙏 🙏 

>But are there other remaining outputs of Nextalign in ncov that are used and need to be precomputed? For example insertions.tsv (these are also in the metadata by the way).

James can give more detail here, but my naive assumption would be to try and keep the outputs the same as much as possible unless we're absolutely sure they're redundant now. We could perhaps do this by incrementally updating more than just the alignment, but these files as well?

>GISAID alignment is handled by ingest fully, okay, but how about these spike-in sequences and otherwise external non-GISAID inputs.

At the moment we actually align twice - once the whole of GISAID in order to get a master file that we can do priorities & etc on. Then once you've picked your set to actually do a build on - because this may contain spiked in sequences - we align again. Because these are often only 10,000 or less, this is actually pretty fast. We could keep doing this.

Two other options:
- We set a rule so that anything that isn't set as a GISAID or Genbank input gets aligned before getting included into the set to be 'actually built' (we already have a set to dedup & clean up such files (though fool-proof deduping obviously lies with the user - if you change all the seq names... well)) & stop aligning after the combining
- We simply expect that users use a simple command to align their own spiked sets before they include them in the workflow, either manually or via a localrule that they write

jameshadfield:
> I'd say that we might not need mutation_summary.tsv at all. 

I agree with Emma - while the mutation summary may not be needed long term, it's straightforward to keep it going in the short-to-medium term.

> Let's say we precompute nuc alignment and mutation_summary.tsv in ingest, and we don't need peptides then. But are there other remaining outputs of Nextalign in ncov that are used and need to be precomputed? For example insertions.tsv (these are also in the metadata by the way).

No extra files needed :) I think the only files from ncov-ingest which this PR should add to the upload are
1. `aligned.fasta.xz`
2. `mutation-summary.tsv.xz`


> GISAID alignment is handled by ingest fully, okay, but how about these spike-in sequences and otherwise external non-GISAID inputs ...

The current `ncov` workflow allows such behavior via a config file with multiple inputs. Each input is aligned (via nextalign) independently, and only if needed. 

>  I don't have a clear idea of how ncov should be changed to accommodate the GISAID cache (because as we discovered it goes beyond just nuc alignment).

I don't think it does go beyond nuc alignment [2] and I don't think ncov needs to change here [1]. There is no merging necessary beyond what we currently have for multiple inputs, which happens after each has been aligned. 

--- 

[1] There is still the matter of the `rule filter` (in nextstrain/ncov) which will be the last remaining part of the "preprocessing" workflow. This can be dealt with separate to this PR and there are some decisions we need to make about what to do with that rule. 

[2] Things are different after we subsample, where we do use the peptides etc. But that's currently handled by a subsequent nextalign run _after_ subsampling.

ivan-aksamentov:
I added mutation summary to both full run and daily. Checked locally on a small dataset.

Now running the full run on AWS Batch to produce full alignment, nextclade.tsv and mutation_summary.tsv. If succesful, I will then wait for gisaid to update and run daily ingest to test the full cycle.

In theory, the additions to `ingest-gisaid` and `ingest-genbank` scripts can be pasted as steps to the Snakemake and should hopefully work.

The bulk of changes are in the full-run script, which is tricky due to batching and bash.

ivan-aksamentov:
Alright, the full jobs succeeded.

The job ids were:

```
8cf3776a-ad98-4c53-8d62-0b61ec3894af
64878851-e332-4970-a417-2a0125ebe03e
```

GISAID outputs are the

```
nextclade.aligned.fasta.xz
nextclade.mutation_summary.tsv.xz
nextclade.tsv.gz
```
under
```
s3://nextstrain-ncov-private/nextclade-full-run-2021-12-08--10-24-50--UTC/
```

After gisaid updates tomorrow I will try to run daily ingest script with these files locally to see if fasta and summary incrementally updates correctly.

ivan-aksamentov:
The mutation_summary results of local ingest seems to be okay:

```bash
$ wc -l data/gisaid/nextclade.mutation_summary.new.tsv | cut -f1 -d' '
32575

$ wc -l data/gisaid/nextclade.mutation_summary.old.tsv | cut -f1 -d' '
5718627

$ wc -l data/gisaid/nextclade.mutation_summary.tsv | cut -f1 -d' '
5751201
```

There is `old + new - 1` lines in the result. `-1` because we throw away the second header row.


At this point, @jameshadfield I can use your help in pasting that into the snakefile somehow.

bin/run-nextclade
```diff
@@ -40,6 +46,8 @@ echo "[ INFO] ${0}:${LINENO}: Running Nextclade"
   --input-fasta="${INPUT_FASTA}" \
   --input-dataset="${NEXTCLADE_DATASET_DIR}" \
   --output-tsv="${OUTPUT_TSV}" \
-  --genes=E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S \
+  --genes="${GENES}" \
   --output-dir="${NEXTCLADE_OUTPUTS_DIR}" \
-  --output-basename="nextclade"
+  --output-basename="nextclade" \
+  --output-fasta="${OUTPUT_FASTA}" \
+  --output-insertions="${OUTPUT_INSERTIONS}" \
```
ivan-aksamentov, line 53: `run-nextclade` script have gained 3 parameters:
 - OUTPUT_FASTA - this is where the newly aligned sequences end up
 - OUTPUT_INSERTIONS - same for insertions (needed for `mutation-summary` script)
 - GENES - list of genes. Nexclade is using comma-delimited list, while `mutation-summary` script prefers spacede-limited for some reason. We might make `mutation-summary` to reconsider its poor and random behavior :)

bin/run-nextclade
```diff
@@ -11,11 +11,17 @@ trap "exit" INT
...
+: "${6:?\"[ERROR] ${0}: Output insertions csv filename as the 6th argument is required.\"}"
+: "${7:?\"[ERROR] ${0}: List of genes as 7th argument is required.\"}"
 
 INPUT_FASTA="${1}"
 OUTPUT_TSV="${2}"
 NEXTCLADE_DATASET_DIR="${3}" # external data required for nextclade to function will be downloaded there
 NEXTCLADE_OUTPUTS_DIR="${4}" # files other than TSV will be written by Nextclade there (aligned sequences, peptides, etc.)
+OUTPUT_FASTA="${5}"
+OUTPUT_INSERTIONS="${6}"
+GENES="${7}"
```
ivan-aksamentov, line 24: `run-nextclade` script have gained 3 parameters:
 - OUTPUT_FASTA - "data/genbank/nextclade.aligned.new.fasta" - this is where the newly aligned sequences end up
 - OUTPUT_INSERTIONS - "data/genbank/nextclade.insertions.csv" - same for insertions (needed for `mutation-summary` script). The filename does not mention that they are new, because there is no "old" corresponding file, i.e. we don't store these on S3.
 - GENES - list of genes - "${GENES}" variable in the caller scripts -  This can be moved into config, similarly to how ncov does it. Note that Nexclade is using comma-delimited list, while `mutation-summary` script prefers space-delimited for some reason. We might make `mutation-summary` to reconsider its poor and random behavior :)  

The entire `bin/run-nextclade` should go away soon, because all it does is downloading the dataset and runs nextclade with the same exact parameters. It only confuses things.

ivan-aksamentov:
I merged master into this branch and which resolved (2) and (4) and in (1) the part with `bin/run-nextclade`. 

TODO:

 - The call sites need to be changed for the seconds part of (1).
 - And (3) of course - convert bash changes to snakemane changes.

With this merge I kept the `ingest-gisaid` and `ingest-genbank` scripts, so that they can be used as a reference for the new snakefile things. We should delete them after porting is done.

ivan-aksamentov:
I just realized that this merge removed the diff between the old and new ingest-gisaid and ingest-genbank scripts. They now all green, because they don't not exist on merged master.

ivan-aksamentov:
Alright, I sketched all of the changes that need to be ported in Snakemake file, but haven't actually run it yet. There are probably mistakes there.

TODO: 
 - [ ] run the Snakemake thing and see if it even runs with my changes
 - [ ] if not, eliminate the mistakes, make it run locally reliably
 - [ ] remove the bash versions of ingest scripts
 - [ ] try to run on AWS Batch

ivan-aksamentov:
For I few days I've been struggling with porting this thing to Snakemake. The workflow would fail when there is no new sequences: output of `rules.get_sequences_without_nextclade_annotations.outputs.fasta`  would be empty -> nextclade does not run in `rules.run_nextclade` and so `rules.compute_mutation_summary` have no required inputs.

Jover proposed to use checkpoints feature of Snakemake [here on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1639432579424300)
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution

I chatted with Richard and ha suggested an idea of emitting an empty file for `rules.get_sequences_without_nextclade_annotations.outputs.fasta` instead of a missing file. This would lead to a chain of rules that all generate empty outputs. However this way, the downloads and uploads always run (reuploading what was downloaded). In case of aligned fasta it's a large thing.

At this point Snakemake is a huge drain of time and delays an important feature, so I decided to just keep the whole bash script in the rule for now. This way, there are no rule dependency problems, because it's all in one rule. This is of course not at all what snakemake flow should be.

Any help and ideas are appreciated.

Let's see how this runs on AWS.

jameshadfield:
@ivan-aksamentov I've refactored this to use snakemake checkpoints. Let's watch the actions ([genbank](https://github.com/nextstrain/ncov-ingest/actions/runs/1580934003) and [gisaid](https://github.com/nextstrain/ncov-ingest/actions/runs/1580934000)) to see if it works as expected.

I have not reviewed this fully, but it looks generally good. Some things I noticed:

**shellcheck**
[is failing](https://github.com/nextstrain/ncov-ingest/runs/4529141573?check_suite_focus=true)

**Consistent fillenames**

It would help (me) a lot were we to systematise the filenames we use. My understanding, based on the PR, is that:
* “old” means the existing files / data on S3.
* “upd” refers to sequences / metadata present in this run but not present in the “old” version of that file
* “new” means “old” + “upd”

However this isn't always the case, e.g., sometimes the middle one is called "new".

ivan-aksamentov:
Hey thanks @jameshadfield,

> shellcheck is failing

Yeah, we need globbing in that case, so I thkn we just need to silence the warning on that line. But maybe there's some sort of a syntax for making it proper. It's bash, so you never know :) 

> It would help (me) a lot were we to systematise the filenames we use. 

Yeah, I wanted to do this as well. The reason why it's different for now is that it's different in the full-run script and on S3, so changing things might be a bit awkward. We need to go through these one by one and see what is affected.


Oh no! Looks like it failed
https://github.com/nextstrain/ncov-ingest/runs/4529141570?check_suite_focus=true#step:4:358

```
[batch] [Wed Dec 15 05:57:02 2021]
[batch] checkpoint get_sequences_without_nextclade_annotations:
[batch]     input: data/gisaid/sequences.fasta, data/gisaid/nextclade_old.tsv
[batch]     output: data/gisaid/nextclade.sequences.fasta
[batch]     jobid: 3
[batch] Downstream jobs will be updated after completion.
[batch]         ./bin/filter-fasta             --input_fasta=data/gisaid/sequences.fasta             --input_tsv=data/gisaid/nextclade_old.tsv             --output_fasta=data/gisaid/nextclade.sequences.fasta         
[batch] Updating job 1 (upload).
[batch] WorkflowError in line 423 of /nextstrain/build/Snakefile:
[batch] Can only use unpack() on list and dict
```

I will try to read and understand what's going on.

Snakefile
```diff
@@ -296,20 +401,30 @@ rule notify_genbank:
...
+        files_to_upload["additional_info.tsv.gz"] =     f"data/{database}/additional_info.tsv"
+        files_to_upload["flagged_metadata.txt.gz"] =    f"data/{database}/flagged_metadata.txt"
+
+    nextclade_sequences_path = checkpoints.get_sequences_without_nextclade_annotations.get().output.fasta
+    if os.path.getsize(nextclade_sequences_path) > 0:
+        files_to_upload["nextclade.tsv.gz"] =                  f"data/{database}/nextclade.tsv"
+        files_to_upload["nextclade.mutation_summary.tsv.gz"] = f"data/{database}/nextclade.mutation_summary.tsv"
+        files_to_upload["nextclade.aligned.fasta.xz"] =        f"data/{database}/nextclade.aligned.fasta"
+
+    return files_to_upload
```
ivan-aksamentov, line 422: I thing that the latest failure was due to missing return statement here. The `unpack()` on that return would then operate on a `None`. Let's see if it fixes things.

bin/run-nextclade-full
```diff
@@ -223,36 +233,88 @@ main() {
...
+    # Note we need genes variable to be unquoted
+    # shellcheck disable=SC2086
+    ./bin/mutation-summary \
+      --directory="${output_dir}" \
+      --alignment="${output_fasta_filename}" \
+      --basename="nextclade" \
+      --genes ${GENES_SPACE_DELIMITED} \
+      --reference="${TMP_DIR_NEXTCLADE_DATASET}/reference.fasta" \
+      --genemap="${TMP_DIR_NEXTCLADE_DATASET}/genemap.gff" \
+      --output="${output_mutation_summary}"
```
ivan-aksamentov, line 295: I also silenced the [SC2086](https://github-wiki-see.page/m/koalaman/shellcheck/wiki/Sc2086) warning for `GENES_SPACE_DELIMITED`. We need  it to be
```
--genes A B C
```
without quotations marks.

Snakefile
ivan-aksamentov: Alright, GenBank ran successfully the "now work" path if I understand correctly. Although I could not find confirmation for that in the logs. There is no mention of Nextclade though. `genbank.ndjson.xz` was identical, but `biosample.ndjson.gz` was not. I am not sure how biosample data is used. Is this correct behavior?

GISAID run failed with 5 download attempts, so inconclusive.

Let's try again? I merged master into this branch this time, to take advantage of Moira's cleanup in annotations, so the logs should not consist mostly of the "REDUNDANT ANNOTATED METADATA" messages.
 
https://github.com/nextstrain/ncov-ingest/actions/runs/1583829615
https://github.com/nextstrain/ncov-ingest/actions/runs/1583829620

joverlee521: In the successful GenBank run, I see the pipeline jumped directly from `checkpoint get_sequences_without_nextclade_annotations` to `rule generate_metadata`. If the `genbank.ndjson.xz` was identical, that means there were not any new sequences. The `biosample.ndjson.gz` only adds additional metadata, it does not contain any sequences. This worked as expected (at least for this path) 🎉

jameshadfield: Thanks for catching my missing `return`! 

Looks like we've seen it correctly skip steps (when there are no new sequences) so the next trial run should test out the other code path 🤞

Snakefile
```diff
@@ -298,7 +298,7 @@ rule mutation_summary:
             --alignment={input.alignment} \
             --insertions={input.insertions} \
             --reference={params.nextclade_input_dir}/reference.fasta \
-            --genemap={params.nextclade_input_dir}/genemap.gff" \
+            --genemap={params.nextclade_input_dir}/genemap.gff \
```
jameshadfield, line 301: 🤦

ivan-aksamentov:
The "quotation mark" jobs also failed, because, as I mentioned above, Nextclade accepts comma-delimited gene list, but mutation summary only accepts space-delimited list. And if not, then this happens:

```
[batch]   File "./bin/mutation-summary", line 107, in <module>
[batch]     gene = (set(fname.split('.')) & genes).pop()
[batch] KeyError: 'pop from an empty set'
```

I made it space delimited. Actions:
https://github.com/nextstrain/ncov-ingest/actions/runs/1587080622
https://github.com/nextstrain/ncov-ingest/actions/runs/1587080625

We could also modify mutation-summary script, but this is a bit out of scope of this PR and will make an exact copy of this script from ncov into an "almost exact" copy, which feels awkward to do. What is the best way to "commonize" this functionality across repos? (make it shared)

ivan-aksamentov:
In the latest news, the 
```
fix: use xz compression, not gz for mutation summary
```
succeede for genbank, but gisaid download keeps failing today ([Slack thread](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1639658495400200)), so again inconclusive result.

ivan-aksamentov:
Alright, GISAID have now ran successfully as well. I think this is ready to be reviewed. We may try to run this version in parallel to the master flow and compare the results.

Snakefile
```diff
@@ -183,49 +184,154 @@ rule get_sequences_without_nextclade_annotations:
...
+    params:
+        dst_source = config["s3_dst"] + '/nextclade.aligned.fasta.xz',
+        src_source = config["s3_src"] + '/nextclade.aligned.fasta.xz',
+    output:
+        alignment = temp(f"data/{database}/nextclade.aligned.old.fasta")
+    shell:
+        """
+        ./bin/download-from-s3 {params.dst_source} {output.alignment} ||  \
+        ./bin/download-from-s3 {params.src_source} {output.alignment}
+        """
```
ivan-aksamentov, line 251: Right now I am only concerned by this comment from James.
@jameshadfield Can you tell a little more about that?

> This file may be updated on the remote after download_nextclade has run but before this rule

So this means someone else would update the alignment on S3?

The relevant steps in the workflow are (roughly ordered):

 1. download from S3 the old nextclade.tsv
 2. start running Nextclade
 3. finish running Nextclade
 4. concatenate old and new nextclade.tsv
 5. upload nextclade.tsv
 6. download old aligned.fasta
 7. concatenate old and new aligned.fasta.tsv
 8. upload aligned.fasta

The steps 4-5 run concurrently with 6-8 and 6 runs concurrently with everything else, including 1. So if someone updates aligned.fasta on S3 after 1, but before 6, then the old nextclade.tsv is no longer in sync with old aligned.fasta. In fact, there are many more permutations in which things can go wrong.

The hope is that if someone updates the alignment.fasta, they'd also update the nextclade.tsv ***at the same time*** and won't update anything while daily ingest is running. Otherwise it can happen that the newly uploaded alignment.fasta and nextclade.tsv are overwritten by the workflow and the new data (e.g. the new clades from the full run) will be lost.

But I cannot think of who is that bad guy who silently updates S3 is. It could be Nextclade full run, but it produces files into a separate directory and then, after a manual verification, the copy to the "master" location is made by hand. So whoever does the copy need to make sure they do this at the right moment. But this was always true.

Everyone with write access to S3 should be aware of that and we need a "social mutex" mechanism to ensure consistency of operations :)


Again, the entire problem stems from the fact that we are using Snakemake, which introduces concurrency that we don't need. Is there a way to suppress it somehow?

(And another fundamental problem of course is that we treat a database a a set of files, and basically reinventing an RDBMS in a really shitty and slow way... We already have filtering, joins, etc. so now we need transactions, I guess)

jameshadfield: > Everyone with write access to S3 should be aware of that and we need a "social mutex" mechanism to ensure consistency of operations :)

Yes -- I think this race condition will rarely, if ever, be hit. I only expect one action to modify these files (the re-run full nextclade action).

> Again, the entire problem stems from the fact that we are using Snakemake

This also existed in the non-snakemake version.

ivan-aksamentov:
Last changes before merging:

Rename
 - nextclade.aligned.fasta -> aligned.fasta
 - nextclade.mutation_summary.tsv -> mutation-summary.tsv

i.e. no `nextclade.` prefix and mutation summary with a dash, not underscore. This is to align with the filenames on S3.

ivan-aksamentov:
The plan on bringing this to production:

 - [x] Make a Nextclade full run

    Done. AWS Batch Job IDs:

    ```
    b0d9fa0d-fdae-432a-8656-01435d1364f3
    ffcc6adc-6b6d-4f3c-bdc4-05a0a0d7e857
    ```

    Wait for result in

    ```
    s3://nextstrain-ncov-private/nextclade-full-run-2022-01-06--08-39-23--UTC/
    s3://nextstrain-data/files/ncov/open/nextclade-full-run-2022-01-06--08-39-22--UTC/
    ```

    Note that full fun was started before renaming. So when copying, I will also need to rename (see below).


 - [x] Copy results of a full run to the location of daily ingest, adjusting filenames to what's already on S3, and overwriting

    ```
    s3://nextstrain-ncov-private/nextclade-full-run-2022-01-06--08-39-23--UTC/nextclade.aligned.fasta.xz -> s3://nextstrain-ncov-private/aligned.fasta.xz

    s3://nextstrain-ncov-private/nextclade-full-run-2022-01-06--08-39-23--UTC/nextclade.mutation_summary.tsv.xz -> s3://nextstrain-ncov-private/mutation-summary.tsv.xz

    s3://nextstrain-ncov-private/nextclade-full-run-2022-01-06--08-39-23--UTC/nextclade.tsv.gz -> s3://nextstrain-ncov-private/nextclade.tsv.gz
    ```

    ```
    s3://nextstrain-data/files/ncov/open/nextclade-full-run-2022-01-06--08-39-22--UTC/nextclade.aligned.fasta.xz -> aligned.fasta.xz

    s3://nextstrain-data/files/ncov/open/nextclade-full-run-2022-01-06--08-39-22--UTC/nextclade.mutation_summary.tsv.xz -> s3://nextstrain-data/files/ncov/open/mutation-summary.tsv.xz

    s3://nextstrain-data/files/ncov/open/nextclade-full-run-2022-01-06--08-39-22--UTC/nextclade.tsv.gz -> s3://nextstrain-data/files/ncov/open/nextclade.tsv.gz

    ```

 - [x] Merge this PR


 - [x] Run "GISAID fetch and ingest" and "GenBank fetch and ingest" GitHub actions to perform daily ingest with the new code and fresh cache.

     Launched at 06:24 UTC

     https://github.com/nextstrain/ncov-ingest/actions/runs/1666402303
     https://github.com/nextstrain/ncov-ingest/actions/runs/1666402782

     AWS Batch Job IDs:

     ```
     045eaf99-4a4b-4186-a7bc-f8548a00638c
     7138b2aa-9702-4f1c-892c-425f2c19b00d
     ```

Issue #242, opened 2021-12-03
Revisit fetch from GISAID


joverlee521:
**Context**  
On Dec 2, 2021, multiple `fetch-and-ingest` runs for GISAID failed. The failure pattern was we would download for a while and the transfer would get closed before it's completed. Subsequent attempts to fetch would hit a 503 error. We manually triggered `fetch-and-ingest` two more times and saw the same failure pattern. 

**Possible solution**   
The scheduled run today had no issues, so this may have just been unfortunate timing of our runs being interrupted by GISAID's reboots. We can revisit the following solutions in anticipation of similar future issues:

1. Manual downloads from the same API endpoint were able to complete successfully when done without streaming decompression. We can update `fetch-from-gisaid` to stop decompression during streaming to lower the open connection time.  However, decompressing in a separate step this would increase the total time to run `fetch-and-ingest`. 
2. Switch to an endpoint with `xz`, which has better compression ratio and decompression time than `bzip2`. Regardless of errors, this would be a huge improvement for us and dramatically decrease `fetch-and-ingest` runtime.

ivan-aksamentov:
@joverlee521 

> Switch to an endpoint with `xz`.

I did not know it exists. Do you know the URL? Does it have the same data in it?

In the meantime we could try parallel bzip also: https://github.com/nextstrain/ncov-ingest/pull/247

tsibley:
> I did not know it exists. Do you know the URL? Does it have the same data in it?

Ah, it does not exist, as far as we know. This would be asking GISAID to switch to xz for us for the current export we get.

PR #245, opened 2021-12-04
GISAID: Decompress after download

tsibley:
This PR drafts a change we discussed [in Slack](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1638495448287000) and summarized in #242. Drafted now so it's ready in case we end up needing it sooner than later. If we never need it, ok to trash this PR.

---

Decompress gisaid.ndjson.bz2 after download instead of during

Streaming decompression during the network fetch leads to decreased network transfer rates[^1] and thus a longer network connection lifetime. As we sometimes see transient network disruptions that cause the transfer to fail, my hypothesis is that reducing connection lifetime by decompressing after download will meaningfully reduce our exposure to transient disruptions.

This may come at the cost of potentially increasing overall runtime by some unknown amount, but we could try to address that other ways such as switching to xz (with GISAID's cooperation) for faster decompression and/or performing decompression concurrently with downloading but using a disk spool in between (assuming disk io isn't a bottleneck).

[^1]: About 75% slower in a single test case I ran on my laptop over the Hutch's wired network.

---

[Update 10 Dec 2021] The concurrent-decompression-with-disk-spool approach could be implemented using [a program like this, dubbed bunzip2-tailf](https://gist.github.com/tsibley/aa16902ae53cd35196d280cf8d22d330) to do the decompression. As it'd be hard to reliably coordinate the concurrent processes (curl and bunzip2-tailf) with separate Snakemake rules/jobs, we'd want to use shell job handling instead. For example, updating `fetch-from-gisaid` to read something like:

```bash
# Download compressed file direct to disk to avoid slowing down the network transfer waiting on decompression.
curl … > gisaid.ndjson.bz2 &

# Decompress from disk to stdout on the fly; will wait for more data to appear in the file if it hits EOF (up to a time limit).
bunzip2-tailf < gisaid.ndjson.bz2 &

# Wait for both background jobs above to complete.
wait
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Relies on data from https://simplemaps.com/data/us-cities.
 ## Running locally
 If you're using Pipenv (see below), then run commands from `./bin/…` inside a `pipenv shell` or wrapped with `pipenv run ./bin/…`.
 
-1. Run `./bin/fetch-from-gisaid > data/gisaid.ndjson`
+1. Run `./bin/fetch-from-gisaid | bunzip2 > data/gisaid.ndjson`
 2. Run `./bin/transform-gisaid data/gisaid.ndjson`
 3. Look at `data/gisaid/sequences.fasta` and `data/gisaid/metadata.tsv`
 
--- a/Snakefile
+++ b/Snakefile
@@ -54,7 +54,7 @@ if config.get("fetch_from_database", False):
         message:
             """Fetching data using the database API"""
         output:
-            ndjson = temp(f"data/{database}.ndjson")
+            ndjson = temp(f"data/{database}.ndjson" + (".bz2" if database == "gisaid" else ""))
         run:
             run_shell_command_n_times(
                 msg = f"Fetching from {database}",
@@ -74,6 +74,13 @@ else:
             ./bin/download-from-s3 {params.file_on_s3_dst} {output.ndjson} ||  \
             ./bin/download-from-s3 {params.file_on_s3_src} {output.ndjson}
         """
+
+
+rule bunzip2:
+    message: "Decompressing {input}"
+    input: "{stem}.bz2"
+    output: "{stem}"
+    shell: "bunzip2 {input:q}"
 
 
 rule notify_on_database_change:
--- a/Snakefile
+++ b/Snakefile
@@ -51,7 +51,6 @@ rule download_main_ndjson:
     message:
         """Fetching data using the database API"""
     params:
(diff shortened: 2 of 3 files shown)


ivan-aksamentov:
@tsibley Is there a possibility to incorporate parallel unbzipping also?
https://github.com/nextstrain/ncov-ingest/pull/247/files

tsibley:
@ivan-aksamentov Yes, certainly, if we do decompression after download. If we do streaming decompression concurrent with download using an on-disk spool, then _maybe_, but not with the `bunzip2-tailf` command I wrote.

tsibley:
So I think the practical questions for a decision are: how much longer than downloading straight to disk does single-threaded decompression from disk take? if longer, how does it compare to parallel decompression from disk?

My guess is that concurrent decompression might win out, even though single-threaded, but not sure. Should be easy enough to benchmark.

ivan-aksamentov:
At this point we don't have ingest results and so builds are pretty much cancelled. Which probably makes the fastest implementable and the most reliable solution (the one that won't fail) the best in my view.

We could also try to run both on separate branch runs and see which completes first. And then if at least one of them succeeds we could kick start the preprocess and builds manually hopefully if scientists find that they need them.

The good new is that it's the first stage in the pipe, so it fails fast.

ivan-aksamentov:
I also wonder of there's a third-party tool or if you could adjust your tool, so that it does the same spool trick, but would allow to pipe the output to the parallel bzip (or potentially to xz when it's available later). I.e. to decouple the trick itself from the compression algorithm.

tsibley:
> At this point we don't have ingest results and so builds are pretty much cancelled. Which probably makes the fastest implementable and the most reliable solution (the one that won't fail on the first run) the best in my view.

Well, none of these are definitely more reliable. The issue appears to be wholly on the network/upstream server, not within our decompression. The guess this is all predicated on is that reducing network transfer time will help avoid the issue.

> I.e. to decouple the trick itself from the compression algorithm. Then we could take advantage of both, the concurrent download and parallel decompress.

You could do this with normal `tail`, I think. I combined the spool + decompression into a single process to avoid the downside of more processes having to read/write the whole data stream, but maybe that doesn't matter.

tsibley:
> You could do this with normal `tail`, I think.

I'm writing this and will open another PR for it.

tsibley:
That was a little more involved than I expected, but try #256.

PR #253, opened 2021-12-14
chore: update local dev setup

ivan-aksamentov:
This refreshes the local dev setup now that we use snakemake (https://github.com/nextstrain/ncov-ingest/pull/231)

 - [x] Revamp the Dockerfile. It is no longer needed for production, so I think it is dev-only. I could not find the source for the [base image](https://hub.docker.com/r/nextstrain/base/tags?name=branch-python-base)  and we don't need it anyways.
 - [x] Add helper script to run the container
 - [x] Add missing `snakemake` package to `Pipfile`
 
This setups is what I used for https://github.com/nextstrain/ncov-ingest/pull/237

I cancelled GHA because this PR only affects local dev flow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,28 +28,26 @@
 # version becomes the new default, if this ncov-ingest image itself is still
 # relevant at that time.
 #   -trs, 19 Jan 2020
-FROM nextstrain/base:branch-python-base
 
-# Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
-RUN apt-get update && apt-get install -y --no-install-recommends \
-        python3-netifaces \
-        time\
-        xz-utils
+# Let's drop dependency on an image of unknown origin and start over.
+#   -ivan.aksamentov 2021-12-13
+FROM python:3.7-buster
+
+SHELL ["/bin/bash", "-c"]
+
+RUN set -euxo pipefail \
+&& apt-get update \
+&& apt-get install -y --no-install-recommends \
+  build-essential \
+  time \
+  xz-utils
+
+COPY Pipfile Pipfile.lock /workdir/
 
 # Install Python deps
-RUN python3 -m pip install pipenv
-COPY Pipfile Pipfile.lock /nextstrain/ncov-ingest/
-RUN PIPENV_PIPFILE=/nextstrain/ncov-ingest/Pipfile pipenv sync --system
-
-# Install Nextclade
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
-         -o /usr/local/bin/nextclade \
- && chmod a+rx /usr/local/bin/nextclade
-
-# Install Nextclade C++
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
-         -o /usr/local/bin/nextclade \
- && chmod a+rx /usr/local/bin/nextclade
+RUN set -euxo pipefail \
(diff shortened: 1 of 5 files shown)


bin/docker-run
```diff
@@ -0,0 +1,45 @@
...
+
+declare -a config
+config+=(
+  database_name="gisaid"
+  s3_src="s3://nextstrain-ncov-private/branch/${GITHUB_BRANCH}"
+  s3_dst="s3://nextstrain-ncov-private/branch/${GITHUB_BRANCH}"
+  fetch_from_database=false
+  trigger_preprocessing=false
+  keep_all_files: true
+)
```
ivan-aksamentov, line 19: Note that, as always, because it interacts with prod data, it's extremely easy to mess things up in production, by just setting a wrong S3 url here. So be careful.

Dockerfile
```diff
@@ -28,28 +28,26 @@
...
-FROM nextstrain/base:branch-python-base
 
-# Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
-RUN apt-get update && apt-get install -y --no-install-recommends \
-        python3-netifaces \
-        time\
-        xz-utils
+# Let's drop dependency on an image of unknown origin and start over.
+#   -ivan.aksamentov 2021-12-13
+FROM python:3.7-buster
```
victorlin, line 34: I'm a little confused on historical context here, but maybe @tsibley's TODO comment can be removed.

joverlee521: The [python-base has been merged](https://github.com/nextstrain/docker-base/pull/21), so I think we can just use `nextstrain/base:latest`.

ivan-aksamentov: A-ha! The mystery of a missing container image solved! 
We don't need anything from the base image though, and pretty much anything other than python and pipenv.

joverlee521 (commented):
Thanks for prompting this update Ivan! It's been a ~while since anyone has touched this Dockerfile.

>Revamp the Dockerfile. It is no longer needed for production, so I think it is dev-only. I could not find the source for the base image and we don't need it anyways.

The GitHub actions [use this docker image](https://github.com/nextstrain/ncov-ingest/blob/master/.github/workflows/fetch-and-ingest-gisaid-master.yml#L50) so it's not just dev-only, but I definitely agree it should be updated. I think we can use `nextstrain/base:latest` with additional dependencies on top. 

> Add helper script to run the container

For the `bin/docker-run` script, I think it would be best to mimic how the pipeline runs in production by using the nextstrain cli (this might also just be me not wanting to deal directly with docker 😅 )

> Add missing snakemake package to Pipfile

The pipeline has been running fine without explicitly adding snakemake here...(Aha, snakemake is [included in the nextstrain base image](https://github.com/nextstrain/docker-base/blob/master/Dockerfile#L68))

ivan-aksamentov:
@joverlee521 Oh no! I overlooked that it's used in prod. In this case, nothing here makes sense.

I deal with ingest a lot lately and it's not a great dev experience, so I sketched this minimal thing that runs and a more convenient dev flow, where I can modify things quickly. I could not even build the old image, because it uses apt repos from Debian version that is too old. I also don't know how the docker option of nextstrain cli works. Only used the AWS Batch path so far. I might try it sometimes and see how it goes.

victorlin:
Still worthwhile to change `nextstrain/base:branch-python-base` to `nextstrain/base:latest` though?

ivan-aksamentov:
@victorlin Probably. Might also mek dockerignore more restrictive like here. It is currently copying the stuff from the repo, including `source-data/` which is heavy, while only 2 pipenv files are needed during build.

Issue #250, opened 2021-12-10
Remove "REDUNDANT ANNOTATED METADATA" messages from technical logs


ivan-aksamentov:
```
[batch] REDUNDANT ANNOTATED METADATA : LR757995 division Hubei
[batch] REDUNDANT ANNOTATED METADATA : LR757995 location Wuhan
[batch] REDUNDANT ANNOTATED METADATA : LR824035 country Switzerland
[batch] REDUNDANT ANNOTATED METADATA : LR824035 region Europe
[batch] REDUNDANT ANNOTATED METADATA : MZ918713.1 gisaid_epi_isl EPI_ISL_2423684
```

It seems that this is domain-specific info that should be read by ingest build shepherds. But right now these messages only clog the logs on Github Actions and AWS Batch making diagnostics harder.

Attached is 9 megs (!) of these messages from genbank ingest logs.

I propose we remove those or redirect them to a file, where shepherds can find them.

Thread on Slack: https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1639149188358600

Attachments:
[REDUNDANT ANNOTATED METADATA.txt](https://github.com/nextstrain/ncov-ingest/files/7693706/REDUNDANT.ANNOTATED.METADATA.txt)

emmahodcroft:
Agree with Ivan - let's pipe this to a file that can be used by build shepherds - perhaps the GISAID one can go to Slack. The Genbank one maybe not, as it's so huge. (But having two different actions may be more difficult than it's worth)

PR #254, opened 2021-12-16
fix: prevent genbank ingest log spam

ivan-aksamentov:
Resolves https://github.com/nextstrain/ncov-ingest/issues/250

The stdout of GenBank transform is currently printed to stdout of AWS Batch. This PR redirects it to a file. The file is then uploaded to Slack. This is similar to how it's handled for GISAID.

This facilitates reading the logs for techies and makes it more convenient to get domain-specific output for scientists.

The code is a copypasta from the similar GISAID code. I hope there are no typos :)

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -129,6 +129,7 @@ rule transform_genbank_data:
     output:
         fasta = "data/genbank/sequences.fasta",
         metadata = "data/genbank/metadata_transformed.tsv",
+        flagged_annotations = temp("data/genbank/flagged-annotations"),
         duplicate_biosample = "data/genbank/duplicate_biosample.txt"
     shell:
         """
@@ -136,7 +137,7 @@ rule transform_genbank_data:
             --biosample {input.biosample} \
             --duplicate-biosample {output.duplicate_biosample} \
             --output-metadata {output.metadata} \
-            --output-fasta {output.fasta}
+            --output-fasta {output.fasta} > {output.flagged_annotations}
         """
 
 rule transform_gisaid_data:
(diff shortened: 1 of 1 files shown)


emmahodcroft:
This sounds like a great idea Ivan. I'm not best to review the code in detail, but the idea is good. 

@MoiraZuber important that this sounds like it'll work for you too! If you have any questions or want to clarify with Ivan about how this works, don't hesitate to ask!

victorlin (approved):

joverlee521 (approved):
Looks good from inspecting the code. I didn't find any typos 👍

Snakefile
```diff
@@ -129,14 +129,15 @@ rule transform_genbank_data:
...
+        flagged_annotations = temp("data/genbank/flagged-annotations"),
         duplicate_biosample = "data/genbank/duplicate_biosample.txt"
     shell:
         """
         ./bin/transform-genbank {input.ndjson} \
             --biosample {input.biosample} \
             --duplicate-biosample {output.duplicate_biosample} \
             --output-metadata {output.metadata} \
-            --output-fasta {output.fasta}
+            --output-fasta {output.fasta} > {output.flagged_annotations}
```
corneliusroemer, line 140: Should we redirect stderr too?

Or even `tee` it so it's visible in Slack and in the logs? That might be most useful. We want to know if there are errs.

ivan-aksamentov: @corneliusroemer The goal of the PR is to add that 1 statement that was forgotten when copied from gisaid script, ant to allow engineers to read logs.

I'd avoid messing with stderr here, but we may consider it if it is needed in a separate PR. In that case we'd probably better tp upload error logs for all stages and all databases, not just this one thing. Transform stage is actually somewhat stable and unlikely to emit errors.

PR #255, opened 2021-12-16
Update repository dispatch triggers

jameshadfield:
This change is a companion to https://github.com/nextstrain/ncov/pull/823 which removes the preprocessing pipelines.


jameshadfield:
See https://github.com/nextstrain/ncov/pull/814 for related discussion

PR #256, opened 2021-12-17
Spool GISAID download to disk while still decompressing concurrently

tsibley:
Streaming decompression pipelined with the network fetch leads to
decreased network transfer rates¹ and thus a longer network connection
lifetime.  As we sometimes see transient network disruptions that cause
the transfer to fail, my hypothesis is that reducing connection lifetime
by decompressing from a disk spool instead will meaningfully reduce our
exposure to transient disruptions.

Caveat emptor: I haven't benchmarked this, and it's possible that
pushing all data through Python (halting-tailf) will add more runtime
overhead than is saved.

For another speed boost, try replacing bunzip2 with lbunzip2 in a future
change.

h/t @ivan-aksamentov for the suggestion to decouple the spooling from
the decompression and do it in two separate processes instead of one
(as I did in my proof-of-concept bunzip2-tailf).

¹ About 75% slower in a single test case I ran on my laptop over the
  Hutch's wired network.

----

Alternative to #245.

### DIFF
--- a/bin/fetch-from-gisaid
+++ b/bin/fetch-from-gisaid
@@ -4,7 +4,34 @@ set -euo pipefail
 : "${GISAID_API_ENDPOINT:?The GISAID_API_ENDPOINT environment variable is required.}"
 : "${GISAID_USERNAME_AND_PASSWORD:?The GISAID_USERNAME_AND_PASSWORD environment variable is required.}"
 
+bin="$(dirname "$0")"
+spool="$(mktemp -t gisaid-XXXXXX.ndjson.bz2)"
+semaphore="$(mktemp)"
+
+# Remove temporary files on exit, regardless of success or failure
+trap "rm -f ${spool@Q} >/dev/null" EXIT
+
+# Download compressed file direct to disk to avoid slowing down the network
+# transfer waiting on decompression.
 curl "$GISAID_API_ENDPOINT" \
     --user "$GISAID_USERNAME_AND_PASSWORD" \
     --fail --silent --show-error --location-trusted --http1.1 \
-    | bunzip2
+    > "$spool" \
+&
+
+# Decompress from disk to stdout on the fly; will wait for more data to appear
+# in the file if it hits EOF.
+"$bin"/halting-tailf "$semaphore" < "$spool" | bunzip2 &
+
+# After the download completes, signal halting-tailf via removal of the
+# semaphore file that it should halt after the next EOF instead of waiting for
+# more data.
+#
+# This wouldn't be necessary if bunzip2 could be configured to stop (and thus
+# break the pipe from tail) when it finds the end of stream marker in the
+# compressed data instead of waiting for another compression stream.
+wait %curl
+rm -f "$semaphore"
+
+# Wait for everything to wrap up.
+wait
--- a/bin/halting-tailf
+++ b/bin/halting-tailf
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+"""
+usage: ... | halting-tailf SEMAPHORE-FILE | ...
+
+Reads stdin, retrying on EOF (e.g. like tail -f) as long as SEMAPHORE-FILE
+still exists, and writes to stdout.  When SEMAPHORE-FILE is removed, exits the
+next time it hits EOF on stdin.
+"""
+import sys
+from pathlib import Path
+from time import sleep
+
+
+def main(semaphore_file):
+    semaphore = Path(semaphore_file)
+
+    def halt_after_semaphore_removed():
+        if not semaphore.exists():
+            return True
+        else:
+            sleep(1)
+            return False
+
+    try:
+        for chunk in tailf(sys.stdin.buffer, halt_after_semaphore_removed):
+            sys.stdout.buffer.write(chunk)
+    except Halted:
+        pass
+
+
+def tailf(stream, halt = lambda: False):
+    while True:
+        try:
+            yield next(stream)
+        except StopIteration:
+            if halt():
+                raise Halted()
+            else:
+                continue
+
+
+class Halted(Exception):
+    pass
+
+
+if __name__ == "__main__":
+    main(*sys.argv[1:])


ivan-aksamentov:
Looks like it worked. AWS Batch job:
```
4ae2af61-5401-454b-8d8e-c328e3c23af1
```

```
[batch] [Fri Dec 17 01:29:23 2021]
[batch] rule download_nextclade:
[batch]     output: data/gisaid/nextclade_old.tsv
[batch]     jobid: 9
[batch]         ./bin/download-from-s3 s3://nextstrain-ncov-private/branch/trs/gisaid/concurrent-decompression/nextclade.tsv.gz data/gisaid/nextclade_old.tsv ||          ./bin/download-from-s3 s3://nextstrain-ncov-private/nextclade.tsv.gz data/gisaid/nextclade_old.tsv
[batch]         
[batch] [Fri Dec 17 01:29:23 2021]
[batch] Job 7: Fetching data using the database API
[batch] Job counts:
[batch]   count jobs
[batch]   1 download_main_ndjson
[batch]   1
[batch] ./bin/fetch-from-gisaid > data/gisaid.ndjson
[batch] [ INFO] Downloading s3://nextstrain-ncov-private/branch/trs/gisaid/concurrent-decompression/nextclade.tsv.gz → data/gisaid/nextclade_old.tsv
[batch] download failed: s3://nextstrain-ncov-private/branch/trs/gisaid/concurrent-decompression/nextclade.tsv.gz to - An error occurred (404) when calling the HeadObject operation: Not Found
[batch] [ INFO] Downloading s3://nextstrain-ncov-private/nextclade.tsv.gz → data/gisaid/nextclade_old.tsv
[batch] [Fri Dec 17 01:30:18 2021]
[batch] Finished job 9.
[batch] 1 of 10 steps (10%) done
[batch] Fetching from GISAID API attempt number 0
[batch] [Fri Dec 17 03:11:57 2021]
[batch] Finished job 7.
```

The job 7 ran from 01:29:23 to 03:11:57 (1 hour, 42 minutes, and 34 seconds, 102.57 minutes, 6,154 seconds). My understanding is that when snakemake prints that job is finished, it means both curl and bzip have finished and exit successfully, so we can use this number as a benchmar. Correct?

Overall I think this solution is nice, but tricky. In particular when IPC is involved I am always suspicious about various edge cases. Like one process fails and not the other. Or inputs fail. Or outputs. In my local testing:

 - I tried to unplug the internet connection to see how it reacts to failures, e.g. GISAID going down for a minute or similar intermittent disruptions. It seems that the script keeps running without connection and does not give any indication of failure. I waited for a few minutes expecting perhaps some sort of timeout. The retry also does not kick-in either. It also does not continue downloading after connection is restored. So if GISAID blinks for a second, we might then wait for hours thinking ingest is running, but it might never complete. Some more digging might be required. My hope right now is that server failure might affect the script differently then my somewhat artificial simulation of yanking the internet plug.
 - It does fail and retries correctly if the connection fails to be established in the beginning. Which is good.
 - A minor inconvenience is that the current reaction to sigint is unexpected - it kills the curl, and requires second sigint to kill the rest of the bash script. And after that then snakemake tries to make another download attempt. So yet another sigint is needed to kill the whole thing. This might come up when cancelling jobs on actions and batch. But hopefully spamming a few times the button in the UI and Ctrl+C in the batch cli should work. Maybe a sigint handler could help here? But that's the least of our worries right now I think.
 - I haven't tested for failures on the other side of the pipe. I imagine situation where bzip may run out of disk space for example. And the temp dir may run out of space too. Not sure what will happen in this case. I'd be happy if it failed fast with a big kaboom, so that we could take an action. But right now the disk is big enough to not worry about this particular situation, for a few months at least.

tsibley:
Thanks for testing those edge cases. I agree IPC is tricky in this regard, and is the reason I used a semaphore file instead of signals because I wanted to avoid the edge cases around signals for IPC. :-) But that doesn't mean all are avoided.

I didn't explicitly try the tests you mentioned, but I did, for example, Ctrl-C the process many many times during dev and never hit the behaviour you described.

> One caveat is that I tested everything in docker.

Docker famously messes up signal handling in various ways both within the container and from outside → inside the container when great care is not taken. I don't know how you ran this in Docker, but I would not be surprised if the local Docker invocation you used is actually the culprit here for many if not all the behaviour you observed. (Though to be fair, I also wouldn't be surprised if it wasn't, because IPC is indeed tricky like you say.)

> In the end I am not sure what review to submit. On one hand we need download to be working here and now and this solution works on the happy path. On the other hand, we might not know if something fails or not and I don't see immediate fix myself.

Nod. I have no problem abandoning this PR in favor of #258 if you think that's better. It's certainly faster!

I don't know if there are any meaningful additional gains to be made by re-applying the disk spool technique in this PR to the aria2c + lbzip2 combination, but that's also an option. If the download is fast enough, probably not.

ivan-aksamentov:
> I don't know if there are any meaningful additional gains to be made by re-applying the disk spool technique in this PR to the aria2c + lbzip2 combination

Download seems to be fast enough currently with aria, but if there are any troubles with multi-connection downloads, then we will need to fall back to curl and then the tailf trick may become the best solution again. But this time with lbzip2.

tailf is probably not possible to use with aria though, because it downloads and writes files in chunks. But if it is, then we could shave off another 1-2 mins probably which is a ~10% gain considering [12 min total time](https://github.com/nextstrain/ncov-ingest/pull/258#issuecomment-997064228) measured in prod.

tsibley:
> tailf is probably not possible to use with aria though, because it downloads and writes files in chunks.

Ah, I was assuming it wrote those chunks in order to the final output file… but maybe not?

My understanding is that to write them out of order, it would either have to pre-allocate the "gaps" with nulls (takes some time because its still a disk write) or use special sparse file support (but this is filesystem/platform dependent, and so hard to make portable). Maybe there's another way, though.

ivan-aksamentov:
I actually don't know how it works. It supports multiple preallocation methods. Might be related
https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-file-allocation

tsibley:
Ah, so looks like aria2 supports all the methods I mentioned: no pre-allocation (and so presumably output is written in order), pre-allocation of the whole file (and so presumably written out of order), and sparse allocation using filesystem system support (and so also presumably written out of order). We can't use the tailing approach if it's writing out of order, for sure.

tsibley:
Closing this without merging as it seems unnecessary. We can always re-visit later.

PR #247, opened 2021-12-10
feat: use parallel version of bzip2 to decompress gisaid snapshot

ivan-aksamentov:
### Description of proposed changes    
I don't know if it's any faster, but why not.

Locally, it does use multiple threads, but not too many. We might be bound by download speed rather then decompression though.

### Related issue(s)  

Fixes #  
Related to https://github.com/nextstrain/ncov-ingest/issues/242

### Testing

This is a drop-in replacement. The results are correct in my local testing.

PR #257, opened 2021-12-17
chore: cleanup docker container image

ivan-aksamentov:
This:

 - Removes duplicated nextclade downloads. Nobody knows what version they are and when they are updated. Latest version of nextclade is already downloaded in the `./bin/run-nextclade` and `./bin/run-nextclade-fill` scripts. We want it to always be cutting edge. Nextclade releases are frequent. We may consider bringing it back into the image later if we setup automated rebuild of the image on Nextclade release. However I also often play with different custom versions of Nextclade in full runs and having other versions may cause mistakes. Mixed feelings here.

 - Adds various tools we might want to play with on branches:
    - aria2 - for downloading on multiple connections (for [#258](https://github.com/nextstrain/ncov-ingest/pull/258))
    - lbzip2  - for parallel unbzipping  (for [#247](https://github.com/nextstrain/ncov-ingest/pull/247) and [#258](https://github.com/nextstrain/ncov-ingest/pull/258))
    - pigz - for parallel gz/ungz
    - pixz - for parallel xz/unxz

 - Adds `build-essential` package. Pipenv insists on building the `netifaces` package from scratch now and if there's no GCC, the following error is produced: [error.txt](https://github.com/nextstrain/ncov-ingest/files/7735449/error.txt). I am not sure why `netifaces` are needed and whether there's a better solution.

 - Changes base image to `nextstrain/base` as suggested in https://github.com/nextstrain/ncov-ingest/pull/253. We don't use much out of it, so I am not super convinced it's needed, but not a big deal. Hopefully it will update the apt cache, so I can install deps again :) 

 - Extends dockerignore to everything but the 2 files we need. There is about 100 megs of data in the repo that gets transferred currently into the build context. If this is accepted, then later on, if you need to COPY or ADD something, don't forget to whitelist that thing in the dockerignore (by adding `!thing` line).

Current inconvenience is that for any changes in deps on branches (e.g. if I want to play with aria2c and lbzip2) we need to change the container on master and/or publish an image and change the tag in the GitHub Action on master. Which both disrupt production. We might consider installing deps dynamically on AWS batch instead. I tired and it failed because currently the debian repo cache in the image is too old. Or maybe there's a better way to play with new tools right now? In any case, mixing prod and dev as it is now seems weird and unproductive, we might think about a dedicated, simpler dev flow.

Related to
 - https://github.com/nextstrain/ncov-ingest/pull/247
 - https://github.com/nextstrain/ncov-ingest/pull/258

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
-/data
+*
+!Pipfile
+!Pipfile.lock
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ FROM nextstrain/base
 # Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
 RUN apt-get update && apt-get install -y --no-install-recommends \
     aria2 \
+    build-essential \
     lbzip2 \
     pigz \
     pixz \
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,33 +23,22 @@
 # ncov-ingest themselves.  This means the image only needs to be updated when
 # dependencies change, not when any pipeline change is made, and thus image
 # updates can be far less frequent.
-
-# XXX TODO: This can be updated to :latest eventually when the python-base
-# version becomes the new default, if this ncov-ingest image itself is still
-# relevant at that time.
-#   -trs, 19 Jan 2020
-FROM nextstrain/base:branch-python-base
+FROM nextstrain/base
 
 # Install Python package for which Python 3.7 wheels do not yet exist on PyPI.
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        python3-netifaces \
-        time\
-        xz-utils
+    aria2 \
+    lbzip2 \
+    pigz \
+    pixz \
+    python3-netifaces \
+    time \
+    xz-utils \
 
 # Install Python deps
 RUN python3 -m pip install pipenv
 COPY Pipfile Pipfile.lock /nextstrain/ncov-ingest/
 RUN PIPENV_PIPFILE=/nextstrain/ncov-ingest/Pipfile pipenv sync --system
-
-# Install Nextclade
-RUN curl -fsSL https://github.com/nextstrain/nextclade/releases/latest/download/nextclade-Linux-x86_64 \
-         -o /usr/local/bin/nextclade \
- && chmod a+rx /usr/local/bin/nextclade
-
-# Install Nextclade C++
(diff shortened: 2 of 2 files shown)


ivan-aksamentov:
I cancelled CI runs, because they still run an image from master, which is not informative for this PR.

tsibley:
I haven't looked at this closely, but a couple quick notes:

> Adds `build-essential` package. Pipenv insists on building the `netifaces` package from scratch

The netifaces situation is a known issue we've been meaning to get around to fixing, and so I just did. It will be resolved by https://github.com/nextstrain/cli/pull/138 once released, meaning we won't have to include the whole build-essentials toolchain.

> We don't use much out of [the nextstrain/base image], so I am not super convinced it's needed, but not a big deal.

It contains critical parts for running on AWS Batch under `nextstrain build`, so we do need it. Thanks for updating from the old `branch-python-base` to `latest` tag!

> Current inconvenience is that for any changes in deps on branches…

My suggestion here is to extend the [update-image GitHub Actions workflow](https://github.com/nextstrain/ncov-ingest/blob/8ac7fd91fe7c5743636b1fcee432ddd34ba7c43e/.github/workflows/update-image.yml#L6) to build images for branches and push them with `branch-*` tags like those used by the `nextstrain/base` image CI, e.g.:

https://github.com/nextstrain/docker-base/blob/b3ba871576a40f1e30fbb0a1211b3a44e5de56f1/devel/push-from-travis#L20-L25

I wanted to do this when I originally introduced the `nextstrain/ncov-ingest` image, but didn't due to time constraints (needed to Just Get It Done Now to lift-and-shift compute to AWS). It makes the dev cycle a bit nicer to have CI do the image building which you can then pull down locally or refer to in AWS Batch runs or have someone else try out.

With per-branch images, we could then also adjust the per-branch ingest workflows use those images for the AWS Batch jobs if they exist.

ivan-aksamentov:
Thanks for comments! This makes sense.

I seek for review/suggestions/approval to this PR in order to merge ASAP and to unblock the https://github.com/nextstrain/ncov-ingest/pull/258, which should bring improvement to the current situation with GISAID snapshot downloads.

Slack thread: https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1639754616496800

tsibley (approved):
This looks good to me.

Ideally we'd follow up to remove `python3-netifaces` and `build-essentials` once https://github.com/nextstrain/cli/pull/138 is merged and the `nextstrain/base:latest` image updated.

PR #258, opened 2021-12-17
perf: use aria2c and lbzip2 for faster* downloads

ivan-aksamentov:
We aretrying to shorten the connection time to GISAID while downloading the snapshot, and potentially speed things up in general. Best explained in https://github.com/nextstrain/ncov-ingest/pull/256

This PR:

 - splits download and decompression into 2 processes running sequentially
 - uses aria2c as downloader. Advantages:
   - concurrent downloads of multiple chunks on multiple connections
   - resumeable downloads (GISAID seems to be supporting it from my local testing) - download will continue from where it's left, for example after intermittent failure
   - built-in retry (so we might remove the handmade solution)
   - periodic progress, bandwidth and ETA reports
 - lbzip2 as a parallel decompressor for bz2
 - changes the snakemake rule so that the decompressed output is written to a file, by lbzip2, not redirected from the script stdout, so that the progress reports and warnings can be read

~!BLOCKED! on https://github.com/nextstrain/ncov-ingest/pull/259. In order to run this, `aria2c` and `lbzip2` should be in the container.~ Unblocked now.

P.S. I use `aria2c` as my daily driver for years now and it has never let me down.

*- Speedup is to be verified, but I think at this point any working solution would be great. Upd: Results are in [this Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1639754616496800).

---

Solutions to compare:

 - [Spool GISAID download to disk while still decompressing concurrently #256](https://github.com/nextstrain/ncov-ingest/pull/256)
 - [perf: split curl and bunzip2 when downloading from gisaid #259](https://github.com/nextstrain/ncov-ingest/pull/259) 
 - [feat: use parallel version of bzip2 to decompress gisaid snapshot #247](https://github.com/nextstrain/ncov-ingest/pull/247) 
 - [perf: use aria2c and lbzip2 for faster* downloads #258](https://github.com/nextstrain/ncov-ingest/pull/258) <-- you are here


ivan-aksamentov:
Currently running on AWS Batch. Job IDs are:
```
e4531fae-69b4-4d04-9b51-cfa59d1f38f5
8aa69496-6cba-4b92-b8de-0e442ff09a9f
```

ivan-aksamentov:
There are also warning messages
```
[batch] 12/17 22:06:44 [WARN] aria2c had to connect to the other side using an unknown TLS protocol. The integrity and confidentiality of the connection might be compromised.
```
This is probably due to aria2c and gnutls in Debian 9 being too. They are not aware about TLS 1.3 ([1](https://unix.stackexchange.com/questions/502763/aria2c-had-to-connect-to-the-other-side-using-an-unknown-tls-protocol-why), [2](https://github.com/aria2/aria2/issues/1464))

```
$ aria2c --version
aria2 version 1.34.0
...
Libraries: ... GnuTLS/3.6.6 ...
...
```

Not sure what to do about it. But upgrading base image to to Debian 10 would help.

huddlej (commented):
This looks really reasonable to me. The only minor issue I noted is that [the README.md still refers to the old API](https://github.com/nextstrain/ncov-ingest/blame/feat/gisaid-download-aria2c-lbzip/README.md#L9) for `fetch-from-gisaid`.

When you say we could use aria2c's retry logic, do you mean we could replace `run_shell_command_n_times` in the Snakefile with a flag for aria2c? If so, this seems like a win for readability and maintainability.

ivan-aksamentov:
> we could replace run_shell_command_n_times in the Snakefile with a flag for aria2c?

Yes, we could probably remove the external retry logic from GISAID script. I'd make it in a followup PR though.

ivan-aksamentov:
> The only minor issue I noted is that the README.md still refers to the old API for fetch-from-gisaid.
I adjusted readme

PR #261, opened 2021-12-17
perf: let's try parallel xz and gz

ivan-aksamentov:
This is more of an experiment, then a real thing. This may or may not give a little speedup.

`pixz` has been a parallel alternative to `xz` for some time before `xz` finally implemented -T flag for compression. And it has parallel decompression, while `xz -d`, ignores this flag until very recent versions (and we use a pretty old Debian distro).

`pigz` is a parallel `gz` from the author of the algorithm himself. Definitely faster, but we only use `gz` on smaller files, will not necessarily be noticeable, but why not squeeze it anyways.

Both `pixz` and `pigz` are compatible with their respective "default" tool, in both directions.

### DIFF
--- a/bin/download-from-s3
+++ b/bin/download-from-s3
@@ -17,9 +17,9 @@ main() {
     if [[ $src_hash != "$dst_hash" ]]; then
         aws s3 cp --no-progress "$src" - |
         if [[ "$src" == *.gz ]]; then
-            gunzip -cfq
+            pigz -cdq
         elif  [[ "$src" == *.xz ]]; then
-            xz -T0 -dcq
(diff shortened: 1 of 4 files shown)


j23414:
Thank you both for the compression exploration! Should we close this since we've already moved and merged`zstd`?

tsibley:
Agreed re: closing this. We do still produce xz-compressed and gzip-compressed files paralleling the new zstd-compressed files, right?, so there may still be gains to be had with parallel compression/decompression of xz and gzip. But it's not a big deal to revisit this down the road if we need to.

PR #262, opened 2021-12-17
chore: cleanup logs

ivan-aksamentov:
This attempts to improve the log readability:

 - Removes error message in case where fetching S3 object metadata cannot find an object. This not an error for us, as objects that never uploaded is a valid case. This change is a bit uncertain, because there might be other errors, not just "object not found", but this change hides all. Maybe there is a better way?
 - Removes printing of hashes of uploaded files. This is probably leftovers from debugging.
 - Prints filenames in the "skip uploading" branch, for more info and symmetry with the "do uploading" branch.

### DIFF
--- a/bin/upload-to-s3
+++ b/bin/upload-to-s3
@@ -25,7 +25,7 @@ main() {
 
     local src_hash dst_hash no_hash=0000000000000000000000000000000000000000000000000000000000000000
     src_hash="$("$bin/sha256sum" < "$src")"
-    dst_hash="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.sha256sum --output text || echo "$no_hash")"
+    dst_hash="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.sha256sum --output text 2>/dev/null || echo "$no_hash")"
 
     echo "$src_hash $src"
     echo "$dst_hash $dst"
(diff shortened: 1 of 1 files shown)


huddlej (commented):
> This change is a bit uncertain, because there might be other errors, not just "object not found", but this change hides all. Maybe there is a better way?

I'm not very familiar with the AWS CLI, but the docs for `head-object` suggest that the possible errors are limited to a "404 Not Found" or "403 Forbidden" and that there's no way to get more details about the exception. As a compromise, could we ignore the 404 but print the 403 if it happens? Maybe this doesn't matter much though, since we next try to upload to the same bucket and if there is a 403 error, we'll see it again at upload.

Also, is this error message something we'd see in the logs through regular inspection or more likely only when something fails and we want to figure out why it did?

ivan-aksamentov:
> As a compromise, could we ignore the 404 but print the 403 if it happens? Maybe this doesn't matter much though, since we next try to upload to the same bucket and if there is a 403 error, we'll see it again at upload.

Yes, I don't think it will be very useful and don't have an immediate idea of how to do it without parsing the output. Which is not great.

> Also, is this error message something we'd see in the logs through regular inspection or more likely only when something fails and we want to figure out why it did?

This happens every time there isn't a file on S3. Which happens when you develop on a new branch. There will be no files in the directory dedicated for this branch. But this is fine, it just means the hashes will be all-zero and the "do upload" branch is taken. But you see an error.

If there's a true error and a file is missing in production there is still and error, but it's ignored and then a file is uploaded anyways. So not helping in any way there.

PR #263, opened 2021-12-20
Updates to accessions.tsv (157727 new, 28 removed) seqs downloaded 2021-12-19

AngieHinrichs:
Thanks & Happy holidays!  :)

BTW I got a large file size warning on this when I pushed, dunno what the implications are:
```
remote: warning: File source-data/accessions.tsv is 61.38 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
```


emmahodcroft:
Thank you Angie! Hm, I'm not sure of the implications of this either.... I've passed this onto the rest of the Nextstrain team, and hopefully we can solve. We could split the accessions into two files perhaps, that might be an easy solution. But we can see what others weigh in!

emmahodcroft:
PS - Happy holidays as well! I hope despite the rising tide of Omicron, you're able to have a warm and happy time!

joverlee521:
GitHub has a [file size limit of 100MB](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#file-size-limits) and will block pushes past that limit. We can split the file like @emmahodcroft suggested or use [Git LFS](https://git-lfs.github.com/) to store the file.

tsibley:
LFS works, but requires anyone cloning the repo to have LFS setup in order to get the files. So it's another bit of overhead. Rather than LFS or splitting into multiple files, I would probably just store the file compressed. This will Just Work, and a diff `textconv` filter can be added to make `git diff` and co display the plain text.

emmahodcroft:
If we're sure that our scripts can read the file in compressed (can they?) and that our other scripts will work with a compressed file, do we want to move to this? Angie says it's no problem on her end, but we should ensure everything else keeps working. Alternatively reading in 2 files probably isn't a big deal either, but might take a little more coding.

tsibley:
> If we're sure that our scripts can read the file in compressed (can they?) and that our other scripts will work with a compressed file, do we want to move to this?

Currently _source-data/accessions.tsv_ is used in this repo (ncov-ingest) in two places:

1. `bin/transform-genbank`
2. `bin/transform-gisaid`

Neither supports reading from a compressed file at the moment, but it could ~easily.

Do we know of anything outside of this repo that's using these files? It will be a breaking change to compress them.

[A GitHub search](https://github.com/search?q=org%3Anextstrain+accessions.tsv&type=code) within Nextstrain repos only finds the two instances above, but those search results are not always comprehensive.

tsibley:
[A broader GitHub search](https://github.com/search?q=%22accessions.tsv%22+%22ncov-ingest%22&type=code) implicates @AngieHinrichs's tooling and something by @j23414. Any other consumers are likely "dark" and not publicly available.

j23414:
Mine is only documentation. Changes won't break my stuff. :)

emmahodcroft:
I imagine we and Angie are the main people who use this, but we can cross-post in SPHERES and ubioinfo to alert those who we can. I think the number of people who are willing/able to put the compute power into ingest is not a huge number - though with our changes to do incremental alignment, that might change.

emmahodcroft:
The Broad might be the other people.

emmahodcroft:
Maybe CZI?

PR #266, opened 2022-01-04
Remove GenBank metadata diff

joverlee521:
The `csv-diff` within `notify-on-metadata-change` began running into
a MemoryError on December 26, 2021. This was patched during the holidays
by increasing the memory for Genbank ingest runs (bd2c2ca).

Further investigation showed that we have not been receiving the
metadata diff output files since we switched to Snakemake on
December 1, 2021 (74de8b1). The `upload` rule has been running before
`notify_genbank` resulting in diffing of the same file that has been
uploaded to S3. This is caused by `upload` not having any dependencies
on `notify_genbank` and `notify_genbank` having an additional dependency
on `check_locations`. This prompted an [internal discussion](https://bedfordlab.slack.com/archives/C0159227X7Y/p1641237304011100) that showed
no one has been using the outputs of this metadata diff.

### DIFF
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -48,14 +48,14 @@ jobs:
           --no-download \
           --image nextstrain/ncov-ingest \
           --cpus 16 \
-          --memory 60GiB \
+          --memory 30GiB \
           --exec env \
           . \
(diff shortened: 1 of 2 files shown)


joverlee521:
Tested locally that `notify_genbank` works as expected after the changes. 
Cancelled the branch runs since they are not affected by the change.

PR #275, opened 2022-01-18
Updates to accessions.tsv (305831 new, 63 removed) seqs downloaded 2022-01-17

AngieHinrichs:
Belated Happy New Year!  :)

There was some discussion in #263 about how to deal with the very large file size, but I don't know if there was a final decision.  I could easily compress the file next time.  gzip reduces the size from 67MB to 13MB.  xz reduces it further to 5MB.


emmahodcroft:
Thanks Angie! Happy New Year to you as well, and thank you for your continued work on this into another glorious year 😁 !  
I bumped the issue of the file size again in the previous thread - hopefully we can come to a decision!

(Apologies, I'm pretty much out of action this month due to grant writing!)

PR #276, opened 2022-01-21
harmonize metadata columns and add breakdown of private mutations

rneher:
@j23414   -- this PR has the simple fixes we discussed. And it adds 3 columns to the metadata that will start to exist next Tue. So we should aim to merge this Tue night or Wed.


j23414:
Thanks, @rneher! Changes looks good to me, and I'll set a calendar reminder to merge on Tues/Weds

PR #277, opened 2022-01-23
chore: delay GISAID download to 16:07 CET

corneliusroemer:
Adjusts the time of the auto-ingest of GISAID from 3pm CET to 4pm CET in order to be more in line with when GISAID seems to be updating their files.

One remaining question - there is no reason to move the Genbank ingest as we pull in realtime, but do we want to keep the ingest times in sync just for neatness? (Currently I modify only GISAID)

### DIFF
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -4,7 +4,7 @@ on:
   schedule:
     # Note times are in UTC, which is 1 or 2 hours behind CET depending on daylight savings.
     #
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Checking recent times it looks like in the past 2 weeks GISAID has mostly been updating later in the afternoon (often around 3-3.20 but that's still missing our current build), so this is probably worth pushing back. Not sure why this seems to be creeping back.

emmahodcroft (approved):

corneliusroemer:
Well the dataset grows and GISAID presumably starts at the same time so it finishes later and later, I would say.

PR #282, opened 2022-02-05
Updates to accessions.tsv (520330 new, 61 removed, 1250 changed) seqs downloaded 2022-02-04

AngieHinrichs:
Half a million new mappings... crazy.  I think NCBI downloads were not updating for almost a week before the last update, and then they go unstuck and whew!   gzipped accessions.tsv.gz just in time!  😃 

Thanks & have a great weekend.


emmahodcroft:
Thank you Angie, as always! Glad the new file format prevented big breakage with the unsticking! 😆

PR #284, opened 2022-02-14
add: annotation updates from 2022 Feb 9 to 17

j23414:

Update geo annotations from slack message: https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1644347002653539

```
Writing updated annotation files to scripts/curate_metadata/output_curate_metadata/...
Attention: gisaid_annotations.tsv was altered! Remember to replace the old file in ../ncov-ingest/source-data/.
No changes to genbank_annotations.tsv.
```

### Testing

Let me know if I missed something. I pasted the final message from our scripts above. I'll work through the other files.

### DIFF
--- a/source-data/genbank_annotations.tsv
+++ b/source-data/genbank_annotations.tsv
@@ -2035,6 +2035,24 @@ OM506401	division	New York # previously Bronx County
 OM506401	location	Bronx County # previously
 OM506404	division	New York # previously Queens County
 OM506404	location	Queens County # previously
+OM640066	division	Upper Austria # previously Upper Austria / Voecklabruck / Frankenburg Am Hausruck
+OM640066	location	Frankenburg Am Hausruck # previously 
+OM640067	division	Upper Austria # previously Upper Austria / Voecklabruck / Voecklabruck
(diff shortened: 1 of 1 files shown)


MoiraZuber:
I think you might have used the wrong script for this purpose? For creating travel-associated annotations, the necessary script would be `parse_additional_info.py`. When running `curate_metadata.py`, the `additional-info-changes.txt` is not used as input and is not parsed. The only thing the script did in this instance was to check for a few regularly occurring issues (e.g. standardize host to "Mink") and sort the annotations file.

MoiraZuber:
The two `annotations` files you pushed look good to me! They correctly contain annotations produced from additional info as well as from curation. The only thing missing for a complete curation would be the new rules in `source-data/gisaid_geoLocationRules.tsv` (as well as the new `lat_longs` and `color_ordering` files, but they're in a different repo, so that's difficult to emulate here...). In case the sorting script doesn't work for you, it's totally fine to just add the rules at the bottom of the file. As long as I clean the file every now and then, no big issues should arise.

PR #290, opened 2022-03-06
Updates to accessions.tsv.gz (211466 new, 2 removed) seqs downloaded 2022-03-06

AngieHinrichs:
A very quiet week for name changes, withdrawals etc.  Gotta appreciate the little things!  Take care and thanks for all you do.


emmahodcroft:
Thank you Angie and thank you for the continued updates on this - it's very much appreciated! We are celebrating Fasnacht this week in Basel - so in that spirit - Schöni Fasnacht! 🎺 🥁 🤡

AngieHinrichs:
Thanks!  :)  I had to look that up, and wow, that sounds pretty awesome.  (Gotta say though, a party that starts at 4am sharp... what could be more Swiss?  😄)

emmahodcroft:
It's so incredibly Swiss! Even though at first glance it can seem quite un-Swiss - it's ok to make noise all night, the whole city centre shuts down as bands march around and people throw confetti - you can even throw rubbish straight on the ground! It's all organized thoroughly and works amazingly smoothly. For example - you are only allowed to throw single-color confetti (there are many colors, but one bag is only one color) so that you can tell if people have picked it up off the ground (then it's mixed colours)!

At the end of the day there are inches of confetti on the ground, but by 7am the next morning, the streets are spotless, thanks to the early-morning cleaning teams that scour the city. In the UK if a town spent the whole day drinking and partying, there would be brawls and probably fires by evening, but in Basel I've never seen anything of the sort :)

PR #291, opened 2022-03-15
feat: add Nextclade_pango column to metadata

ivan-aksamentov:
Copies the [new](https://github.com/nextstrain/nextclade/releases/tag/1.11.0) `Nextclade_pango` column from nextclade.tsv to metadata.tsv. This column contains Pango lineages assigned using Nextclade.

Symmetrical PR in ncov: https://github.com/nextstrain/ncov/pull/892

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -15,6 +15,7 @@ reference_day = datetime(2020,1,1).toordinal()
 
 column_map = {
     "clade": "Nextstrain_clade",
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):

ivan-aksamentov:
@corneliusroemer Yes, I think so. The usual ingest full run checklist need to be executed first and then the daily ingest will be able to produce nextclade.tsv and metadata.tsv with the new cloumn.

PR #293, opened 2022-03-16
Mask bad collection date

joverlee521:
Masks the collection date with `XXXX-XX-XX` if the collection date is
the same day as or after the submission date.

This was suggested by @trvrb when @corneliusroemer noticed that a subset
of GISAID data had collection dates identical to the submitting date.
It is most likely impossible to sample, PCR, sequence, and submit on
same day. However, if we do determine this is possible, the masking
can be overridden for specific cases via the manual annotations.

### DIFF
--- a/bin/transform-genbank
+++ b/bin/transform-genbank
@@ -25,6 +25,7 @@ from utils.transformpipeline.transforms import (
     DropSequenceData,
     FillDefaultLocationData,
     FixLabs,
+    MaskBadCollectionDate,
     MergeBiosampleMetadata,
     MergeUserAnnotatedMetadata,
     ParseGeographicColumnsGenbank,
@@ -169,6 +170,7 @@ if __name__ == '__main__':
                               | FixLabs()
                               | ParsePatientAge()
                               | ParseSex()
+                              | MaskBadCollectionDate()
                               | StandardizeGenbankStrainNames()
                               | ParseGeographicColumnsGenbank( base / 'source-data/us-state-codes.tsv' )
                               | AbbreviateAuthors()
--- a/bin/transform-gisaid
+++ b/bin/transform-gisaid
@@ -24,6 +24,7 @@ from utils.transformpipeline.transforms import (
     FillDefaultLocationData,
     FixLabs,
     MergeUserAnnotatedMetadata,
+    MaskBadCollectionDate,
     ParsePatientAge,
     ParseSex,
     RenameAndAddColumns,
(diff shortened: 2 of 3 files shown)


corneliusroemer (approved):
LGTM
How do we test this, just on the live ingest?
Easiest to confirm will be to filter for `submitting_lab == 'WSSE w Warszawie'` and see if there are a number of `XXXX-XX-XX` there
Also want to filter for `XXXX-XX-XX` dates and see how many there are now.

GISAID failed again with `OSError: [Errno 28] No space left on device` hmmm
https://github.com/nextstrain/ncov-ingest/runs/5576442356?check_suite_focus=true#step:4:692 https://github.com/nextstrain/ncov-ingest/runs/5576442356?check_suite_focus=true#step:4:725

tsibley:
> How do we test this, just on the live ingest?

Diff the metadata files produced by the PR checks with their contemporaries produced by `master`, and make sure that the changes reported align with expectations.

corneliusroemer:
> Diff the metadata files produced by the PR checks with their contemporaries produced by master, and make sure that the changes reported align with expectations.

Where do we find the PR run output files?

@tsibley shall we restart the GISAID run? Or wait so as not to overburden the harddrive while full run is running?

tsibley:
Restart should be fine. These jobs will run on separate nodes since they're large enough and request CPU/mem that corresponds to a whole node (of which no larger size is a multiple).

Output from this PR runs ends up in two places:

- GISAID → s3://nextstrain-ncov-private/branch/mask-collection-date/
- GenBank → s3://nextstrain-staging/files/ncov/open/branch/mask-collection-date/

joverlee521:
For both GISAID and GenBank, I compared the PR runs with the closest runs on master. I converted the metadata TSVs to SQLite and then used `daff` to diff them. I ignored differences of entire entries being added or removed within GenBank since these are caused by upstream data changes. 

| | GISAID | GenBank|
|-|--------|----------|
|master S3 version-id|`9ec3PjZq0S00c8I_4KhvLAu8QG0hzU8O`| `F5MJFKJG9vmhCg8.H72pW3igMHJTcXwR` |
|total updated entries| 3805 | 5175 |
|date -> XXXX-XX-XX| 1180 | 30 |
|clock_deviation - 1.0 | 2625 | 5000 |
|pango_lineage updated | 0 | 145 | 

I confirmed that the updated dates had the same date as the submission date, so these updates are working as intended. 

The `pango_lineage` updates in GenBank came from upstream data changes, so these can be ignored for this PR. 
The `clock_deviation` decreasing by 1.0 was unexpected for me. It is [calculated during the joining of metadata with NextClade](https://github.com/nextstrain/ncov-ingest/blob/master/bin/join-metadata-and-clades#L114) so maybe it can be ignored for this PR as well?

PR #299, opened 2022-04-19
Trigger ncov counts

joverlee521:
Add trigger for nexstrain/counts clade counts to run after the metadata TSV has been uploaded to S3.

Cannot be fully tested until https://github.com/nextstrain/counts/pull/2 has been merged since repository dispatch events only run on the default branch.

### DIFF
--- a/bin/trigger
+++ b/bin/trigger
@@ -1,13 +1,13 @@
 #!/bin/bash
 set -euo pipefail
 
-TOKEN="${PAT_GITHUB_DISPATCH:-}"
+: "${PAT_GITHUB_DISPATCH:=}"
 
 repo="${1:?A repository name is required as the first argument.}"
 event_type="${2:?An event type is required as the second argument.}"
 shift 2
 
-if [[ $# -eq 0 && -z $TOKEN ]]; then
+if [[ $# -eq 0 && -z $PAT_GITHUB_DISPATCH ]]; then
     cat >&2 <<.
 You must specify options to curl for your GitHub credentials.  For example, you
 can specify your GitHub username, and will be prompted for your password:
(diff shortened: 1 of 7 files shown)


bin/trigger
```diff
@@ -1,18 +1,23 @@
 #!/bin/bash
 set -euo pipefail
 
+TOKEN="${PAT_GITHUB_DISPATCH:-}"
```
tsibley, line 4: You can do this to preserve the original env var name but ensure it's a declared var for `set -u`.

```suggestion
: "${PAT_GITHUB_DISPATCH:=}"
```

Using the original env var name in the code rather than a new name makes it more clear where the env var is used when grepping.

bin/trigger
```diff
@@ -25,14 +30,21 @@ and then tell curl to use it:
...
 ¹ https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line
-² https://ec.haxx.se/usingcurl/usingcurl-netrc
+² https://docs.github.com/en/rest/overview/other-authentication-methods#via-username-and-password
+³ https://ec.haxx.se/usingcurl/usingcurl-netrc
 .
     exit 1
 fi
 
+auth=':'
+if [[ $TOKEN ]]; then
```
tsibley, line 41: Should be `-n "$TOKEN"` I think.

bin/trigger
joverlee521: Thanks for reviewing! Updated in [46962f8](https://github.com/nextstrain/ncov-ingest/pull/299/commits/46962f8d0f37d0b4e93ab03811dfd61bf3056e7c)

bin/trigger
joverlee521: Yup! Updated in [46962f8](https://github.com/nextstrain/ncov-ingest/pull/299/commits/46962f8d0f37d0b4e93ab03811dfd61bf3056e7c)

joverlee521:
Tested locally with:
```
touch data/genbank/upload.done
envdir ../env.d/github_pat/ snakemake --cores all --configfile config/genbank.yaml --config trigger_rebuild=True
```
and
```
touch data/genbank/upload.done
envdir ../env.d/github_pat/ snakemake --cores all --configfile config/genbank.yaml --config trigger_counts=True
```
Both the [genbank rebuild](https://github.com/nextstrain/ncov/actions/runs/2272838237) and the [genbank clade counts](https://github.com/nextstrain/counts/actions/runs/2272841160) workflows were successfully triggered and then manually cancelled. Trusting the minimal additions the workflow YAMLs will work fine.

PR #302, opened 2022-04-27
Failure notifications

joverlee521:
Add Slack failure notifications for both errors within the Snakemake pipeline and the GitHub Action workflows. 
See commits for details.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -497,8 +497,9 @@ onsuccess:
         shell("./bin/clean")
 
 onerror:
-    message = "❌ This pipeline has FAILED 😞. Please see linked thread for more information."
-    print(message)
+    print("Pipeline failed.")
+    if send_notifications:
+        shell("./bin/notify-on-job-fail")
     if not config.get("keep_all_files", False):
         print("Removing intermediate files (set config option keep_all_files to skip this)")
         shell("./bin/clean")
--- a/bin/notify-on-job-fail
+++ b/bin/notify-on-job-fail
@@ -10,7 +10,7 @@ aws_batch_job_id="${AWS_BATCH_JOB_ID:-}"
 github_run_id="${GITHUB_RUN_ID:-}"
 
 echo "Notifying Slack about failed ingest job."
-message=":x: Ingest job failed. "
+message="❌ Ingest job has FAILED 😞 "
 
 if [ -n "${aws_batch_job_id}" ]; then
     message+="See AWS Batch job \`${aws_batch_job_id}\` (<https://console.aws.amazon.com/batch/v2/home?region=us-east-1#jobs/detail/${aws_batch_job_id}|link>) for error details. "
--- a/bin/notify-on-job-fail
+++ b/bin/notify-on-job-fail
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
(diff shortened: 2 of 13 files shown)


bin/notify-on-job-fail
```diff
@@ -0,0 +1,21 @@
...
+: "${SLACK_TOKEN:?The SLACK_TOKEN environment variable is required.}"
+: "${SLACK_CHANNELS:?The SLACK_CHANNELS environment variable is required.}"
+
+bin="$(dirname "$0")"
+
+aws_batch_job_id="${AWS_BATCH_JOB_ID:-}"
+github_run_id="${GITHUB_RUN_ID:-}"
+
+echo "Notifying Slack about failed ingest job."
+message=":x: Ingest job failed. "
```
jameshadfield, line 13: Have the emojis gone because they can't be passed through as arguments? I find them useful when scanning the (often numerous) messages to the slack channel. 

Formerly the text was "❌ This pipeline has FAILED 😞" (I realise this wasn't sent to slack, but the phylogenetics pipeline does send an analogous message to slack).

Snakefile
```diff
@@ -497,8 +497,9 @@ onsuccess:
         shell("./bin/clean")
 
 onerror:
```
jameshadfield, line 499: Noticed while reviewiing this PR, and perhaps out-of-scope. In the `onsuccess` block immediately above this, the message will not be sent to slack, correct? Could we add a `"$bin"/notify-slack "{message}"` shell block there? (It would be better if it were threaded, but I remember this not being straightforward with the ncov-ingest set-up.)

jameshadfield (commented):
Looks good from reading through the code Jover, thanks! 

I do think threading messages on slack is the way forward to avoid the numerous messages posted to the channel daily (assuming we want to keep using slack for this).

bin/notify-on-job-fail
emmahodcroft: Agree, the emoji are useful for spotting 'bad' messages!

joverlee521: Ah, the `:x:` will show up in the Slack message as ❌
I can keep the original message if it is clearer to have the emojis directly in the script.

Snakefile
joverlee521: Hmm, I thought an `onsuccess` message would just be extra noise. There's already slack notifications for the file updates, which indicate the pipeline successfully completed. The successful pipeline will also trigger the ncov builds, which will send the "Pipeline starting..." messages.

bin/notify-on-job-fail
joverlee521: Added the emoji's back in [dcce209](https://github.com/nextstrain/ncov-ingest/pull/302/commits/dcce209cb321c18592d63068e531ffd57586f05f)

joverlee521:
> I do think threading messages on slack is the way forward to avoid the numerous messages posted to the channel daily (assuming we want to keep using slack for this).

Yeah, I will probably save threading messages for another PR 😅 
I'll merge this on Monday so I'm around if anything goes wrong

Snakefile
jameshadfield: Good points Jover, I agree!

bin/notify-on-job-fail
jameshadfield: > Ah, the :x: will show up in the Slack message as ❌

🤦  my mistake!

PR #310, opened 2022-05-19
notify-on-record-change: notify on all changes

joverlee521:
We've been having stale fetches from the GISAID endpoint that lack new
records. It would be helpful to have specific notifications about the
different types of record changes, not just new records. This way we get
early notifications about the status of our fetches instead of having
to look for messages buried in the pipeline logs.


victorlin (approved):

emmahodcroft:
Thanks @joverlee521 - this is really useful!!

PR #313, opened 2022-06-04
Store record count in S3 metadata

### LINKED ISSUE #243: Store line count in object metadata to avoid network transfer just to count lines
`notify-on-record-change` compares counts lines in a local file and remote file. It counts lines in the remote file by downloading and decompressing it in its entirety.

https://github.com/nextstrain/ncov-ingest/blob/94081fdcc0d54f01d4a10ef7a53197b4f6296dd8/bin/notify-on-record-change#L16-L18

This presumably takes some non-trivial amount of network transfer time! Instead of doing that unnecessary work, we could store line counts in the object metadata, like we already do for `sha256sum`, and then very quickly be able to get a remote file line count.

rajivshah3:
This PR adds the record count of a file to the metadata in S3, similar to what has been done for the SHA-256 hash. If `notify-on-record-change` is run on a file where the destination has `recordcount` in its metadata, the lines only need to be counted on the source file. This avoids the need to download the destination file, uncompress it, and count the lines locally before determining whether records have been added/removed

### Related issue(s)
Fixes #243 

### Testing
I've tested some of these commands manually and they seem to work, but I haven't verified whether it works in the context of the rest of the pipeline.

### DIFF
--- a/bin/notify-on-record-change
+++ b/bin/notify-on-record-change
@@ -20,7 +20,7 @@ key="${s3path#*/}"
 src_record_count="$(wc -l < "$src")"
 
 # Try getting record count from S3 object metadata
-dst_record_count="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.recordCount --output text 2>/dev/null)"
+dst_record_count="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.recordcount --output text 2>/dev/null)"
 if [[ $dst_record_count == "None" ]]; then
   # This object doesn't have the record count stored as metadata
   # We have to download it and count the lines locally
--- a/bin/notify-on-record-change
+++ b/bin/notify-on-record-change
@@ -20,7 +20,7 @@ key="${s3path#*/}"
 src_record_count="$(wc -l < "$src")"
 
 # Try getting record count from S3 object metadata
-dst_record_count="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.recordcount --output text 2>/dev/null)"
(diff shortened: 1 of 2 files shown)


bin/notify-on-record-change
```diff
@@ -13,8 +13,20 @@ source_name=${3:?A record source name is required as the third argument.}
...
 
+s3path="${dst#s3://}"
+bucket="${s3path%%/*}"
+key="${s3path#*/}"
+
 src_record_count="$(wc -l < "$src")"
-dst_record_count="$(wc -l < <(aws s3 cp --no-progress "$dst" - | xz -T0 -dcfq))"
+
+# Try getting record count from S3 object metadata
+dst_record_count="$(aws s3api head-object --bucket "$bucket" --key "$key" --query Metadata.recordcount --output text 2>/dev/null | tr -d '"')"
```
tsibley, line 23: With `--output text` it shouldn't be necessary to remove quotes with `tr` (unless extra quotes were added at some point by bad quoting, but quoting during upload looks fine now).

bin/notify-on-record-change
```diff
@@ -13,8 +13,20 @@ source_name=${3:?A record source name is required as the third argument.}
...
+if [[ $dst_record_count == "None" ]]; then
```
tsibley, line 24: Instead of relying on this conversion by `aws` of a JSON `null` to "text", it's possible to extend the `--query` expression to include `|| ''` to explicitly return an empty string and then test for that with `-z` here.

tsibley (commented):
@rajivshah3 Thanks for this PR! I've slightly tweaked it (preserving your authorship) and pushed that as #316. I'll launch a test run for it to make sure things are working in the full pipeline context, but I expect it'll be good. It'll be nice to have this speedup!

bin/notify-on-record-change
rajivshah3: Ooh, good point

PR #320, opened 2022-06-23
Modularize Snakemake rules

joverlee521:
Modularizes Snakemake rules so that any rules involving AWS S3, Slack notifications, and GitHub Action triggers are optional. The main motivation for these changes is to be able to port our ncov-ingest pipeline to Terra for external users.

See commits for details.

### Testing
Testing the main pipeline without bells and whistle on Terra:
- [GISAID](https://app.terra.bio/#workspaces/Nextstrain/Development/job_history/624ba3c2-14c6-4c1b-8666-95cb43cf93ed)
- [GISAID w/ workspace NextClade cache](https://app.terra.bio/#workspaces/Nextstrain/Development/job_history/d1f04b7a-ea5e-4591-bcf1-fc908db875b0)

Testing our internal runs via GH Actions on AWS Batch (although the branch runs do _not_ include Slack notifications or GH Action triggers):
- [GenBank](https://github.com/nextstrain/ncov-ingest/actions/runs/2552550268) ([AWS Batch Job](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/1b2827ba-bee7-49cd-b6f1-bf12aaa652da) completed successfully.)


ivan-aksamentov:
It's difficult for me to do a meaningful review, because I am not a huge specialist in Snakemake, but I felt like I am responsible for the Nextclade parts, so I glanced over and it all looks alright. Great job!

A few notes:

 - It's probably not related to the PR (i.e. it was not introduced here), but one thing I noticed is that `threads` in Nextclade's  is hardcoded to `64`, while daily ingest runs on `16` it seems:
https://github.com/nextstrain/ncov-ingest/blob/f27c9c8de125906c3ce7d5f62067cb92790e2f8b/.github/workflows/fetch-and-ingest-gisaid-master.yml#L68

    Some oversubscription can make it a bit faster under certain conditions. But 4x oversubscription likely don't.

    We might need to align these two numbers somehow. Perhaps in a separate PR.



 - Running with the entirely empty cache would require substantial resources - aligning the whole GISAID is no joke (and costs accordingly). That's why Nextclade full runs exist. If this is intended for third-party users, they better know about that. Otherwise, if they happen to make their first run on 16 threads and empty cache, it may take a while and they might not be able to understand what's happening. Perhaps we could add a note in the readme and/or in the inline comments/docs.

 - Sometimes later we could also phase-out the `run-nextclade` script, because it does not do much, and we could call Nextclade binary directly.

joverlee521:
Thanks for the review @ivan-aksamentov 🙏 

> Perhaps we could add a note in the readme and/or in the inline comments/docs.

I [added a warning note](https://github.com/nextstrain/ncov-ingest/pull/320/commits/d12cd96962785857d1c00d8faadc663023d375d7) to the README for empty cache runs and made separate issues for the other notes.

Snakefile
```diff
@@ -34,370 +34,17 @@ rule all:
...
-        files_to_upload["biosample.tsv.gz"] =           f"data/{database}/biosample.tsv"
-        files_to_upload["duplicate_biosample.txt.gz"] = f"data/{database}/duplicate_biosample.txt"
-    elif database=="gisaid":
-        files_to_upload["additional_info.tsv.gz"] =     f"data/{database}/additional_info.tsv"
-        files_to_upload["flagged_metadata.txt.gz"] =    f"data/{database}/flagged_metadata.txt"
-
-    nextclade_sequences_path = checkpoints.get_sequences_without_nextclade_annotations.get().output.fasta
-    if os.path.getsize(nextclade_sequences_path) > 0:
-        files_to_upload["nextclade.tsv.gz"] =           f"data/{database}/nextclade.tsv"
-        files_to_upload["aligned.fasta.xz"] =           f"data/{database}/aligned.fasta"
```
tsibley, line 357: _non-blocking, minor_

I noticed that git's moved line detection flagged these two lines as not being moved, i.e. the original lines were deleted here but different lines added over there instead of the same lines added. When I looked into why, it was just an internal whitespace change, presumably accidental. Here's [a fixup patch](https://github.com/nextstrain/ncov-ingest/files/9203539/0001-fixup-Split-Snakefile-into-workflow-modules.txt) for that, which you can apply if you want with:

```
curl -fsSL https://github.com/nextstrain/ncov-ingest/files/9203539/0001-fixup-Split-Snakefile-into-workflow-modules.txt \
  | git am
```

and then actually fixup the original commit with `git rebase -i`.

workflow/snakemake_rules/fetch_sequences.smk
```diff
@@ -29,48 +30,72 @@ def run_shell_command_n_times(cmd, msg, cleanup_failed_cmd, retry_num=5):
...
+
+    # Set ruleorder since these fetch rules have the same output
+    # Fetch directly from databases when `fetch_from_database=True`
+    # or else fetch files from AWS S3 buckets
+    if config.get("fetch_from_database", False):
+        ruleorder: fetch_main_ndjson > fetch_main_ndjson_from_s3
+        ruleorder: fetch_biosample > fetch_biosample_from_s3
+    else:
+        ruleorder: fetch_main_ndjson_from_s3 > fetch_main_ndjson
+        ruleorder:  fetch_biosample_from_s3 > fetch_biosample
```
tsibley, line 73: _non-blocking_

An alternative to defining two sets of rules and then using `ruleorder` to disambiguate them is to conditionally define the rules, e.g.:

```python
if config.get("fetch_from_database", False):
    rule fetch_main_ndjson:
        # fetch from upstream database
    rule fetch_biosample:
        # fetch from upstream database

elif config.get("s3_dst") and config.get("s3_src"):
    rule fetch_main_ndjson:
        # fetch from S3
    rule fetch_biosample:
        # fetch from S3
```

This might make things easier? It does mean you can't, for example, directly run `rule fetch_main_ndjson_from_s3` even when `fetch_from_database` is true, but I don't think we need to do that.

This also brings to mind that the conditional used by the replaced rules:

```
run:
    if config.get("fetch_from_database", False):
        # fetch from upstream database
    else:
        # fetch from S3
```

could have alternatively to this change been modified from `else` → `elif config.get("s3_dst") and config.get("s3_src")`.

Is there a reason the separate rules are preferable? They seem cleaner, but not sure about other reasons.

workflow/snakemake_rules/nextclade.smk
```diff
@@ -162,18 +162,26 @@ rule combine_alignments:
...
     nextclade_sequences_path = checkpoints.get_sequences_without_nextclade_annotations.get().output.fasta
     if os.path.getsize(nextclade_sequences_path) > 0:
-        return f"data/{database}/nextclade.tsv"
-    return f"data/{database}/nextclade_old.tsv"
+        # Return all three output files here so that they get pulled through
+        # the Snakemake DAG even if we are not uploading them to S3.
+        return [
+            f"data/{database}/nextclade.tsv",
+            f"data/{database}/aligned.fasta",
+        ]
```
tsibley, line 175: `s/three/two/`?

workflow/snakemake_rules/nextclade.smk
```diff
@@ -162,18 +162,26 @@ rule combine_alignments:
...
+    return [
+        f"data/{database}/nextclade_old.tsv",
+        f"data/{database}/nextclade.aligned.old.fasta",
+    ]
 
 rule generate_metadata:
     input:
         existing_metadata = f"data/{database}/metadata_transformed.tsv",
-        new_metadata = _get_nextclade_info
+        nextclade_output = _get_nextclade_output
```
tsibley, line 184: If `_get_nextclade_output` returned a dictionary instead of a list, you could do:

```
input:
    unpack(_get_nextclade_output),
    existing_metadata = f"data/{database}/metadata_transformed.tsv"
```

to make the inputs individually keyed. See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#input-functions-and-unpack.

This would help with the shell block for this rule… (see below)

workflow/snakemake_rules/nextclade.smk
```diff
@@ -182,6 +190,6 @@ rule generate_metadata:
         """
         ./bin/join-metadata-and-clades \
             {input.existing_metadata} \
-            {input.new_metadata} \
+            {input.nextclade_output[0]} \
```
tsibley, line 193: Kinda awkward and unclear why this accesses the first item in the list, i.e. you have to go tracing what that item is.

Would be nicer to be able to reference `input.new_metadata` instead, which should be possible with the suggestion above.

tsibley (changes requested):
Nice work! This is a good change to be making. I haven't tried running the updated workflow, but I assume you have and it works. I read thru all the commits and offered a few comments, but overall this looks pretty close to merging to me.

Snakefile
joverlee521: Ah, thank you for the patch!

workflow/snakemake_rules/nextclade.smk
joverlee521: Ah yup, just recently learned by about the `unpack` keyword. I'll update this.

workflow/snakemake_rules/fetch_sequences.smk
joverlee521: I can see how conditionally defined rules is cleaner. I wanted to make the fetch from upstream databases the default behavior even when nothing is specified in the config, which is why I went with the current approach.

fanninpm:
> * Sometimes later we could also phase-out the `run-nextclade` script, because it does not do much, and we could call Nextclade binary directly. Although a separate step of downloading the dataset will be necessary.

https://github.com/nextstrain/nextclade/pull/939 may obviate the need for that separate step.

workflow/snakemake_rules/fetch_sequences.smk
tsibley: That makes sense.

workflow/snakemake_rules/nextclade.smk
joverlee521: Fixed up in 30f0e6a6fd94c24996377a7434671605dfb0a2e2. I renamed `input.new_metadata` to `input.nextclade_tsv` to make it clearer that we are merging metadata with NextClade data.

workflow/snakemake_rules/nextclade.smk
joverlee521: (See [comment](https://github.com/nextstrain/ncov-ingest/pull/320#discussion_r933554543) above)

joverlee521:
Doing final test runs before merging:

- [ ] GISAID ([GH Action](https://github.com/nextstrain/ncov-ingest/actions/runs/2929206449) / [AWS Batch](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/37a8d57a-838e-48d2-b34d-4c93a57001c6)) 
- [ ] GenBank ([GH Action](https://github.com/nextstrain/ncov-ingest/actions/runs/2929207763) / [AWS Batch](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/cf98359c-43bc-4b45-a652-e41bc68d3a7d))

Once test runs complete without errors, I'll merge this next Monday so I can monitor the pipeline throughout the week.

PR #321, opened 2022-06-27
add: annotation updates up to 2022-06-27

j23414:
### Description of proposed changes:
Update annotations up to June 27th. Let me know if I missed anything.

### Related Issue(s):
Related to https://github.com/nextstrain/ncov/pull/966


trvrb:
Hey @j23414. What's the reason for lines like?
```
# Belgium/AZDelta-2223-10887/2022	EPI_ISL_13393039	additional_location_info: ZIP CODE: 8700
```

We don't use this additional location info and this is commented out anyway.

j23414:
Hmm, @MoiraZuber may be able to answer better. We can try reverting this change to see if it fixes the problem. 
My impression was that the commented and normal sections were supposed to merge.
Agree that we're not using "additional_location_info" so I would expect dropping it should be fine... @MoiraZuber ?

emmahodcroft:
Worth waiting for @MoiraZuber to weigh in, but these may be lines that contained extra, more detailed information, which we ourselves can't use but that we had from some source or another. This way they're available somewhere publicly in such a way that other could access them if they'd be useful for them. At least this was the logic behind some of the info we include - may or may not be the case here!

MoiraZuber:
We started adding these comments to preserve additional information we could not use at that moment in case we'd need them later on. The Zip codes we sometimes used as a  reference when resolving location conflicts. But I think we're now past the point where this is useful (especially since we're not curation location anymore). I'd be in favor of removing those lines.

trvrb:
Great! I agree to just stop including unused `additional_location_info`.

PR #323, opened 2022-06-28
feat: retry reverse complement in nextclade

corneliusroemer:
Use our new neat `retry-reverse-complement` Nextclade feature - who knows, maybe some sequences are actually reverse complemented.

### DIFF
--- a/bin/run-nextclade
+++ b/bin/run-nextclade
@@ -59,5 +59,6 @@ echo "[ INFO] ${0}:${LINENO}: Running Nextclade"
   --genes="${GENES}" \
   --output-translations="${NEXTCLADE_OUTPUTS_DIR}/nextclade.gene.{gene}.fasta" \
   --output-fasta="${OUTPUT_FASTA}" \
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
What's the relation between this and https://github.com/nextstrain/ncov-ingest/pull/324?

This change only affects daily runs. Any sequences salvaged during dailies will be wiped out after the next the full run. Full run flags are here: https://github.com/nextstrain/ncov-ingest/blob/f201352f939323356328661f93b2209b51cb6c88/bin/run-nextclade-full#L121-L129

There is also Nextclade invocation in ncov here:
https://github.com/nextstrain/ncov/blob/9adfff82387357b8c2be27518ffff37442457bbc/workflow/snakemake_rules/main_workflow.smk#L502-L508 It may or may not need to be fixed. But I cannot tell for sure. I think it is running the sequences that are added beyond the aligned fasta from S3. So likely yes.

If the goal is to salvage as many broken sequences as possible, then we might also consider `--replace-unknown`.

PR #324, opened 2022-06-28
feat: allow reverse complemented sequences to be aligned

corneliusroemer:
Allow reverse complemented sequences to be aligned - who knows, maybe there are some.

We should maybe do a full run with this setting on and see what happens before we invest time into adding a `is_reverse_complemented` column that may not be needed

### Tests

- Full run genbank: `fd6edce8-7f2d-413f-84d5-98b414d5a042`


ivan-aksamentov:
I am asked to review this, but don't understand what's the relation between this and https://github.com/nextstrain/ncov-ingest/pull/323. 

Also why alignment path is changed. I think this will break. 

We should not mix formatting and logic changes in the same PR.

corneliusroemer:
Since only 1 sequence in open data seems to be reverse complemented, this PR isn't worth it so I'll close.

PR #349, opened 2022-10-08
Updates to accessions.tsv.gz (173518 new, 873 removed, 168 updated) seqs downloaded 2022-10-07

AngieHinrichs:
Thanks!  🙂


emmahodcroft:
Thank you Angie! 😁 Hope you are having a nice start to autumn! 🍂

AngieHinrichs:
Wow, that was fast!  😄  Autumn is kind of a relative thing in the non-mountainous parts of California, but very nice!

emmahodcroft:
Haha, I guess that's true! Well, I'm sure you can still find fun pumpkins to decorate with! Here in Switzerland the leaves are turning and looking lovely!

PR #350, opened 2022-10-19
dev: Upgrade Python from 3.7 → 3.10

tsibley:
Accounts for the same change being made to nextstrain/base¹, soon to be released.

Upgrades Python deps too since that it was necessary to let Pipenv re-resolve and lock the dep tree given the Python major version change. There may be fallout from this, e.g. from breaking changes in newer package versions that we were insulated from by previous locking.

¹ https://github.com/nextstrain/docker-base/pull/102

### Testing

- [x] Checks pass
- [x] `docker build -t nextstrain/ncov-ingest:latest .` works locally


victorlin (approved):
Looks good pending revert back to `FROM nextstrain/base`.

tsibley:
Done. Will merge once CI passes again.

PR #355, opened 2022-10-31
feat: scrap checkpoint and parallelize s3 uploads

corneliusroemer:

Uploads to s3 currently consume 6hr, which is about half of the total time taken by ingest.

The main reason for upload taking so long is that uploads are currently sequential.

This PR changes the Snakemake workflow to parallelize uploads. Uploads are done as soon as a file is ready.

I have also removed a checkpoint that does not seem to save any time: if I understand correctly the goal of the checkpoint is to avoid uploading a file if it is unchanged - but in our upload-to-s3 function we already check if the file is identical or not. If it is identical upload is skipped.

The checkpoint means uploads are deferred and the workflow is more complicated. It's cleaner snakemake without.

I tested the Snakefile in dry run mode and it seemed to work fine.

We can cut another hour with PR #354.

Test run:  `nextstrain build --aws-batch --no-download --attach 85c4cf28-27bd-457d-9fc2-12b911c30849  .`

Test worked well, shaving off 3 hours by parallelizing uploads.

### DIFF
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -69,7 +69,7 @@ if config.get("s3_dst") and config.get("s3_src"):
             """
 
 
-checkpoint get_sequences_without_nextclade_annotations:
+rule get_sequences_without_nextclade_annotations:
     """Find sequences in FASTA which don't have clades assigned yet"""
     input:
         fasta = f"data/{database}/sequences.fasta",
@@ -162,25 +162,10 @@ rule combine_alignments:
         fi
         """
 
-def _get_nextclade_output(wildcards):
-    ## the nextclade metadata should represent the entire dataset. If there are new sequences
-    ## this has to be generated; if not then we can use the previous (cached) file.
-    nextclade_sequences_path = checkpoints.get_sequences_without_nextclade_annotations.get().output.fasta
-    output = {
-        "nextclade_tsv": f"data/{database}/nextclade_old.tsv",
-        "aligned_fasta": f"data/{database}/nextclade.aligned.old.fasta",
-    }
-    if os.path.getsize(nextclade_sequences_path) > 0:
-        # Return all two output files here so that they get pulled through
-        # the Snakemake DAG even if we are not uploading them to S3.
-        output["nextclade_tsv"] = f"data/{database}/nextclade.tsv"
-        output["aligned_fasta"] = f"data/{database}/aligned.fasta"
-
-    return output
-
 rule generate_metadata:
     input:
-        unpack(_get_nextclade_output),
+        nextclade_tsv = f"data/{database}/nextclade.tsv",
+        aligned_fasta = f"data/{database}/aligned.fasta",
         existing_metadata = f"data/{database}/metadata_transformed.tsv",
     output:
         metadata = f"data/{database}/metadata.tsv"
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -103,7 +103,6 @@ rule run_nextclade:
     params:
         nextclade_input_dir = temp(directory(f"data/{database}/nextclade_inputs")),
         nextclade_output_dir = temp(directory(f"data/{database}/nextclade")),
-    threads: 64
     output:
         info = f"data/{database}/nextclade_new.tsv",
         alignment = temp(f"data/{database}/nextclade.aligned.upd.fasta"),
@@ -117,8 +116,7 @@ rule run_nextclade:
             {params.nextclade_output_dir} \
             {output.alignment} \
             {output.insertions} \
-            {GENES} \
-            {threads}
+            {GENES}
         """
 
 rule nextclade_info:
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -139,7 +139,7 @@ rule nextclade_info:
                 {input.new_info:q} \
                 -o {output.nextclade_info:q}
         else
-            cp {input.new_info} {output.nextclade_info}
+            mv {input.new_info} {output.nextclade_info}
         fi
         """
 
@@ -156,9 +156,10 @@ rule combine_alignments:
     shell:
         """
         if [[ -s {input.old_alignment} ]]; then
-            cat {input.old_alignment} {input.new_alignment} > {output.alignment}
+            mv {input.old_alignment} {output.alignment}
+            cat {input.new_alignment} >> {output.alignment}
         else
-            cp {input.new_alignment} {output.alignment}
+            mv {input.new_alignment} {output.alignment}
         fi
         """
 
--- a/workflow/snakemake_rules/upload.smk
+++ b/workflow/snakemake_rules/upload.smk
@@ -41,7 +41,16 @@ def compute_files_to_upload(wildcards):
                         "sequences.fasta.xz":           f"data/{database}/sequences.fasta",
 
                         "metadata.tsv.zst":             f"data/{database}/metadata.tsv",
-                        "sequences.fasta.zst":          f"data/{database}/sequences.fasta"}
+                        "sequences.fasta.zst":          f"data/{database}/sequences.fasta",
+
+                        # It shouldn't harm to upload these as upload-to-s3 only updates if hashes differ
+                        "nextclade.tsv.gz":           f"data/{database}/nextclade.tsv",
+                        "aligned.fasta.xz":           f"data/{database}/aligned.fasta",
+
+                        "nextclade.tsv.zst":           f"data/{database}/nextclade.tsv",
+                        "aligned.fasta.zst":           f"data/{database}/aligned.fasta",
+                    }
+
     if database=="genbank":
         files_to_upload["biosample.tsv.gz"] =           f"data/{database}/biosample.tsv"
         files_to_upload["duplicate_biosample.txt.gz"] = f"data/{database}/duplicate_biosample.txt"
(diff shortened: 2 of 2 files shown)


corneliusroemer:
New test runs:
```
nextstrain build --aws-batch --no-download --attach cf1de17f-21f0-4f33-9ee9-f00c1d8e89d6 .
nextstrain build --aws-batch --no-download --attach 268c8c4d-8f96-46a8-b1a3-84cbde65ccec  .
```

joverlee521:
> I have also removed a checkpoint that does not seem to save any time: if I understand correctly the goal of the checkpoint is to avoid uploading a file if it is unchanged - but in our upload-to-s3 function we already check if the file is identical or not. If it is identical upload is skipped.

I haven't looked through the changes yet, but the checkpoint is to check if we need to run Nextclade on any sequences. By removing the checkpoint, the pipeline will always run Nextclade even when it doesn't have new sequences.

corneliusroemer:
> I haven't looked through the changes yet, but the checkpoint is to check if we need to run Nextclade on any sequences. By removing the checkpoint, the pipeline will always run Nextclade even when it doesn't have new sequences.

Since we cache running Nextclade on no files takes 1 minute, which isn't something we need to avoid. The checkpoint complicates parallel uploads and in general complicates the workflow logic. Could it be that the checkpoint predates Nextclade caching?

joverlee521:
> Since we cache running Nextclade on no files takes 1 minute second, which isn't something we need to avoid. The checkpoint complicates parallel uploads and in general complicates the workflow logic. Could it be that the checkpoint predates Nextclade caching?

Ah, this might be a Nextclade v1 issue where an empty FASTA file caused an error?

workflow/snakemake_rules/nextclade.smk
```diff
@@ -84,7 +84,7 @@ checkpoint get_sequences_without_nextclade_annotations:
                 --input_tsv={input.nextclade} \
                 --output_fasta={output.fasta}
         else
-            cp {input.fasta} {output.fasta}
```
joverlee521, line 87: I'm not sure this `mv` is safe to do considering the `input.fasta` (`f"data/{database}/sequences.fasta"`) is required for upload. I don't think there's anything in place to guarantee that the upload of `sequences.fasta` occurs before this rule runs?

workflow/snakemake_rules/nextclade.smk
```diff
@@ -84,7 +84,7 @@ checkpoint get_sequences_without_nextclade_annotations:
...
+            mv {input.fasta} {output.fasta}
```
joverlee521, line 87: I'm not sure this mv is safe to do considering the input.fasta (f"data/{database}/sequences.fasta") is required for upload. I don't think there's anything in place to guarantee that the upload of sequences.fasta occurs before this rule runs?

corneliusroemer:
```suggestion
            cp {input.fasta} {output.fasta}
```

corneliusroemer: Thanks, good point!

ivan-aksamentov:
Context:

https://github.com/nextstrain/ncov-ingest/pull/237#issuecomment-993674154

(Found using GitHub search: https://github.com/nextstrain/ncov-ingest/search?q=checkpoint+is%3Apr&type=issues)


The mutation summary is no longer a thing if I remember correctly, so it seems reasonable to remove the checkpoint.

ivan-aksamentov:
I don't however understand most of the changes other than changing the `checkpoint` to the `rule`. In particular, the commented code and `mv` thingy. Some explanation might help a review. As well as commit messages better than
```
Update workflow/snakemake_rules/nextclade.smk
```

workflow/snakemake_rules/nextclade.smk
```diff
@@ -156,31 +154,17 @@ rule combine_alignments:
     shell:
         """
         if [[ -s {input.old_alignment} ]]; then
-            cat {input.old_alignment} {input.new_alignment} > {output.alignment}
+            mv {input.old_alignment} {output.alignment}
+            cat {input.new_alignment} >> {output.alignment}
         else
-            cp {input.new_alignment} {output.alignment}
+            mv {input.new_alignment} {output.alignment}
```
corneliusroemer, line 160: Copy of a 300GB file takes 30 minutes - since we don't use the old file anywhere else we can just move, or move and cat whatever needs to be added.
@ivan-aksamentov

workflow/snakemake_rules/nextclade.smk
```diff
@@ -103,7 +103,6 @@ rule run_nextclade:
     params:
         nextclade_input_dir = temp(directory(f"data/{database}/nextclade_inputs")),
         nextclade_output_dir = temp(directory(f"data/{database}/nextclade")),
-    threads: 64
```
corneliusroemer, line 106: We usually run ingest with 16 CPUs, assigning 64 threads means that this rule cannot run in parallel to other rules which slows down ingest.

Since we use incremental alignments, the alignment part is fast, so we can just let Nextclade use all cores available to get it done with - this may be slightly subideal from a performance point of view but fine-tuning threads for different parallel rules is a nightmare and not worth it.

For the foreseeable future no problem will arise from parallelizing this rule.

Snakefile
```diff
@@ -93,14 +93,14 @@ onstart:
 onsuccess:
     message = "✅ This pipeline has successfully finished 🎉"
     print(message)
-    if not config.get("keep_all_files", False):
-        print("Removing intermediate files (set config option keep_all_files to skip this)")
-        shell("./bin/clean")
+    # if not config.get("keep_all_files", False):
```
corneliusroemer, line 96: Temporary to help with debugging

workflow/snakemake_rules/upload.smk
```diff
@@ -55,25 +64,31 @@ def compute_files_to_upload(wildcards):
...
-    run:
-        for remote, local in input.items():
-            shell("./bin/upload-to-s3 {params.quiet} {local:q} {params.s3_bucket:q}/{remote:q} {params.cloudfront_domain}")
+        s3_bucket = config.get("s3_dst",""),
+        cloudfront_domain = config.get("cloudfront_domain", ""),
+        local_filename = lambda wildcards: compute_files_to_upload(wildcards)[wildcards.remote_filename]
+    shell:
+        "./bin/upload-to-s3 {params.quiet} {input:q} {params.s3_bucket:q}/{wildcards.remote_filename:q} {params.cloudfront_domain}"
+
+rule upload_all:
```
corneliusroemer, line 86:
```suggestion
rule upload:
```

workflow/snakemake_rules/upload.smk
```diff
@@ -55,25 +64,31 @@ def compute_files_to_upload(wildcards):
...
-        files_to_upload["nextclade.tsv.zst"] =           f"data/{database}/nextclade.tsv"
-        files_to_upload["aligned.fasta.zst"] =           f"data/{database}/aligned.fasta"
     return files_to_upload
 
+def files_to_upload(wildcards):
+    files_to_upload_dict = compute_files_to_upload(wildcards)
+    return [f"data/{database}/{remote_file}.upload" for remote_file in files_to_upload_dict.keys()]
+
 
 rule upload:
```
corneliusroemer, line 74:
```suggestion
rule upload_single:
```

workflow/snakemake_rules/upload.smk
```diff
@@ -55,25 +64,31 @@ def compute_files_to_upload(wildcards):
...
 
+def files_to_upload(wildcards):
+    files_to_upload_dict = compute_files_to_upload(wildcards)
+    return [f"data/{database}/{remote_file}.upload" for remote_file in files_to_upload_dict.keys()]
 
-rule upload:
-    input:
-        unpack(compute_files_to_upload)
+
+rule upload_single:
```
corneliusroemer, line 74: We introduce a rule `upload_single` that uploads a single file as soon as it's done, rather than all files sequentially and in the end.

corneliusroemer:
Thanks for the comment @ivan-aksamentov, I've added some context.

The main change is to start upload (and hence compression) of each file as soon as it's ready, and not do it sequentially in the end.

Compression is slow and there's no reason to do things sequentially. It takes an extra 3 hours to do it this way.

The pattern I use to parallelize is a standard Snakemake pattern.

joverlee521:
@corneliusroemer could you add the context comments to the commit messages? It would be easier for future us to understand context in the commits rather than having to track down the comments in this PR.

joverlee521 (approved):
I rebased to clean up the commits and added additional context from conversations in this PR to the commit messages.

The changes themselves look good to me! Thanks for tackling this @corneliusroemer!

workflow/snakemake_rules/upload.smk
```diff
@@ -55,25 +64,31 @@ def compute_files_to_upload(wildcards):
...
+    input: lambda wildcards: compute_files_to_upload(wildcards)[wildcards.remote_filename]
+    output:
+        touch("data/{database}/{remote_filename}.upload")
+    params:
+        quiet = "" if send_notifications else "--quiet",
+        s3_bucket = config.get("s3_dst",""),
+        cloudfront_domain = config.get("cloudfront_domain", ""),
+        local_filename = lambda wildcards: compute_files_to_upload(wildcards)[wildcards.remote_filename]
```
tsibley, line 82: This `local_filename` param duplicates the `input` and appears unused?

workflow/snakemake_rules/nextclade.smk
tsibley: Making sure I understand the effect and rationale here: by removing the `threads:` parameter and it's passing into `./bin/run-nextclade` (where it ends up as `nextclade run --jobs N`), Nextclade will automatically use all cores available (as it did before when available cores were <= 64) but Snakemake's scheduler will no longer know that. Instead, Snakemake will consider this `run_nextclade` rule to only consume 1 core/thread and thus schedule other jobs to run concurrently.

[Looking at the DAG](https://tsibley.net/paste/2022-11-03wTFK2IDq-ncov-ingest-dag.svg), it seems like the `download_previous_alignment` rule and a bunch of `upload_single` rule jobs and the `upload_raw_ndjson` rule are all candidates for concurrent execution with `run_nextclade` now. These call `download-from-s3` or `upload-from-s3`, which also use all cores available and also aren't annotated to Snakemake as such, so decompression/compression jobs will be in CPU contention with the Nextclade alignment.

The rationale is that this isn't much of an issue in practice and the Nextclade alignment job is fast anyway (at least when not contending for CPU?) and the conceptual overhead of accounting for concurrency is too hard/not worth it.

Have I got that all right? :-)

corneliusroemer: Yes exactly @tsibley, well put!

The only way this oversubscription of threads could harm us is if the bottleneck process that runs on say 1 thread doesn't run at 100% as a result.

If this was the case, we may want to limit threads for some of the compression jobs, but not for Nextclade as it's so fast in most situations anyways.

workflow/snakemake_rules/upload.smk
corneliusroemer, line 82:
```suggestion
```

workflow/snakemake_rules/nextclade.smk
tsibley: Got it. Thanks for confirming!

PR #360, opened 2022-11-28
Add `genbank_accession_rev` to output metadata

### LINKED ISSUE #359: Add genbank_accession_rev field
### Context
For the "open" instance of LAPIS, we use data from https://data.nextstrain.org/files/ncov/open/. It would be very useful if users can know which version of the sequences they are getting from LAPIS but unfortunately, the `genbank_accession` field in the metadata.tsv file only contains the accession without the version number suffix (e.g., `OV377246`).

The mpox data already contains a field with the version number.

### Description
It would be great to have an additional column (`genbank_accession_rev`?) that contains the version number as well (e.g., `OV377246.1`).

(cc @cecivale)

corneliusroemer:
Best practice to include sequence revision - Genbank provides this so lets use it.

As a result, genbank revision metadata will also appear in gisaid as a `?` but I don't think that's so terrible.

This also reminds users that we still don't have GISAID revision numbers which would be really good to have.

Resolves #359

Test build output here: s3://nextstrain-staging/files/ncov/open/branch/add-genbank-accession-rev

### DIFF
--- a/lib/utils/transform.py
+++ b/lib/utils/transform.py
@@ -6,7 +6,7 @@ from typing import Optional, Set, Union
 
 # Note: 'sequence' should NEVER appear in these lists!
 METADATA_COLUMNS = [  # Ordering of columns in the existing metadata.tsv in the ncov repo
-    'strain', 'virus', 'gisaid_epi_isl', 'genbank_accession', 'sra_accession', 'date', 'region',
+    'strain', 'virus', 'gisaid_epi_isl', 'genbank_accession', 'genbank_accession_rev', 'sra_accession', 'date', 'region',
     'country', 'division', 'location', 'region_exposure', 'country_exposure',
     'division_exposure', 'segment', 'length', 'host', 'age', 'sex', 'pango_lineage', 'GISAID_clade',
     'originating_lab', 'submitting_lab', 'authors', 'url', 'title', 'paper_url',


j23414 (approved):

lib/utils/transform.py
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
-import regex
-import pandas as pd
 from datetime import datetime
 from typing import Optional, Set, Union
 
+import pandas as pd
+import regex
+
```
victorlin, line 7: Unnecessary diff

corneliusroemer: Comes from sorted imports

victorlin: I understand, but we aren't enforcing that on any other files in this repo. I'd consider this change in another PR that applies sorted imports to all files at once.

victorlin: An argument against bundling this change in c69e0bf064f8f8e0f588ada475ee846ef64d40aa is that it is confusing to anyone searching through `git blame` to understand file history:

image

Without digging further, this indicates that the change adding `genbank_data_rev` uses `pandas` and `regex` in some fashion, when really it doesn't.

corneliusroemer: Yep you're right! The tricky thing is: I want new files to start off correctly so I'd have to set formatting on a file by file basis.

I know @tsibley isn't a fan of code formatters. I guess we'd need to agree and enforce it otherwise things deteriorate.

PR #362, opened 2022-11-29
Pin shellcheck ubuntu to 20.04

corneliusroemer:
@tsibley correctly identified that the recent shellcheck failures was due to us using ubuntu-latest which caused an implicit upgrade to 22.04 which caused a failure.

Pinning to 20.04 seems to fix it.


tsibley:
Wish I knew why… but I'm trying hard to avoid digging into that now.

PR #364, opened 2022-11-30
Only trigger ncov builds with new data 

joverlee521:

We do not want to waste time and resources to run rebuilds when the
ncov-ingest outputs have not changed. This commit updates the pipeline
to check the outputs from the uploads of the metadata TSV and sequences
FASTA files to verify they have been updated on S3. If the output
includes the message "files are identical" then we skip the trigger
for rebuilds.

Prompted by internal conversation on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1669678401065859). 

### Testing

- [x] Modified trigger-on-new-data script worked as expected locally
- [x] Checks pass

### DIFF
--- a/workflow/snakemake_rules/trigger.smk
+++ b/workflow/snakemake_rules/trigger.smk
@@ -14,14 +14,19 @@ These output files are empty flag files to force Snakemake to run the trigger ru
 rule trigger_rebuild_pipeline:
     message: "Triggering nextstrain/ncov rebuild action (via repository dispatch)"
     input:
-        f"data/{database}/upload.done"
+        metadata_upload = f"data/{database}/metadata.tsv.gz.upload",
+        fasta_upload = f"data/{database}/sequences.fasta.xz.upload",
     output:
         touch(f"data/{database}/trigger-rebuild.done")
     params:
         dispatch_type = f"{database}/rebuild"
     shell:
         """
-        ./bin/trigger ncov {params.dispatch_type}
+        ./bin/trigger-on-new-data \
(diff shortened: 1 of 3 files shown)


joverlee521:
Merging pre-review since a similar trigger has been working fine in the monkeypox ingest pipeline. 
We can always trigger builds manually if something goes wrong.

workflow/snakemake_rules/trigger.smk
```diff
@@ -14,14 +14,19 @@ These output files are empty flag files to force Snakemake to run the trigger ru
 rule trigger_rebuild_pipeline:
     message: "Triggering nextstrain/ncov rebuild action (via repository dispatch)"
     input:
-        f"data/{database}/upload.done"
+        metadata_upload = f"data/{database}/metadata.tsv.gz.upload",
```
corneliusroemer, line 17: ~metadata_upload = f"data/{database}/metadata.tsv.zst.upload",~

~Let's use zst compressed upload as it's:
a) the future
b) done faster so we gain an hour or so~

corneliusroemer: Ignore what I said above - since ncov currently uses the `gz` output we need that one to be ready!

corneliusroemer (approved):
Looks great. There's not much harm in trying this out soon so I'll merge and we just check in the next few days if it worked successfully.

Ah I see you've already merged - even better.

emmahodcroft:
Thanks Jover, this sounds great!

PR #366, opened 2022-12-01
Parallelize raw uploads analogous to non-raw uploads

corneliusroemer:
I noticed we don't do raw uploads in parallel - in contrast to how we now do non-raw uploads.

I've refactored raw uploads to fit the new paradigm, this actually makes it easier to follow (in my view).

Test runs:
- [ ] GISAID fetch and ingest: `8a026939-3a83-4b0e-9205-4f8cbb60b41d`
- [ ] Genbank fetch and ingest: `61cf1e0f-0e8b-43e0-96e7-f93fde518719`

### DIFF
--- a/workflow/snakemake_rules/upload.smk
+++ b/workflow/snakemake_rules/upload.smk
@@ -58,6 +58,12 @@ def compute_files_to_upload():
             files_to_upload.update({
                 "biosample.ndjson.gz": f"data/biosample.ndjson",
                 "biosample.ndjson.zst": f"data/biosample.ndjson",
+
+                "cog_uk_accessions.tsv.gz": f"data/cog_uk_accessions.tsv",
+                "cog_uk_accessions.tsv.zst": f"data/cog_uk_accessions.tsv",
+
+                "cog_uk_metadata.csv.gz": f"data/cog_uk_metadata.csv",
+                "cog_uk_metadata.csv.zst": f"data/cog_uk_metadata.csv",
             })
     return files_to_upload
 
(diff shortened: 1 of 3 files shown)


workflow/snakemake_rules/upload.smk
```diff
@@ -57,24 +35,34 @@ def compute_files_to_upload(wildcards):
...
-    return [f"data/{database}/{remote_file}.upload" for remote_file in files_to_upload_dict.keys()]
+files_to_upload = compute_files_to_upload()
 
 
 rule upload_single:
-    input: lambda wildcards: compute_files_to_upload(wildcards)[wildcards.remote_filename]
+    input: lambda w: files_to_upload[w.remote_filename]
     output:
-        "data/{database}/{remote_filename}.upload"
+        "data/{remote_filename}.upload"
```
joverlee521, line 65: Removing the `{database}/` prefix here will break the triggers for the ncov builds 
https://github.com/nextstrain/ncov-ingest/blob/eafb1d5d32b3d613a6c958815b61625ab9dd88e6/workflow/snakemake_rules/trigger.smk#L17-L18

Is there any particular reason to remove the database prefix here? I think it's better organization to keep the database files in separate directories.

joverlee521 (changes requested):
Thanks for working on these changes @corneliusroemer! 

The separate raw files upload is definitely left over from when we only had a single upload job that uploaded all files. It makes sense to fit the raw files upload into the current parallelized uploads. 

I have one comment below that seems separate from the raw upload change itself.

workflow/snakemake_rules/upload.smk
corneliusroemer: I see - I didn't know these `.upload` touch files were used and the extra `{database}` seemed misleading when not all files were actually in the `{database}` folder (data/biosample.ndjson for example isn't under data/genbank/biosample.ndjson).

I'll happily change!

joverlee521 (approved):

PR #367, opened 2022-12-02
Add subsampled ingest config.yml for faster feedback loop

corneliusroemer:
Right now, ingests even without fetch take hours due to the massive amount of data being piped through pandas and the like.

This PR allows fast CI checks by subsampling whatever arrives from s3 (ndjson, biosample, nextclade alignments etc.), that way, ingest doesn't require 300GB disk space and can live with < 1GB and run in ~5-10 minutes depending on speed of internet connection.

Of course, because we don't load the whole biosample.tsv, there will be differences to a real ingest, so don't take the output data as correct. This is still useful for debugging though. And one could simply not subsample biosample.tsv if we need to test correctness. It's only 10GB so should be ok locally, albeit inconvenient.

To speed up local testing, I'd probably take the various ndjsons out of the snakemake `temp` state - so that once downloaded they don't get deleted - otherwise transforms are hard to debug without changing the pipeline manually to un-temp.

### Testing

- [x] Check that all other configs still work correctly and are not negatively impacted

Test runs:
- [x] GISAID ingest on branch: `750ab0fe-a2e0-48b6-bd94-2eb0002eb842`	https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/750ab0fe-a2e0-48b6-bd94-2eb0002eb842	
- [x] Genbank ingest on branch: `ddf38426-1865-473c-9f99-564e4c126e8b` https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/ddf38426-1865-473c-9f99-564e4c126e8b

### DIFF
--- a/workflow/snakemake_rules/fetch_sequences.smk
+++ b/workflow/snakemake_rules/fetch_sequences.smk
@@ -104,12 +104,13 @@ if config.get("s3_dst") and config.get("s3_src"):
         params:
             file_on_s3_dst=f"{config['s3_dst']}/{database}.ndjson.zst",
             file_on_s3_src=f"{config['s3_src']}/{database}.ndjson.zst",
+            lines = config.get("subsample",{}).get("main_ndjson", 0)
         output:
             ndjson = temp(f"data/{database}.ndjson")
         shell:
             """
-            ./bin/download-from-s3 {params.file_on_s3_dst} {output.ndjson} ||  \
-            ./bin/download-from-s3 {params.file_on_s3_src} {output.ndjson}
+            ./bin/download-from-s3 {params.file_on_s3_dst} {output.ndjson} {params.lines} ||  \
+            ./bin/download-from-s3 {params.file_on_s3_src} {output.ndjson} {params.lines}
             """
 
     rule fetch_biosample_from_s3:
@@ -118,10 +119,11 @@ if config.get("s3_dst") and config.get("s3_src"):
         params:
             file_on_s3_dst=f"{config['s3_dst']}/biosample.ndjson.zst",
             file_on_s3_src=f"{config['s3_src']}/biosample.ndjson.zst",
+            lines = config.get("subsample",{}).get("biosample", 0)
         output:
             biosample = temp("data/biosample.ndjson")
         shell:
             """
-            ./bin/download-from-s3 {params.file_on_s3_dst} {output.biosample} ||  \
-            ./bin/download-from-s3 {params.file_on_s3_src} {output.biosample}
+            ./bin/download-from-s3 {params.file_on_s3_dst} {output.biosample} {params.lines} ||  \
+            ./bin/download-from-s3 {params.file_on_s3_src} {output.biosample} {params.lines}
             """
--- a/workflow/snakemake_rules/fetch_sequences.smk
+++ b/workflow/snakemake_rules/fetch_sequences.smk
@@ -85,6 +85,14 @@ rule fetch_cog_uk_metadata:
             f"rm {output.cog_uk_metadata}"
         )
 
+rule uncompress_cog_uk_metadata:
+    input:
+        "data/cog_uk_metadata.csv.gz"
+    output:
+        cog_uk_metadata = temp("data/cog_uk_metadata.csv")
+    shell:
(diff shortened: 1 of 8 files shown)


workflow/snakemake_rules/fetch_sequences.smk
```diff
@@ -104,12 +104,13 @@ if config.get("s3_dst") and config.get("s3_src"):
         params:
             file_on_s3_dst=f"{config['s3_dst']}/{database}.ndjson.zst",
             file_on_s3_src=f"{config['s3_src']}/{database}.ndjson.zst",
+            lines = config["subsample"].get("main_ndjson", 0)
```
joverlee521: I would change all instances of `config["subsample"]` to `config.get("subsample", {})` since the `subsample` param is optional for this pipeline. 

Currently the automated builds would fail with a `KeyError` because they do not have `subsample` in their configs.

config/subsample_genbank.yaml
```diff
@@ -0,0 +1,19 @@
...
+keep_all_files: True
+fetch_from_database: false
+trigger_rebuild: false
+trigger_counts: false
+
+# How many lines to extract from s3 fetch
+subsample:
+  main_ndjson: 1000
+  biosample: 100000
+  nextclade: 10000
```
joverlee521: I'm slightly confused by the nextclade subsampling. Are you leaving it up to chance of how many of the `main_ndjson` records will be in the subsampled nextclade files?

workflow/snakemake_rules/fetch_sequences.smk
```diff
@@ -65,7 +65,7 @@ rule fetch_cog_uk_accessions:
     message:
         """Fetching COG-UK sample accesions (GenBank only)"""
     output:
-        cog_uk_accessions = temp("data/cog_uk_accessions.tsv")
+        cog_uk_accessions = "data/cog_uk_accessions.tsv" if config.get("keep_temp",False) else temp("data/cog_uk_accessions.tsv")
```
joverlee521, line 68: If you want to keep the temp files, you can just run the snakemake command with `--notemp` flag.

joverlee521 (changes requested):
The changes here come at the cost of making the pipeline more complicated and less readable for users/future us, but I understand the need to make the dev/debug cycle smoother for developers.

The term "subsample" itself is overloaded and can potentially mislead users to thinking that we have a subsampling scheme for the ingest pipeline like we do for ncov builds. I recommend renaming the `subsample` param to `debug_sample` or `dev_sample` to make it abundantly clear that this is used for dev/debug purposes only.

Could you also add documentation to the README for the subsampling parts of the pipeline so that other future devs could use it as well?

workflow/snakemake_rules/fetch_sequences.smk
joverlee521: Ah, I see this got updated in a later commit. Please fix up into the original commit before merging.

config/subsample_genbank.yaml
corneliusroemer: Yes exactly - numbers are big enough so that by chance we definitely have _some_ previously aligned sequences there. All paths are tested but the output data should not be trusted of course.

This also applies to the biosample: random biosample records are chosen, not matched to main_ndjson.

I'll describe it in dev docs.

workflow/snakemake_rules/fetch_sequences.smk
corneliusroemer: Ah that's great to know

corneliusroemer: I've gotten rid of these constructs, thanks!

corneliusroemer:
I think the marginal increased complication is totally worth it for being able to run this out of the box without having to hack some testing input together. It's not that much that changes, mostly in fetch_sequences.

I've added the config to the README so future selfs can find it.

There's no matching of lines, it's all stochastic, but if you pick 10k lines from sqrt(10^7) is 10^3.5 ~ 3k, so if we pick 10k random lines from to two 10m line files, there will be some matches by the birthday problem, hence we cover most paths (as long as merging two files is required only, if we merge 3 we'd need some more - but that's configurable.

PR #369, opened 2022-12-07
Run Nextclade 2nd time on 21L dataset for immune_escape/ace2

corneliusroemer:

To add `immune_escape` and `ace2_binding` scores, we run all sequences through the `sars-cov-2-21L` dataset (in addition to the normal `sars-cov-2` dataset.

Results are cached as usual.

Only 2 new columns are added to the `metadata.tsv`. 

### Testing

- [ ] Genbank ingest on branch
- [ ] GISAID ingest on branch

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -16,7 +16,11 @@ send_notifications = "SLACK_CHANNELS" in os.environ and "SLACK_TOKEN" in os.envi
 ################ work out what steps to run #####################
 #################################################################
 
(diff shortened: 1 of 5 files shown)


corneliusroemer:
The full run on all sequences took only 4 hours with the existing configuration - so we really don't need full runs anymore.

If we update Nextclade datasets, we just delete the nextclade.tsv/alignment.fasta files - done.

Normal ingest will then recompute everything by not using caches.

corneliusroemer:
Tested and it worked perfectly

PR #372, opened 2022-12-12
Refactor /bin/run-nextclade into nextclade.smk

corneliusroemer:
Since we now run Nextclade twice, once with normal dataset
and once with 21L dataset, it's important to keep datasets
separate.

The ./bin/run-nextclade does not allow concurrent running
as it will for example download nextclade twice and download

It's cleaner to turn the legacy shell script into a few
snakemake rules, this is the general way forward for ingest,
anyways

Local tests with config/debug_sample_genbank.yaml passed
This is in any case better than the current buggy solution

Test runs:
- [x] Genbank ingest on branch:  `62f54a4b-7b7b-469d-91d5-601f33d08d75`
- [x] GISAID ingest on branch: `4444f3ee-b50d-4fce-9bba-d8cf667c7e46` -> failed only due to later error


ivan-aksamentov:
I don't see any obvious defects in the diff, but I am also not in the loop about the recent developments. In particular, where this is called, especially twice. But so far LGTM.

ivan-aksamentov (approved):

ivan-aksamentov:
This sounds like something Nextclade should be doing, instead of hacking it in Snakemake, and throwing away alignment and stuff :)
https://github.com/nextstrain/ncov-ingest/blob/aa96fbdf513adfcf9a2d0f7d0f8a0cffec8dfd35/workflow/snakemake_rules/nextclade.smk#L5-L8
I am not sure what needs to be done, but feel free to open a conversation on Slack and/or GitHub.

workflow/snakemake_rules/nextclade.smk
```diff
@@ -102,6 +102,37 @@ rule get_sequences_without_nextclade_annotations:
         echo "[ INFO] Number of {wildcards.reference} sequences to run Nextclade on: $(grep -c '^>' {output.fasta})"
         """
 
+rule download_nextclade_executable:
```
joverlee521, line 105: Do we still need to do this even though Nextclade is part of the nextstrain/base Docker image? Is this to ensure we always have the latest version of Nextclade in our ncov-ingest runs?

workflow/snakemake_rules/nextclade.smk
```diff
@@ -102,6 +102,37 @@ rule get_sequences_without_nextclade_annotations:
...
+    input: "nextclade"
+    output:
+        dataset = "data/nextclade_data/{dataset_name}.zip"
+    shell:
+        """
+        ./nextclade dataset get --name="{wildcards.dataset_name}" --output-zip={output.dataset} --verbose
+        """
 
 GENES = "E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S"
 GENES_SPACE_DELIMITED = GENES.replace(",", " ")
```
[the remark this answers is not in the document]
corneliusroemer: Should probably do next time I touch Nextclade - this doesn't harm as is 🙃

workflow/snakemake_rules/nextclade.smk
corneliusroemer: Not needed, but also doesn't do harm as far as I see it. We can remove it later. Will create an issue to remove in future.

PR #380, opened 2022-12-15
Use touchfiles to indicate when Nextclade cache should not be used

corneliusroemer:
This PR replaces the need to manually do full runs after Nextclade dataset
updates with a touchfile based logic.
Ingest now preferentially uses an empty touchfile as input cache,
in place of the cached nextclade.tsv.zst.
Hence, to tell ingest to not use the cache, one simply needs to add
an appropriately named touchfile to s3
Once ingest has produced new output files, ingest removes the
touchfile, so that future reruns will automatically use the cache again

Instructions on how to add the appropriate touchfile have been added
to the README

Advantages of these changes:
- Less code to maintain, we no longer need all the full-run scripts
  and Github workflows related to full-run
- Less manual intervention necessary, fewer possible errors
- No downtime of cache availability that would happen if we just delete the cache

### Testing:
I put touchfiles in a staging aws bucket, they triggered rerun and
were deleted after upload, so seems to work

I don't think we need to test this with full runs as we'll see whether it works
in our normal runs very quickly

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,67 +127,37 @@ Then, click on 'More'.
 You can then copy your Slack member ID from the menu that appears.
 Enter this into the `slack_member_id` field of your alert configuration.
 
-## Nextclade full run after Nextclade dataset is updated
+## Rerunning Nextclade ignoring cache after Nextclade dataset is updated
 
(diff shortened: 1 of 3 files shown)


corneliusroemer:
We've had a dataset release yesterday so this comes in handy now, will merge and we can revert if it causes any issues.

PR #382, opened 2022-12-20
Update Docker Hub login

victorlin:

Use new token and make the implicit Docker registry explicit.

### Related issue(s)

- https://github.com/nextstrain/private/issues/73

### Testing

- [x] Checks pass


victorlin:
Merging without review since this is a trivial change and checks passed, which confirms that the new token works.

PR #384, opened 2023-01-17
NCBI Datasets: switch to "host-name" field

### LINKED ISSUE #383: Host field is empty
Since very recently, in the open [metadata.tsv](https://data.nextstrain.org/files/ncov/open/metadata.tsv.gz), the host field is (mostly or always) empty. Is it intended to remove the field?

This is unfortunately breaking the open instance of CoV-Spectrum at the moment because it only shows human sequences by default (https://github.com/GenSpectrum/LAPIS/issues/70).

joverlee521:

Previously used the "host-common-name" field to pull in the common names of hosts (e.g. "human"), however this field is empty in recent downloads. This commit switches to the other host field, "host-name", which has the scientific name of the host.

### Related issue(s)

Resolves #383 

### Testing

- [x] Checks pass
- [x] Test run ([GH Action](https://github.com/nextstrain/ncov-ingest/actions/runs/3943510619) - AWS Batch job `89cd19a7-4c18-4a2e-9b29-d9590adcec62`)

### DIFF
--- a/bin/transform-genbank
+++ b/bin/transform-genbank
@@ -61,7 +61,7 @@ NCBI_COLUMN_MAP = {
     "Update date": "date_updated",
     "Virus Pangolin Classification": "pango_lineage",
     "Length": "length",
-    "Host Common Name": "host",
+    "Host Name": "host",
     "Isolate Lineage source": "isolation_source",
     "BioSample accession": "biosample_accession",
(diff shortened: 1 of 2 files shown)


corneliusroemer:
Compared first 1k lines of old and new and indeed, the only change is `homo sapiens` is added where there was nothing `` before

PR #386, opened 2023-01-24
Reduce memory

joverlee521:

Reduce memory usage of the `join-metadata-and-clades` script as this was the most recent culprit an OOM error in the pipeline. 

### Testing

- [x] Checks pass

In my local testing/profiling with [Fil](https://pythonspeed.com/fil/docs/) using a subset of 100,000 records, these changes reduced total memory usage from ~750MiB to ~380MiB. 

Triggered a GISAID ingest run without fetch (`82bce0fa-a633-4a0e-95f0-28c11fcf99be`)

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -40,6 +40,7 @@ column_map = {
     "aaSubstitutions": "aaSubstitutions"
 }
 
+clades_21L_columns = {"immune_escape","ace2_binding"}
 
 def reorder_columns(result: pd.DataFrame):
     """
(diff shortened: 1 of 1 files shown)


corneliusroemer (commented):
Looks good! Let's just check the outputs are unchanged.

We can actually reduce the size of the 21L tsv by using the new column selection feature of Nextclade v2.10.1

But extra memory saving of this is likely low - so not priority. It'd help save a bit of disk space but not really that much either.

joverlee521:
Compared the first million lines from the test run with the existing metadata.tsv and the outputs are the same.

joverlee521:
The Snakemake benchmarks show the `generate_metadata` rule's max memory was ~45.9GiB compared to the current pipeline's max of ~64.7GiB, so we should be able to keep our workflow's memory at 68GiB for a while.

PR #388, opened 2023-01-30
fix: tsv-uniq use column name instead of brittle column order

corneliusroemer:
Should fix bug where new sequences don't get output in nextclade.tsv
Discussed in https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1675087025824499
Requires full runs of both reference datasets

### Testing

- [x] Running locally with debug profile using partial cached tsv

To fully verify that it works, we should check after the full rerun that new sequences actually end up in nextclade.tsv - this will take a few days.

### DIFF
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -193,15 +193,8 @@ rule nextclade_info:
         nextclade_info = f"data/{database}/nextclade{{reference}}.tsv"
     shell:
         """
-        TMPFILE="data/{database}/nextclade{{reference}}.tmp.tsv"
-        if [[ -s {input.old_info} ]]; then
-            mv {input.old_info} $TMPFILE
-            tail -n +2 {input.new_info} >> $TMPFILE
-        else
-            mv {input.new_info} $TMPFILE
-        fi
-        tsv-uniq -H -f seqName $TMPFILE > {output.nextclade_info}
-        rm $TMPFILE
+        tsv-append -H {input.old_info} {input.new_info} \
+        | tsv-uniq -H -f seqName > {output.nextclade_info}
         """
 
 rule combine_alignments:
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -193,9 +193,15 @@ rule nextclade_info:
         nextclade_info = f"data/{database}/nextclade{{reference}}.tsv"
     shell:
         """
-        # Header taken from first non-empty file
-        keep-header {input.old_info:q} {input.new_info:q} -- \
-        tsv-uniq -f 1 > {output.nextclade_info}
+        TMPFILE="data/{database}/nextclade{{reference}}.tmp.tsv"
+        if [[ -s {input.old_info} ]]; then
+            mv {input.old_info} $TMPFILE
+            tail -n +2 {input.new_info} >> $TMPFILE
+        else
+            mv {input.new_info} $TMPFILE
+        fi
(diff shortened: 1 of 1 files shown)


workflow/snakemake_rules/nextclade.smk
```diff
@@ -193,9 +193,15 @@ rule nextclade_info:
...
-        tsv-uniq -f 1 > {output.nextclade_info}
+        TMPFILE="data/{database}/nextclade{{reference}}.tmp.tsv"
+        if [[ -s {input.old_info} ]]; then
+            mv {input.old_info} $TMPFILE
+            tail -n +2 {input.new_info} >> $TMPFILE
+        else
+            mv {input.new_info} $TMPFILE
+        fi
+        tsv-uniq -H -f seqName $TMPFILE > {output.nextclade_info}
+        rm $TMPFILE
```
joverlee521, line 204: Hmm, I'm worried about disk space with the `TMPFILE`. 

Is there any reason we can't use `tsv-append`?
```
tsv-append -H {input.old_info} {input.new_info} \
    | tsv-uniq -H -f seqName > {output.nextclade_info}
```

corneliusroemer: I had no idea `tsv-append` existed! That's neat!

Should work.

corneliusroemer: Regarding diskspace, the tsv is only something like 10-20GB uncompressed, peanuts compared to the fasta which is maybe 400 GB but in any case this is nicer!

ivan-aksamentov: I haven't tried, but is it not enough to just replace `1` to `seqName` in the original code? Why other changes are necessary?

```bash
keep-header {input.old_info:q} {input.new_info:q} -- \
        tsv-uniq -f seqName > {output.nextclade_info}
```

(notice `-f seqName` instead of `-f 1`)

ivan-aksamentov:
I haven't tried, but is it not enough to just replace `1` to `seqName` in the original code? Why other changes are necessary?

```bash
keep-header {input.old_info:q} {input.new_info:q} -- \
        tsv-uniq -f seqName > {output.nextclade_info}
```

(notice `-f seqName` instead of `-f 1`)

corneliusroemer:
> Why other changes are necessary?

They aren't necessary but it's good practice to not rely on exact column indexes for TSV files, but use column names instead. Using the above code would have prevented this issue. Who knows, maybe we'll reorder columns in the future again.

ivan-aksamentov:
@corneliusroemer So if you change `1` to `seqName` it should achieve the same goal?

ivan-aksamentov:
I also disagree with the statement that index is brittle. It's the sturdiest column we have :) The wonky stuff that we have right now in ingest is brittle.

corneliusroemer:
> So if you change 1 to seqName it should achieve the same goal?

Nope, because keep-header does not pass the header on to `tsv-uniq`. I shouldn't have used `keep-header` in the first place, I just didn't know `tsv-append` existed, it's the natural command to use.

I'm not saying that the index column is brittle. I'm saying that relying on column order is brittle.

corneliusroemer:
Test worked, I'll merge, stop any running ingests and restart full runs.

ivan-aksamentov:
> I'm not saying that the index column is brittle. I'm saying that relying on column order is brittle.

Oops. Looks like a aggroed to the wrong index :D

ivan-aksamentov:
> Nope, because keep-header does not pass the header on to tsv-uniq.

Sigh.

Implementing a relational database in bash is so tricky ;)

PR #389, opened 2023-01-31
Maintain status quo by migrating from `clade` to `clade_legacy`

corneliusroemer:
Nextclade will shortly change the content of the `clade` column
in SC2 datasets. Rather than being long `22F (Omicron)` they
will be just year letter `22F`, without WHO label.

For backwards compatibility, Nextclade datasets have had a
`clade_legacy` column for a while now, which contains
exactly what `clade` has been until now.

So by switching from `clade` to `clade_legacy` in ingest
nothing will change in any ncov-ingest outputs.

See more about the context of the migration here:
https://github.com/nextstrain/nextclade_data/issues/39

## Testing

- [x] Local debug profile test completes successfully: `3dd25c13-5bb8-4fe6-8664-dccdca262082`
- [x] Ingest on branch completes successfully
- [x] Manual verification that all relevant mentions of `clade` column in code have been changed to `clade_legacy`

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -6,6 +6,7 @@ import pandas as pd
 import numpy as np
 
 INSERT_BEFORE_THIS_COLUMN = "pango_lineage"
+COLUMN_TO_REORDER = "Nextstrain_clade"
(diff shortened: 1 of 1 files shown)


tsibley (approved):
Seems ok to me!

Issue #392, opened 2023-03-01
Malformed/wrong nextclade.tsv


chaoran-chen:
I think that [nextclade.tsv](https://data.nextstrain.org/files/ncov/open/nextclade.tsv.gz) of the open dataset is malformed (or wrong in a different way). The `deletions` column contains values for mutations. 

One way to reproduce the error is using [tsv-utils](https://github.com/eBay/tsv-utils) to inspect the `deletions` column:

```
➜  tsv-select -H -f deletions nextclade.tsv | grep C241T
C241T,G443A,T670G,C2790T,T2954C,C3037T,C3096T,G4184A,C4321T,C8956T,C9344T,A9424G,C9534T,C10029T,C10198T,G10447A,C10449A,C11750T,G12160A,C12880T,T14257C,C14408T,C15714T,C16733T,G16935A,A17039G,C17410T,A18163G,C19955T,A20055G,C21618T,C21714T,G21987A,T22200G,G22578A,G22599C,C22674T,T22679C,C22686T,A22688G,G22775A,G22813T,T22882G,A22893C,T22917G,G22992A,C22995A,A23013C,T23018G,A23055G,A23063T,T23075C,A23403G,C23525T,T23599G,C23604A,C23758A,C23854A,G23948T,A24424T,T24469A,C25000T,C25584T,C26060T,G26529A,C26577G,G26709A,C27807T,C27889T,A28271T,C28311T,C28312T,G28681T,G28881A,G28882A,G28883C,A29510C
C241T,T670G,C2790T,T2954C,C3037T,G4184A,C4321T,C9344T,A9424G,C9534T,C10029T,C10198T,G10447A,C10449A,C11750T,G12160A,C12880T,T14257C,C14408T,C15714T,G16935A,A17039G,C17410T,A18163G,C19955T,A20055G,A21417G,C21618T,G21987A,T22200G,G22578A,G22599C,C22674T,T22679C,C22686T,A22688G,G22775A,G22813T,T22882G,A22893C,G22992A,C22995A,A23013C,T23018G,A23055G,A23063T,T23075C,A23403G,C23525T,T23599G,C23604A,C23854A,G23948T,A24424T,T24469A,C25000T,C25584T,C26060T,G26529A,C26577G,G26709A,C27807T,C27889T,A28271T,C28311T,C28312T,G28681T,G28881A,G28882A,G28883C,A29510C
[...]

PR #393, opened 2023-03-06
feat: add nextclade (dataset) version info to nextclade.tsv

corneliusroemer:
The info gets added as extra columns nextclade_version, dataset_version
and run_timestamp on new nextclade lines only

This way we have metadata for each line in the cached data
and can automatically rerun when there is a mismatch

The extra info does not end up in metadata as we only add columns
we specify explicitly

In a second step (not contained in this PR) we can trigger reruns when there's a mismatch.

## Testing
I tested this locally with debug config, worked well

- [x] Will trigger a full genbank test run (ingest only) just in case: 16181387-2802-45b5-bb40-d83a736d625a

Note: this requires a manual rerun, otherwise tsv-utils won't be happy with differing column counts

### DIFF
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -174,7 +174,7 @@ rule run_nextclade:
     params:
         genes=GENES_SPACE_DELIMITED,
     output:
-        info=f"data/{database}/nextclade{{reference}}_new.tsv",
+        info=f"data/{database}/nextclade{{reference}}_new_raw.tsv",
         alignment=temp(f"data/{database}/nextclade{{reference}}.aligned.upd.fasta"),
     shell:
         """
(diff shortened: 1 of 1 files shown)


joverlee521 (commented):
Instead of adding these as columns to the nextclade.tsv, I wonder if we should just add version number and dataset timestamp to the filename? (Especially if we expect/want the data to all be generated from the same versions.)

Then, the pipeline can check for the file `nextclade_<nextclade_version>_<dataset_version>.tsv`. If the file does **not** exist, then the pipeline can just create an empty file and do a full-run without cache.

corneliusroemer:
> Instead of adding these as columns to the nextclade.tsv, I wonder if we should just add version number and dataset timestamp to the filename? (Especially if we expect/want the data to all be generated from the same versions.)
>Then, the pipeline can check for the file nextclade_<nextclade_version>_<dataset_version>.tsv. If the file does not exist, then the pipeline can just create an empty file and do a full-run without cache.

Possible but not ideal.

We don't really need to run for every nextclade version, nor do we need to usually run it for new 21L datasets (i.e. saving 50% of full runs).

We may end up running for all nextclade versions just to avoid breaking things, but this is not a strict requirement.

Checking versions are right is as easy as looking at the second line of the file, so not difficult.

Extra space is minimal.

One upside for encoding in file name would be that we have an easily accessible archive of previous versions, but that's already available through aws s3 cli - so not something major gained.

joverlee521:
> We don't really need to run for every nextclade version, nor do we need to usually run it for new 21L datasets (i.e. saving 50% of full runs).

Hmm, maybe I misunderstood. I thought the purpose of adding version info is to be able to automatically do full runs when there's a mismatch in the versions so we can avoid issues like #392.

corneliusroemer:
There are three reasons:
1. @chaoran-chen requested to have nextclade dataset version in nextclade.tsv for covSpectrum
2. We want to improve debugging of cache rerun issues
3. As much as possible automate the necessary rerunning when new datasets and (big) change to Nextclade come out

This PR addresses 1 and 2 and makes a flexible implementation of 3 in the future easy.

Rerun logic is currently as follows:

**must**:
- Main dataset with every version
- Changes to nextclade column output (until now it's been minor version, never patch)
- 21L dataset when the immune escape score calculation algorithm changes (so far 0 times in ~3 months)

**may**
- Run with every nextclade release, even patch, just to be sure - we also don't that often release nextclade so that's cheap

**unnecessary**
- 21L datasets releases without algorithm changes

Having nextclade/dataset version and timestamp in the tsv for every record means we are flexible as to when we want to rerun.

Sure we could append nextclade and dataset versions to the file path and pull caches dynamically from s3 after we know the version of nextclade and dataset. For 21L we could not invalidate for new datasets by default. And only use minor version. I guess both works. But we loose information in those cases in case we _do_ want to debug where something came from.

Disadvantage of putting it in file name is that there is no stable path for the latest version of the nextclade.tsv. Unless we add touch files, but that's duplicating things.

Does that make sense?

corneliusroemer:
I'm going to merge this later unless there are objections. We can always add or replace with file names if we prefer. I don't see a downside of this approach and it's tested.

PR #396, opened 2023-03-15
transforms/AbbreviateAuthors: remove double quotes

joverlee521:

For some reason, certain author values from GISAID are double quoted and this leaves an unclosed quote when we truncate the authors to `"...et al`. This unclosed quote can lead to errors in table readers such as Excel as reported in #395.

This commit strips all `"` from the truncated authors string as an easy and quick fix for the issue.

### Related issue(s)

Fixes #395

### Testing

- [x] Checks pass
- [x] Test GISAID ingest w/o fetch ([GH Action](https://github.com/nextstrain/ncov-ingest/actions/runs/4431104248) | [AWS Batch job](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/db1c24e2-0fdd-40b9-803a-40e0db302e28) `db1c24e2-0fdd-40b9-803a-40e0db302e28`)
- [ ] Test GenBank ingest w/o fetch ([GH Action](https://github.com/nextstrain/ncov-ingest/actions/runs/4431105670) | [AWS Batch job](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/61c7d176-1350-4df7-8352-71f0edfa913b) `61c7d176-1350-4df7-8352-71f0edfa913b`)

### DIFF
--- a/lib/utils/transformpipeline/transforms.py
+++ b/lib/utils/transformpipeline/transforms.py
@@ -389,6 +389,11 @@ class AbbreviateAuthors(Transformer):
             if not entry['authors'].strip('. ').endswith(" et al"): # if it does not already finishes with " et al.", add it
                 entry['authors'] += ' et al'
 
+            # Strip all double quotes from the authors value to prevent
+            # unclosed quotes from causing errors in table readers when reading
+            # the transformed metadata TSV
+            entry['authors'] = entry['authors'].replace('"', '')
+
         return entry
 
 


emmahodcroft:
LGTM! (From code, not seeing all testing)

lib/utils/transformpipeline/transforms.py
```diff
@@ -389,6 +389,11 @@ def transform_value(self, entry: dict) -> dict:
             if not entry['authors'].strip('. ').endswith(" et al"): # if it does not already finishes with " et al.", add it
                 entry['authors'] += ' et al'
 
+            # Strip all double quotes from the authors value to prevent
+            # unclosed quotes from causing errors in table readers when reading
+            # the transformed metadata TSV
+            entry['authors'] = entry['authors'].replace('"', '')
```
tsibley, line 395: Embedded quotes in column values in TSV/CSV are fine and well-supported. We shouldn't have to do this mangling, and it's a band-aid anyway as any other column could also contain unbalanced quotes.

Why is our CSV emitter not properly escaping the quotes in our values?

tsibley (changes requested):

lib/utils/transformpipeline/transforms.py
joverlee521: Good point! Looking at this again, in the output from `transform-gisaid`, the authors values with the unclosed `"` are double quoted. I believe this is the default behavior for `csv.DictWriter`. 

Then looking at the output from `join-metadata-and-clades`, these are still double quoted. This is the expected default behavior for `pandas.to_csv`. 

So, it appears that the issue is purely in the acknowledgements.tsv which would make this an Auspice issue.

jameshadfield: 🙈

joverlee521:
Closing this PR and moving the corresponding issue to the auspice repo.

jameshadfield:
Currently (i.e. prior to this PR) we will export the following: 

```json
"author": {
  "author": "\"Flores et al",
  "value": "\"Flores et al"
},
```

This will be handled without error by Auspice (including [in TSV export](https://github.com/nextstrain/auspice/pull/1649)), with the leading quote retained. A tooltip would look like:
image


Why do we have a leading `"` in the JSON at all?
* are the authors provided to us (via GISAID) with an unbalance quote, which we're preserving?
* are the authors provided to us surrounded by quotes when we collapse them to "et al" we end up with unbalanced quotes?
* are we are adding the leading quote when we collapse them?

I would think this leading quote shouldn't be here.

tsibley:
It's a data quality issue with GISAID, yeah. We could adjust the `transform-gisaid` pipeline to unwrap surrounding quotes from the authors field—not just blindly strip all quotes—if we want to bother with fixing these up. Not sure it seems worth it to me, there are bigger fish to fry, and https://github.com/nextstrain/auspice/pull/1649 will fix our bug here. Did anyone look at how prevalent the surrounding quotes are in the GISAID data?

emmahodcroft:
I didn't look at all GISAID data, but in the dataset where I originally detected the problem, from the original issue:
>This (unclosed `"`) happens at least 17 times in a dataset of 6300 seqs

This is just looking at the `author` field.

PR #397, opened 2023-03-22
Shellcheck

tsibley:
See commit messages.

### Related issue(s)

https://github.com/nextstrain/.github/pull/37

### Testing

- [x] Checks pass


tsibley:
With a [temporary commit](https://github.com/nextstrain/ncov-ingest/commit/b00e4e9c9fb15777611eccef56bf0e977810b12d) to point to the shellcheck action on the not-yet-merged branch, [checks here pass](https://github.com/nextstrain/ncov-ingest/actions/runs/4493001995/jobs/7904268225).

Checks show as failed now because I dropped the temporary commit.

.github/workflows/shellcheck.yml
```diff
@@ -6,12 +6,7 @@ on:
 
 jobs:
   shellcheck:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
victorlin, line 9: _non-blocking_

Any reason not to use `ubuntu-latest` here, or not to use `ubuntu-22.04` in other workflows?

victorlin (approved):

.github/workflows/shellcheck.yml
tsibley: Since this workflow was already using an explicit version, I kept it so when updating it.

I guess we could set a blanket policy about preferring one of `-latest` or `-x.y.z` OS versions in CI… but I think it's fine to leave it to individual judgement on a case-by-case basis.

victorlin: I'm fine with individual judgement. I couldn't tell from the file contents alone but wasn't curious enough to look at the blame. Now I see the judgement in 04cc40a2a4571b46f29a74af3dda28e457e7e1a0.

For this example, I'm not sure "this workflow was already using an explicit version" is a good enough reason to continue doing so. I assume the original reason was tied to having `sudo snap install --channel=edge shellcheck` and now that basis is gone.

Anyways, not a big deal!

tsibley: Relevant to that judgement is my comment: https://github.com/nextstrain/ncov-ingest/pull/362#issuecomment-1334495448 :upside_down_face:

tsibley: I would have no issue with reverting this workflow back to ubuntu-latest.

> For this example, I'm not sure "this workflow was already using an explicit version" is a good enough reason to continue doing so. I assume the original reason was tied to having `sudo snap install --channel=edge shellcheck` and now that basis is gone.

I dunno, maybe, maybe not… I'm not sure it was related to the snap-based install. Note that even with the new shellcheck action, we're still using the image-provided shellcheck not a separately downloaded version.

My thinking (FWIW and not saying it's not flawed!) was roughly: There was reason to use a pinned OS version in the past for this job. Since I'm in here working on things anyway, I might as well bump that version since I can address any breakages at the same time I'm getting the new thing going. But since there's still reason to think it might break again on some future OS update, I'm going to leave it pinned.

tsibley: The proper way to do what I did would have been the separate the OS version bump from the shellcheck switch… and that would have made things clearer.

(But this was already something I was trying to stop spending too much time on, so I didn't, and well, maybe that was the wrong choice!)

PR #399, opened 2023-04-07
Add more clade columns to metadata

emmahodcroft:
We have updated Nextclade to separate out clade names (`23A`, `Omicron`, `23A (Omicron)`), but we have never followed up by doing the same in ingest, to give us more columns both to work with, and some overlap before we potentially change what `Nextstrain_clade` is. 
This PR adds `clade_nextstrain` (`23A`) and `clade_who` (`Omicron`).
It also adds `clade_legacy` (which would be `23A (Omicron`) - I'm happy to remove this after discussion about what our longer-term plans are - if we don't plan to adjust `Nextstrain_clade` then we don't need it.

~~This also adds a column `display_clade` which currently is the legacy clade (`23A (Omicron)`) but could be changed to reflect whatever we want to display, and be guaranteed to be 'unstable' (don't base any workflows on this).~~
12 Apr: This column is now removed in favour of doing this 'name creation' for the labels somewhere in `ncov`.

This pulls in all available 'new' clade names from Nextclade, allowing users time to start changing workflows to use other columns apart from `Nextstrain_clade` (if we may want to change this in future).
It also gives us more flexibility on columns we can use for things like labelling but also for processing/filtering.
~~Finally it gives us a new column `display_clade` that we can modify to help us easily show whatever type of clade labels we want in the future, without breaking workflows (hopefully, if we make it know it's not stable).~~ _See above_

### Related issue(s)

This is related to wanting to change the display labels for ncov builds, but running into issues with coloring etc. [ncov ingest issue 1050](https://github.com/nextstrain/ncov/pull/1050)

### Testing

This needs to be tested, but also - feedback on the column names is appreciated.

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -16,6 +16,10 @@ reference_day = datetime(2020,1,1).toordinal()
 
 column_map = {
     "clade_legacy": "Nextstrain_clade",
+    "clade_nextstrain": "year_letter_clade",
+    "clade_who": "clade_who",
+    "clade_legacy": "clade_legacy",
+    "clade_legacy": "display_clade",
     "Nextclade_pango": "Nextclade_pango",
     "immune_escape": "immune_escape",
     "ace2_binding": "ace2_binding",
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -18,7 +18,6 @@ column_map = {
     "clade_legacy": "Nextstrain_clade",
     "clade_nextstrain": "clade_nextstrain",
     "clade_who": "clade_who",
-    "clade_legacy": "clade_legacy",
     "Nextclade_pango": "Nextclade_pango",
     "immune_escape": "immune_escape",
     "ace2_binding": "ace2_binding",
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -16,10 +16,9 @@ reference_day = datetime(2020,1,1).toordinal()
 
 column_map = {
     "clade_legacy": "Nextstrain_clade",
-    "clade_nextstrain": "year_letter_clade",
+    "clade_nextstrain": "clade_nextstrain",
     "clade_who": "clade_who",
     "clade_legacy": "clade_legacy",
-    "clade_legacy": "display_clade",
     "Nextclade_pango": "Nextclade_pango",
     "immune_escape": "immune_escape",
     "ace2_binding": "ace2_binding",


bin/join-metadata-and-clades
```diff
@@ -16,6 +16,10 @@ reference_day = datetime(2020,1,1).toordinal()
 
 column_map = {
     "clade_legacy": "Nextstrain_clade",
+    "clade_nextstrain": "year_letter_clade",
+    "clade_who": "clade_who",
+    "clade_legacy": "clade_legacy",
+    "clade_legacy": "display_clade",
```
tsibley, line 22: `year_letter_clade` seems like it weirdly omits that this is the Nextclade/Nextstrain clade… esp. given `clade_who`. What about being consistent with `clade_nextstrain` or similar instead? (I know this gets potentially confusing with `Nextstrain_clade`, but that's already sort of an outlier of a name.)

Instead of `display_clade`, I'd probably just call it `clade`. Assuming that it will continue to include multiple naming systems, e.g. in the style of `23A (Omicron, XBB.1.5)`. Since this is ncov-ingest and we're producing general data files, it doesn't seem appropriate to pre-suppose what column will be used for display by downstream, and the bare `clade` being inclusive of multiple naming systems seems appropriate.

I'd probably drop

```python
    "clade_legacy": "clade_legacy",
```

until we need it, i.e. if/when we change `Nextstrain_clade`. (We might chose to just leave `Nextstrain_clade` alone indefinitely and instead switch ncov to using `display_clade` (or `clade`).) If we _do_ use `clade_legacy`, we might want to consider a more specific name for it than just "legacy".

emmahodcroft: I'd be happy to rename `year_letter_clade` to `clade_nextstrain` as long as it doesn't seem too likely to be confusing. 

For the other things, more discussion on the long-term strategy would beneficial I think. While `display_clade` is probably a little specific as you say, my hope would be for some kind of name that indicates one shouldn't use it in workflows (or use it at your own peril) - so that we can feel more free to adjust it as it suits our needs. I feel like `clade` might be too generic to warn people away from using it - but perhaps something like `clade_combination` or `clade_concat` or similar, might be 'weird sounding' enough that people are more likely to stick to the other options? I'm definitely not overly attached to the current name! On the other hand, people may actually like having something that is guaranteed to always match up with "what Nextstrain is displaying"? (I could see some use to this, though probably narrow.)

I'm also happy to drop `clade_legacy` if we think we may just keep `Nextstrain_clade` as legacy for the foreseeable future. But, if we think we might not do this, then I think having some overlap period where you can switch workflows over (aka, where using both `Nextstrain_clade` works and using `clade_legacy` works) is most useful. Again, would be useful to maybe make a decision around which path we think we'll do - but that's just why I made the PR, to hopefully inspire some discussions/decisions & push forward! 😁

jameshadfield, line 22: > Finally it gives us a new column display_clade that we can modify to help us easily show whatever type of clade labels we want in the future, without breaking workflows 

This colum will be identical to "Nextstrain_clade" (both are just using Nextclade's "clade_legacy"). Is the idea to introduce this column in this PR in preperation for a future new column in Nextclade which will become the "display_clade"? Or will there be code in ncov-ingest to generate this name?

As far as I understand it a value such as “21K (BA.1)” doesn't exist in the metadata ncov-ingest produces as it currently stands.

emmahodcroft: Yea, it indeed may be best to not have the `display_clade` column at all, and have this added internally in `ncov` (see Tom's arguments above). 

I mostly added this because from the complexity of changing stuff in ncov it seemed like a possibly easy solution, but if we can come up with better ways to do this, I'd be happy to drop having this column in metadata!

emmahodcroft: Given the thinking and the other PRs in progress, I'd be happy to drop `display_clade` entirely, if we think we can create this internally in `ncov` or otherwise find a way to construct the branch labels we want! I mostly added this because with the complexity of some of the PRs it seemed like the most-least-destructive way to do this, but I now think a better solution (if we can engineer it) would be more stable & clearer.

Yes, happy to wait to introduce the `clade_legacy` until we know exactly what our plan is, as long as we've got some kind of overlap (which we do at the moment, presuming we don't alter this PR to immediately change `Nextstrain-clade`)

tsibley (approved):
+1 from me.

Ideally, before merge, I'd say squash the two commits here together and update the commit message of the first to reflect the current reality rather than the outdated original (like you've updated the PR description).

emmahodcroft:
Decided in meeting 13 Apr, just introduce `clades_nextstrain` & `clades_who` for now, and circle back to adding more or changing things in future.

emmahodcroft:
Ok, I can see the option within github.com to "squash and merge" which will squash the commits as Tom suggests above.
Currently looking into how to edit commit messages.

tsibley:
@emmahodcroft Not a big deal here. Don't worry about it. I could have been clearer that I meant that when describing it as "Ideally…". :pray:

emmahodcroft:
Thanks Tom! I'll do the squash so things are tidier in the commits, but won't adjust the message. Still, nice to even know that the squash option is there! 😁

corneliusroemer:
Had a look at our metadata.tsv and we now indeed have `clade_nextstrain` and `clade_who` in there.

emmahodcroft:
Fantastic!

PR #403, opened 2023-06-14
Update divisions to countries

joverlee521:
The following divisions have been updated to countries:

* British Virgin Islands
* Cayman Islands
* Gibraltar

This change keeps these locations consistent with being top-level geographic units in GISAID and allows them to be subsampled as individual countries within the ncov builds.

This change was motivated by a question from sequence submitters asking why their sequences from Cayman Islands was not showing up in the Nextstrain builds: https://bedfordlab.slack.com/archives/CBVHPFMGX/p1686581931753179.


corneliusroemer:
I'd prefer if we discuss this slightly more. How do we currently decide what is a country and what isn't? It would be nice if we could follow some international standards, like those countries that have ISO 3-letter codes. Then it becomes much easier to do things like population weighting etc, see for example here: https://github.com/neherlab/flu_frequencies/blob/2671ca0d52c4c1d588e1d7870dae64920c670594/defaults/iso3_to_pop.tsv

corneliusroemer:
Actually all the three places that get promoted _have_ ISO3 codes, so I guess we can merge and just try to do even more breaking out into separate countries (if possible)

Issue #394, opened 2023-03-09
Publish gene translation files


chaoran-chen:
Hi. May I bother you with another request? Would it be possible for you to also publish the gene translation fasta files for the open data?

(We need these files for LAPIS. At the moment, we import most of the data including nextclade.tsv directly from the downloaded files but then still run nextalign ourselves on all sequences just to obtain the AA sequences.)

corneliusroemer:
Yes of course, I'll have a look. This shouldn't be too difficult.

By the way, the nextclade.tsv now contains timestamp of dataset, as well as version of nextclade and time of run.

chaoran-chen:
Hi @corneliusroemer, I just would like to remind you of this issue. LAPIS v2 is coming closer and it would be amazing if we can avoid any data preprocessing for the open instance and just use the downloaded data from Nexttrain.

chaoran-chen:
Hi @corneliusroemer, LAPIS v2 is now very close and we have started setting up the test server (https://github.com/GenSpectrum/LAPIS/issues/124). It would be really amazing if we can download the translated sequences directly so that we don't have to set up the whole preprocessing pipeline again (with caching, etc.) and redo all the work. Do you think that it would be possible to implement this issue within the next two weeks?

chaoran-chen:
@corneliusroemer, do I understand correctly that after the next run, the translations will be available at https://data.nextstrain.org/files/ncov/open/translation_E.fasta.zst?

corneliusroemer:
@chaoran-chen Yes, that's right. Translations should be there in <24hr

PR #404, opened 2023-06-15
Add translations as extra ingest output 

corneliusroemer:
Feature request from @chaoran-chen which would save covSpectrum a lot of money by not having to run nextclade for open data anymore.

Translations will be uploaded to same bucket as "aligned.fasta.zst"
but with "aligned" replaced by "translation_N" (etc. for other genes, E, S)
Cache is used like for aligned sequences
Only producing zst compressed output as we're moving there in the future
and xz/gz is only for backwards compatibility of existing scripts

Resolves https://github.com/nextstrain/ncov-ingest/issues/394

Testing
- I tested locally with the debug config file
- I will run a genbank trial ingest to see, if that works should be good to merge, if there's a bug we can always revert: `nextstrain build --aws-batch --attach ce172ab3-0ece-48df-83e8-0fbea3ba02f5 /home/runner/work/ncov-ingest/ncov-ingest`: previous run had an error, trying again :(

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,6 +1,10 @@
 from subprocess import CalledProcessError
 import os
 
+GENES = "E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S"
+GENES_SPACE_DELIMITED = GENES.replace(",", " ")
+GENE_LIST = GENES.split(",")
+
 #################################################################
 ####################### general setup ###########################
 #################################################################
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -31,7 +31,7 @@ from shlex import quote as shellquote
 
 wildcard_constraints:
     reference="|_21L",
-    seqtype="[^.]+",
+    seqtype="aligned|translation_[^.]+",
 
 
 rule create_empty_nextclade_info:
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -30,6 +30,7 @@ Produces the following outputs:
 
 wildcard_constraints:
     reference="|_21L",
+    seqtype="[^.]+",
 
 
 rule create_empty_nextclade_info:
@@ -43,7 +44,15 @@ rule create_empty_nextclade_aligned:
     message:
         """Creating empty NextClade aligned cache file"""
     output:
-        touch(f"data/{database}/nextclade.aligned.old.fasta"),
+        alignment=f"data/{database}/nextclade.aligned.old.fasta",
+        translations=[
+            f"data/{database}/nextclade.translation_{gene}.old.fasta"
+            for gene in GENE_LIST
+        ],
+    shell:
+        """
+        touch {output}
+        """
 
 
 # Only include rules to fetch from S3 if S3 config params are provided
@@ -81,13 +90,13 @@ if config.get("s3_dst") and config.get("s3_src"):
         ## (1) race condition. This file may be updated on the remote after download_nextclade has run but before this rule
         ## (2) we may get `download_nextclade` and `download_previous_alignment` from different s3 buckets
         params:
-            dst_source=config["s3_dst"] + "/aligned.fasta.zst",
-            src_source=config["s3_src"] + "/aligned.fasta.zst",
+            dst_source=config["s3_dst"] + "/{seqtype}.fasta.zst",
+            src_source=config["s3_src"] + "/{seqtype}.fasta.zst",
             dst_rerun_touchfile=config["s3_dst"] + "/nextclade.tsv.zst.renew",
             src_rerun_touchfile=config["s3_dst"] + "/nextclade.tsv.zst.renew",
             lines=config.get("subsample", {}).get("nextclade", 0),
         output:
-            alignment=temp(f"data/{database}/nextclade.aligned.old.fasta"),
+            alignment=temp(f"data/{database}/nextclade.{{seqtype}}.old.fasta"),
         shell:
             """
             ./bin/download-from-s3 {params.dst_rerun_touchfile} {output.alignment} 0 ||  \
@@ -156,10 +165,6 @@ rule download_nextclade_dataset:
         """
 
 
-GENES = "E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S"
-GENES_SPACE_DELIMITED = GENES.replace(",", " ")
-
-
 rule run_nextclade:
     message:
         """
@@ -173,9 +178,19 @@ rule run_nextclade:
         sequences=f"data/{database}/nextclade{{reference}}.sequences.fasta",
     params:
         genes=GENES_SPACE_DELIMITED,
+        translation_arg=lambda w: (
+            f"--output-translations=data/{database}/nextclade{w.reference}.translation_{{gene}}.upd.fasta"
+            if w.reference == ""
+            else ""
+        ),
+        translations=lambda w: [
+            f"data/{database}/nextclade{w.reference}.translation_{gene}.upd.fasta"
+            for gene in GENE_LIST
+        ],
     output:
         info=f"data/{database}/nextclade{{reference}}_new_raw.tsv",
-        alignment=temp(f"data/{database}/nextclade{{reference}}.aligned.upd.fasta"),
+        alignment=f"data/{database}/nextclade{{reference}}.aligned.upd.fasta",
+        touchfile=touch(f"data/{database}/nextclade{{reference}}.touch"),
     shell:
         """
         if [[ -s {input.sequences} ]]; then
@@ -184,9 +199,10 @@ rule run_nextclade:
             --input-dataset={input.dataset} \
             --output-tsv={output.info} \
             --genes {params.genes} \
+            {params.translation_arg} \
             --output-fasta={output.alignment}
         else
-            touch {output.info} {output.alignment}
+            touch {output.info} {output.alignment} {params.translations} {output.touchfile}
             echo "[ INFO] Skipping Nextclade run as there are no new sequences"
         fi
         """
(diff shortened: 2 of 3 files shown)


corneliusroemer:
It works! Check through e.g.
```
aws s3 cp s3://nextstrain-staging/files/ncov/open/branch/publish-translations/translation_E.fasta.zst .
```

Would be good to double check translations occasionally after reruns. @chaoran-chen, it would be super useful if you could do sanity checks on occasion and report back, as you'll be the main user of this file at least for now (I can't think of anyone else using these, we likely won't have a use at this point).

chaoran-chen:
Fantastic, thank you very much, Cornelius! Yes, we will put it into use very soon and then keep you updated if anything works!

workflow/snakemake_rules/nextclade.smk
```diff
@@ -241,17 +257,17 @@ rule nextclade_info:
...
-        """
-        Generating full alignment by combining newly aligned sequences with previous (cached) alignment
-        """
+    """
+    Generating full alignment by combining newly aligned sequences with previous (cached) alignment
+    """
     input:
-        old_alignment=f"data/{database}/nextclade.aligned.old.fasta",
-        new_alignment=f"data/{database}/nextclade.aligned.upd.fasta",
+        nextclade_finished=f"data/{database}/nextclade.touch"
```
joverlee521, line 264:
```suggestion
        nextclade_finished=f"data/{database}/nextclade.touch",
```

workflow/snakemake_rules/nextclade.smk
```diff
@@ -241,17 +257,17 @@ rule nextclade_info:
...
+        old_alignment=f"data/{database}/nextclade.{{seqtype}}.old.fasta",
     output:
-        alignment=f"data/{database}/aligned.fasta",
+        alignment=f"data/{database}/{{seqtype}}.fasta",
     params:
         keep_temp=config.get("keep_temp", "false"),
+        new_alignment=f"data/{database}/nextclade.{{seqtype}}.upd.fasta",
```
joverlee521, line 270: Can this be an input rather than a param? It shares the same wildcards as the `old_alignment` so I don't think there should be an issue.

joverlee521 (changes requested):
Thanks for cleaning up the commits @corneliusroemer! It's much easier to review now 🙏 

If I'm understanding the workflow correctly, the 21L version only needs to output the nextclade.tsv while the main Nextclade run needs to output the nextclade.tsv, aligned.fasta, and translation_gene.fasta.  I think a lot of the complexity can be avoided by splitting the `run_nextclade` rule into `run_nextclade` and `run_nextclade_21L`.

workflow/snakemake_rules/nextclade.smk
joverlee521: Ah, I see. My question was answered in [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1686935320199049?thread_ts=1686858096.401379&cid=C01LCTT7JNN). This currently cannot be an input because it's not an output for the `run_nextclade` rule.

tsibley:
I've pushed a commit to address the use of params and touch files instead of outputs, as well as two smaller separate commits. These are only tested in dry run mode, so may need minor adjustment. Please feel free to rework them as needed.

workflow/snakemake_rules/nextclade.smk
```diff
@@ -179,18 +180,19 @@ rule run_nextclade:
...
-        translations=lambda w: [
-            f"data/{database}/nextclade{w.reference}.translation_{gene}.upd.fasta"
-            for gene in GENE_LIST
-        ],
     output:
         info=f"data/{database}/nextclade{{reference}}_new_raw.tsv",
-        alignment=f"data/{database}/nextclade{{reference}}.aligned.upd.fasta",
-        touchfile=touch(f"data/{database}/nextclade{{reference}}.touch"),
+        alignment=temp(f"data/{database}/nextclade{{reference}}.aligned.upd.fasta"),
+        translations=[temp(f"data/{database}/nextclade{{reference}}.translation_{gene}.upd.fasta") for gene in GENE_LIST],
```
tsibley, line 188: There may be an issue here during a non-dry run because the builds where `reference == "_21L"` won't pass `--output-translations` and so won't produce these output files. I think Snakemake is likely to notice and error, but I don't know it's exact behaviour off hand. The simplest solution, since we can't condition the outputs on a wildcard, may be to remove the condition on `--output-translations` and always produce them, even if they're immediately removed because no downstream rules use them. Alternative solutions include producing them instead in the `shell:` block with `touch` when necessary for `reference != ""`.

workflow/snakemake_rules/nextclade.smk
```diff
@@ -179,18 +180,19 @@ rule run_nextclade:
     params:
         genes=GENES_SPACE_DELIMITED,
         translation_arg=lambda w: (
-            f"--output-translations=data/{database}/nextclade{w.reference}.translation_{{gene}}.upd.fasta"
+            # Nextclade takes a filename template in which it replaces {gene}
+            # itself, so we want to pass thru {gene} literally to it and make
+            # sure it isn't interpretted by the shell as a glob.  We shellquote
+            # here instead of with :q below because we don't want to pass an
+            # empty string argument when this param is empty.
+            shellquote(f"--output-translations=data/{database}/nextclade{w.reference}.translation_{{gene}}.upd.fasta")
```
tsibley, line 181: A minor imprecision in my comment: `{...}` is a shell expansion (specifically brace expansion) not a shell glob.

Shell quoting may not have been an issue in practice, as the brace expansion in Bash only applies when there's a sequence, e.g. `{a,b}` and not a single item `{a}`.

tsibley:
@corneliusroemer Not sure what I pushed is correct as-is, but it's something to chew on at least if you want. Please feel free to adjust as you see fit. Perhaps I should have pushed to another branch, but I thought it'd be easiest to discuss here in context.

workflow/snakemake_rules/nextclade.smk
joverlee521: > even if they're immediately removed because no downstream rules use them

Oh! I didn't realize that Snakemake removes temp files immediately if no downstream rules use them!

workflow/snakemake_rules/nextclade.smk
corneliusroemer: Yes that's right, we only use the tsv from the 21L build. I'm not sure that this is causing complexity beyond having an extra wildcard `{reference}` but I'll see whether splitting it up into main Nextclade and 21L Nextclade is easier to read.

workflow/snakemake_rules/nextclade.smk
corneliusroemer: Ok now I see how differentiating between main and 21L as suggested by @joverlee521 above makes things simpler :)

workflow/snakemake_rules/nextclade.smk
corneliusroemer: Slightly off topic as Nextclade related but worth having the context here

Pinging @ivan-aksamentov as he devised the template format (I think) :)

We haven't had issues with this yet (I think), but given we heavily use snakemake with nextclade, we may want to template things slightly differently in the future? @tsibley do you have a suggestion for how else we could specify the template string to be less of a pain with snakemake?

corneliusroemer:
Doing a test run after I implemented the suggestions from review from @tsibley and @joverlee521 
```
nextstrain build --aws-batch --attach --no-download .
```

Translations should be available via:
```
aws s3 cp s3://nextstrain-staging/files/ncov/open/branch/publish-translations/translation_E.fasta.zst .
```
in around 12 hours.

corneliusroemer:
Test runs worked out well. I'll merge this in 24hr unless there are objections @joverlee521 @tsibley

Snakefile
joverlee521, line 1: One final question I have is should the new translation outputs be added to `all_targets` in the main Snakefile? Currently these translation outputs will only be generated if they are being uploaded to S3 or if they are specified as Snakemake targets.

corneliusroemer: Good point, I wasn't aware of `all_targets` being a thing. Any downside to putting it there? Happy for you to just make a decision and merge.

joverlee521: > Any downside to putting it there?

Not really. It might be considered a waste of time and resources to produce these files if they are not needed as outputs? `all_targets` are just the default outputs of pipeline, so I guess the question is are these translation outputs considered default outputs of the ncov-ingest workflow or should they be treated as additional targets that we only want for the Nextstrain automated runs? 

Since covSpectrum is the only user of these files from S3, I'm fine with just treating them as additional targets for the automated workflow. If someone _really_ wants them in a manual run, it's not hard to add them as additional targets.

corneliusroemer: Do we ever do manual runs? I thought the whole purpose of ingest was to produce files that get uploaded to s3 😄 I'll merge as is given that these files are just produced for @chaoran-chen

joverlee521: > Do we ever do manual runs? I thought the whole purpose of ingest was to produce files that get uploaded to s3

We don't do manual runs but trying to be conscious of the external users of the ncov-ingest who run the workflow without the S3 interactions.

corneliusroemer: If they want translations they can specify them under targets 🙃 If there are external users they should say hi so we know who they are. Nothing breaks the way it is. In fact, we may _not_ want to advertise these translations as this is purely a feature for Chaoran and we may want to discontinue in the future.

corneliusroemer:
@chaoran-chen I'm sorry, I forgot to trigger a full rerun on Nextclade, so the translations are not complete yet. They should be in ~24hr though. Let me know if the aren't. Sorry for the oversight!

corneliusroemer:
Translations should be there now and complete!

```
❯ aws s3 cp s3://nextstrain-data/files/ncov/open/translation_E.fasta.zst .
download: s3://nextstrain-data/files/ncov/open/translation_E.fasta.zst to ./translation_E.fasta.zst

~/Downloads/tmp via 🐍 v3.11.4 via 🅒 py11 took 16s 
❯ seqkit seq -n translation_E.fasta.zst| wc -l
 8114018
```

PR #405, opened 2023-06-20
Fix: add clade_legacy back via mapping

corneliusroemer:

This PR fixes ingest to work with the latest Nextclade dataset release: at the moment it fails due to `clade_legacy` no longer being output by Nextclade starting with dataset release `2023-06-16`.

(This is due to https://github.com/neherlab/nextclade_data_workflows/pull/42 which in turn was triggered by a refactor in ncov of how we annotate clades https://github.com/nextstrain/ncov/pull/1065.)

So as not to break downstream workflows that rely on ingest output `metadata.tsv` having `clade_legacy`, this PR adds a `clade_legacy` column to `metadata.tsv`

The values are defined as a simple mapping from `clade_nextstrain` (year-letter, e.g. 22F) to `clade_legacy` in `defaults/clade-legacy-mapping.yml`

This file lives in ingest for now to make this PR work without requiring changes to `ncov`.

### Testing

- [x] Locally with `snakemake -c all --configfile config/debug_sample_gisaid.yaml  -p --ri -F`
- [x] Full test runs succeeded (at least `open`, `gisaid` was still running)
  - AWS run: https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/d4606714-a821-4373-9cd7-8fa1a23a81ee
  - Output files downloadable from: `s3://nextstrain-staging/files/ncov/open/branch/cornelius/add-legacy-clade-names`

### Future work

- [ ] Add a reasonable default for `clade_legacy` so that new clade doesn't require changes in ingest. E.g., we could use the `WHO` column, or simply add ` (Omicron)`, or add nothing at all, i.e. map to itself. Tracked in https://github.com/nextstrain/ncov-ingest/issues/406

### DIFF
--- a/bin/join-metadata-and-clades
+++ b/bin/join-metadata-and-clades
@@ -4,6 +4,7 @@ import sys
 from datetime import datetime
 import pandas as pd
 import numpy as np
+import yaml
 
 INSERT_BEFORE_THIS_COLUMN = "pango_lineage"
 COLUMN_TO_REORDER = "Nextstrain_clade"
@@ -15,7 +16,6 @@ rate_per_day = 0.0007 * 29903 / 365
 reference_day = datetime(2020,1,1).toordinal()
 
 column_map = {
-    "clade_legacy": "Nextstrain_clade",
     "clade_nextstrain": "clade_nextstrain",
     "clade_who": "clade_who",
     "Nextclade_pango": "Nextclade_pango",
@@ -43,13 +43,16 @@ column_map = {
     "aaSubstitutions": "aaSubstitutions"
 }
 
+# Nextstrain_clade is added later based on clade_nextstrain and a yml mapping
+new_columns = list(column_map.values()) + ["Nextstrain_clade"]
+
 clades_21L_columns = {"immune_escape","ace2_binding"}
 
 def reorder_columns(result: pd.DataFrame):
     """
     Moves COLUMN_TO_REORDER right before INSERT_BEFORE_THIS_COLUMN
     """
-    if COLUMN_TO_REORDER not in list(column_map.values()):
+    if COLUMN_TO_REORDER not in new_columns:
         raise ValueError(f"Column {COLUMN_TO_REORDER} not found in values of column_map {column_map}")
     columns = list(result.columns)
     if INSERT_BEFORE_THIS_COLUMN not in columns:
@@ -62,11 +65,12 @@ def reorder_columns(result: pd.DataFrame):
 
 def parse_args():
     parser = argparse.ArgumentParser(
-        description="Joins metadata file with Nextclade clade output",
+        description="Joins metadata file with Nextclade tsvs. Adds clade_legacy column.",
     )
-    parser.add_argument("first_file")
-    parser.add_argument("second_file")
-    parser.add_argument("nextclade_21L")
+    parser.add_argument("--metadata", required=True)
+    parser.add_argument("--nextclade-tsv", required=True)
+    parser.add_argument("--nextclade-21L-tsv", required=True)
+    parser.add_argument("--clade-legacy-mapping", required=True)
     parser.add_argument("-o", default=sys.stdout)
     return parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


bin/join-metadata-and-clades
```diff
@@ -119,7 +127,7 @@ def main():
         right_index=True,
         how='left'
     )
-
+
```
corneliusroemer, line 130:
```suggestion

```

bin/join-metadata-and-clades
```diff
@@ -106,11 +110,15 @@ def main():
...
+    # Use yml mapping
+    clade_legacy_mapping= yaml.safe_load(open(args.clade_legacy_mapping))
+    clades["Nextstrain_clade"] = clades["clade_nextstrain"].map(clade_legacy_mapping)
+
     # Remove immune_escape and ace2_binding when clade <21L and not recombinant
     clades.loc[clades.Nextstrain_clade < "21L",["immune_escape","ace2_binding"]] = float('nan')
 
-
-    clades = clades[list(column_map.values())]
+    clades = clades[new_columns]
```
corneliusroemer, line 121:
```suggestion

    clades = clades[new_columns]
```

bin/join-metadata-and-clades
```diff
@@ -106,11 +110,16 @@ def main():
 
     clades = pd.merge(clades, clades_21L, left_index=True, right_index=True, how='left')
 
+    # Add clade_legacy column as Nextstrain_clade
+    # Use yml mapping
+    clade_legacy_mapping= yaml.safe_load(open(args.clade_legacy_mapping))
```
joverlee521, line 115:
```suggestion
    with open(args.clade_legacy_mapping, 'r') as legacy_mapping_file:
         clade_legacy_mapping = yaml.safe_load(legacy_mapping_file)
```

joverlee521 (approved):
Thanks for fixing this @corneliusroemer! I'm getting more and more confused by the circular data dependencies of ncov-ingest, ncov, and Nextclade 😥 

I left some comments, but should be okay to merge with a full re-run of Nextclade.

emmahodcroft:
@joverlee521 Every time I step away from this issue I have to go back and try to remember it all again (which I haven't done this morning) but my memory is that hopefully these steps are unwrapping these circular dependencies...

Nextclade now no longer knows about the legacy clade names from Nextstrain & thus doesn't output them. But due to that, we have to add them back into Nextstrain metadata to support legacy use (at least until such time as we decide to stop supporting that, with some warning or etc). This should mean that future changes to Nextclade should be more independent - and same on Nextstrain end as far as how we want to rename clades or map them to various labels, etc....

That's the plan anyway, but I'm a bit lost on what step we're on ATM TBH 🤷 !

bin/join-metadata-and-clades
```diff
@@ -43,13 +43,16 @@ column_map = {
     "aaSubstitutions": "aaSubstitutions"
 }
 
+# Nextstrain_clade is added later based on clade_nextstrain and a yml mapping
+new_columns = list(column_map.values()) + ["Nextstrain_clade"]
```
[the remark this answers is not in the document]
corneliusroemer: That's a great point. I _think_ it _doesn't_ change order, as "Nextstrain_clade" is the column that's reordered later in the script. See:
- https://github.com/nextstrain/ncov-ingest/blob/93fb79c51ecccfce4bd750c22a5f0ff93ab25857/bin/join-metadata-and-clades#LL152C5-L152C37
- https://github.com/nextstrain/ncov-ingest/blob/93fb79c51ecccfce4bd750c22a5f0ff93ab25857/bin/join-metadata-and-clades#L60
- https://github.com/nextstrain/ncov-ingest/blob/93fb79c51ecccfce4bd750c22a5f0ff93ab25857/bin/join-metadata-and-clades#L10

bin/join-metadata-and-clades
```diff
@@ -106,11 +110,16 @@ def main():
 
     clades = pd.merge(clades, clades_21L, left_index=True, right_index=True, how='left')
 
+    # Add clade_legacy column as Nextstrain_clade
+    # Use yml mapping
+    clade_legacy_mapping= yaml.safe_load(open(args.clade_legacy_mapping))
+    clades["Nextstrain_clade"] = clades["clade_nextstrain"].map(clade_legacy_mapping)
```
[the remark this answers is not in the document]
corneliusroemer: Great point that I wanted to address later, after getting this fix out.

I think it would be good to not require us to update the `clade-legacy-mapping.yml` for each new clade we define. Since new clades will be part of Omicron, I suggest I make the default mapping to add ` (Omicron)` as everything from now on will almost certainly be Omicron (if it isn't it's easy to overwrite by adding a line to the mapping).

Since the whole reason for this column is backwards compatibility, this won't break anything and it is also not wrong.

I'll suggest that change in a separate PR so we can merge the fix first.

corneliusroemer:
@emmahodcroft your summary is entirely correct. Legacy definitions are now an ingest-only concern for backwards compatibility.

corneliusroemer:
- [ ] I have added touchfiles for `open/gisaid` x `wuhan/21L`

PR #412, opened 2023-07-31
Use nextstrain/ingest

victorlin:
### Description of proposed changes

Begin usage of centralized ingest scripts and add details on how to pull new updates from the central repo.

### Testing

- [x] Checks pass (Update-image failure is [unrelated](https://github.com/nextstrain/ncov-ingest/issues/413) and can be ignored. It was triggered by a non-functional change in update-image.yml).
- [x] [Run GenBank fetch and ingest](https://github.com/nextstrain/ncov-ingest/actions/runs/5933678225) on PR branch, verify successful run.

    <details>
    <summary>Evidence</summary>

    I checked for the presence of `✅ This pipeline has successfully finished 🎉` in the AWS Batch logs.

    I also spot checked some lines in the generated metadata file:

    ```sh
    # Download and decompress metadata files from latest rebuild and PR rebuild.
    aws s3api get-object --bucket nextstrain-data --key files/ncov/open/metadata.tsv.zst metadata-current.tsv.zst --version-id 3wF7ccfqiJ73cw7dVjq.mlrPf5A6Vah6
    unzstd metadata-current.tsv.zst
    aws s3 cp s3://nextstrain-staging/files/ncov/open/branch/victorlin/centralized-ingest-git-subrepo/metadata.tsv.zst metadata-branch.tsv.zst
    unzstd metadata-branch.tsv.zst

    # There is no difference in the first and last 10k lines.
    # Other differences can be assumed to be due data availability differences based on time of run.
    diff <(head -n 10000 metadata-current.tsv) <(head -n 10000 metadata-branch.tsv)
    diff <(tail -n 10000 metadata-current.tsv) <(tail -n 10000 metadata-branch.tsv)
    ```

    </details>

- [x] [Run GISAID fetch and ingest](https://github.com/nextstrain/ncov-ingest/actions/runs/5933681450) on PR branch, verify successful run.

    <details>
    <summary>Evidence</summary>

    Since I already checked an output file in the GenBank run, here I just checked for the presence of `✅ This pipeline has successfully finished 🎉` in the AWS Batch logs.

    </details>

- [ ] Post-merge: Slack notifications still work as intended.

PR #418, opened 2023-09-07
Add build dir for notify-on-job-start

joverlee521:
ncov-ingest is the only repo where the ingest pipeline is in a separate repository than the phylogenetic workflow. Therefore, the build dir that should be sent with the onstart Slack message is the current directory instead of the default "ingest" directory.

Follow-up on #412 

## Checklist

- [x] Checks pass

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -92,7 +92,7 @@ onstart:
         print(f"\t${{{var}}}: " + ("YES" if os.environ.get(var, "") else "NO") + f"({description})")
     if send_notifications:
         message="🥗 GISAID ingest" if database=="gisaid" else "🥣 GenBank ingest"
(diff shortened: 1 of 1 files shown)


joverlee521:
Merging pre-review since I'm confident in the minor change. Will monitor tomorrow's automated runs.

joverlee521:
Confirmed that both [GISAID](https://github.com/nextstrain/ncov-ingest/actions/runs/6125055457) and [GenBank](https://github.com/nextstrain/ncov-ingest/actions/runs/6125039728) runs triggered as expected with the correct attach messages in Slack ([1](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1694197073199489), [2](https://bedfordlab.slack.com/archives/C0159227X7Y/p1694196950577069))

PR #422, opened 2024-01-16
Download nextclade 2.14.0 binary instead of latest

corneliusroemer:
Next step is to make ingest work with nextclade3 which should be easy

But fixing the version is safer for now


corneliusroemer:
Merging now to prevent an error tomorrow when we release v3

We have a month to update to v3 to keep getting the latest dataset

PR #424, opened 2024-01-26
Fix upload diff race

### LINKED ISSUE #423: Race condition between diff and uploads to S3
Similar to the race condition noted in https://github.com/nextstrain/mpox/commit/ad3481a479e5b795f9ce0b8f9ecb422256e903d5

This has resulted in the lack of notifications for the `additional-info-changes.txt`. Without these notifications, we have not been alerted to update the [gisaid_geoLocationRules.tsv](https://github.com/nextstrain/ncov-ingest/blob/master/source-data/gisaid_geoLocationRules.tsv). 
 
This issue was prompted by @trvrb flagging [messy location names for Washington state internally](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1706053243079609?thread_ts=1706047888.568289&cid=C7SDVPBLZ).

joverlee521:

Include the notification's touch file (`data/{databas}/notify.done`) as an input to the `upload_single` rule to ensure that uploads only run after the notifications rules have run.

This prevents the race condition between diffs and upload of files. This will slightly delay the uploads of files but is necessary to support the diffs of local files and files on S3.

Resolves #423 

## Checklist

- [ ] Checks pass

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -41,10 +41,7 @@ if config.get("s3_dst"):
 # and the `s3_src` is provided in config since some notify scripts depend
 # do diffs with files on S3 from previous runs
 if send_notifications and config.get("s3_src"):
-    all_targets.extend([
-        f"data/{database}/notify-on-record-change.done",
-        f"data/{database}/notify.done"
-    ])
+    all_targets.append(f"data/{database}/notify.done")
 
 rule all:
     input: all_targets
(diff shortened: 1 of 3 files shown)


joverlee521:
Comparing the [current DAG](https://github.com/nextstrain/ncov-ingest/assets/40774762/47a95fea-ef3c-4f29-bc9f-264bfb00a6ed) and the [new DAG](https://github.com/nextstrain/ncov-ingest/assets/40774762/33998e61-ad1d-41f8-93af-0a028e49aa8b) with changes from this PR: now `notify_gisaid` _must_ run before any of the `upload_single` rules.

joverlee521:
I plan to merge on Monday, 2024-01-29 when there's more likely to be new data and I can monitor ingest runs.

PR #440, opened 2024-02-28
Overhaul GH Action workflows

joverlee521:

Overhaul the GH Action workflows with the main goal to use the shared `pathogen-repo-build` workflow. 
This allows the GH Action workflow success/failure status to match the AWS Batch job status and will allow us to monitor the success/failure of automated workflows via the [Pathogen workflow status page](https://nextstrain.github.io/status/pathogen-workflows.html).

I made other improvements to use workflow inputs in order to reduce the 8 different GH Action workflows down to 2 workflows. See commits for details.

## Related issue(s)

Related to https://github.com/nextstrain/.github/issues/46

## Checklist

- [x] Checks pass
- [x] Test [GISAID build](https://github.com/nextstrain/ncov-ingest/actions/runs/8086935526)
- [x] Test [GenBank build](https://github.com/nextstrain/ncov-ingest/actions/runs/8102509411)
- [ ] Post-merge: Remove the `AWS_DEFAULT_REGION` repo secret


joverlee521:
The [test GISAID build](https://github.com/nextstrain/ncov-ingest/actions/runs/8086935526) completed successfully:
- using the trial name and uploaded files to to `s3://nextstrain-ncov-private/trial/overhaul-gh-action-workflows/`
- using the custom image `nextstrain/ncov-ingest:branch-fix-image`
- using the `TEST_SLACK_CHANNEL` variable and sent messages to #scratch  

It still fetched from the upstream database when the `fetch_from_database` input was `false`, but I believe this should be fixed with 4d5b6113dc2b6a7c00d340c482b0b992dd1e99eb. 

---

The [test GenBank build](https://github.com/nextstrain/ncov-ingest/actions/runs/8086932212) failed during upload due to permission issues because I forgot the `NextstrainNcovIngestUploader` policies allow the "files/ncov/open/" prefix (not "files/**workflows**/ncov/open/"). 

I'll update the trial prefix to "files/ncov/open/trial/<TRIAL_NAME>/" in order to match the production prefix. I don't think we will ever want to update the ncov production URLs since they are [advertised in our SARS-CoV-2 docs](https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html#remote-inputs-open-files).

joverlee521:
Updated GenBank trial prefix in [force-push](https://github.com/nextstrain/ncov-ingest/compare/4d5b6113dc2b6a7c00d340c482b0b992dd1e99eb..9ccf797b0327aae48301df88336113da37367c2f)

Running a new [GenBank trial build](https://github.com/nextstrain/ncov-ingest/actions/runs/8102509411).

joverlee521:
[GenBank trial build](https://github.com/nextstrain/ncov-ingest/actions/runs/8102509411) completed successfully and uploaded outputs to `s3://nextstrain-data/files/ncov/open/trial/overhaul-gh-action-workflows/`. 

I'm doing one [final trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/8115557847) today to test the `fetch_from_database` input which should be fixed by 4c15c84539157db5f09366033050cacfdcc118fc. If all goes well, I'll merge this on Monday so that I can monitor the automated runs.

joverlee521:
Okay, actual [final trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/8115818792) with final fixes to `fetch_from_database` in b926bd15dedc13e59d6bd36d9a71b5823d27858a 🤞

.github/workflows/fetch-and-ingest-genbank-master.yml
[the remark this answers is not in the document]
joverlee521: Yup, I didn't want to complicate the workflow even more by adding checks for `github.event.client_payload.<input_name>` if it's not being used!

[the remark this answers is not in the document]
joverlee521: > Did you choose not to do this to allow the [actions page](https://github.com/nextstrain/ncov-ingest/actions) to easily distinguish between the two?

Yeah, I think it's nice to be able to easily distinguish between GISAID and GenBank workflows. (The [pathogen status page](https://nextstrain.github.io/status/pathogen-workflows.html) currently groups by [repo/workflow name](https://github.com/nextstrain/status/blob/fba6b23f0dca559cd13d35356e0fd19799f3f8c1/pathogen-workflows.html.js#L29-L30)). Also, the scheduled GitHub Action workflows do not use inputs, so I think the conditionals to trigger scheduled workflows with different default values could get pretty messy. 

> if we are using the actions page to monitor progress for the foreseeable future, is the best view into this to filter by [event:scheduled](https://github.com/nextstrain/ncov-ingest/actions?query=event%3Aschedule) to ignore trial / branch runs?

Yup, that's should be the easiest way to monitor the automated runs for now. I think it'd also be reasonable to add a new [status page](https://nextstrain.github.io/status/) that only filters to the scheduled events.

.github/workflows/fetch-and-ingest-genbank-master.yml
```diff
@@ -31,6 +31,13 @@ on:
 
   # Manually triggered using GitHub's UI
   workflow_dispatch:
+    inputs:
+      trial_name:
```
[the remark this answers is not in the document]
joverlee521: Added description in [force-push](https://github.com/nextstrain/ncov-ingest/compare/b926bd15dedc13e59d6bd36d9a71b5823d27858a..3d92f48ec57fc9f943af38c24c5a456e878a74b7)

joverlee521:
Ah, missed today's automated runs. I'll merge EOD today in case there are other comments.

corneliusroemer:
Thanks for cleaning up the workflows @joverlee521!

I just tried this out and it took a bit of time to understand how to configure the workflow input params to get it to do what I want (run on branch, do a test run, no fetch).

Maybe there's scope for a single "test" workflow that simply does repository dispatch with these params? The new single workflow is great for maintenance but might cause errors when misused due to complexity in selecting the right inputs. A neutered test dispatch might reduce accidental misuse likelihood.

joverlee521:
> Oh wow, I just discovered that the logs are streamed from AWS batch - that's amazing!

This is all thanks to @tsibley's [update to the `pathogen-repo-build` workflow](https://github.com/nextstrain/.github/pull/54) to let it stay attached to the AWS Batch jobs 🙌

joverlee521:
> I just tried this out and it took a bit of time to understand how to configure the workflow input params to get it to do what I want (run on branch, do a test run, no fetch).
> 
> Maybe there's scope for a single "test" workflow that simply does repository dispatch with these params? The new single workflow is great for maintenance but might cause errors when misused due to complexity in selecting the right inputs. A neutered test dispatch might reduce accidental misuse likelihood.

Are there additional docs that I can add to clarify the use of the workflow input params? 

If we wanted to support repository dispatch, we would have to complicate the workflows a bit to check for `github.event.client_payload.<input_name>` params in addition to the `input.<input_name>` checks. I'm not sure that's worth adding right now. I'm not too worried about "misuse" since cancelling the GitHub workflow run will also cancel the attached AWS Batch job now.

.github/workflows/fetch-and-ingest-genbank-master.yml
```diff
@@ -99,6 +113,7 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         PAT_GITHUB_DISPATCH: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
         TRIAL_NAME: ${{ inputs.trial_name }}
+        FETCH_FROM_DATABASE: ${{ inputs.fetch_from_database != '' && inputs.fetch_from_database || true }}
```
tsibley: This, combined with the `type: boolean` of `inputs.fetch_from_database`, is going to result in `FETCH_FROM_DATABASE` being either the strings `true` or `false`, right?

And then the Bash condition (earlier up) is `[[ "$FETCH_FROM_DATABASE" ]]`, which evaluates to true (exits 0) for both `[[ "true" ]]` and `[[ "false" ]]`. Yeah?

So does setting `inputs.fetch_from_database` to `false` actually work? (or am I missing something?)

bin/write-envdir
tsibley, line 1: Love to see this. :-)

.github/workflows/fetch-and-ingest-genbank-master.yml
```diff
@@ -96,22 +100,10 @@ jobs:
...
-        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        PAT_GITHUB_DISPATCH: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
-        TRIAL_NAME: ${{ inputs.trial_name }}
-        FETCH_FROM_DATABASE: ${{ inputs.fetch_from_database != '' && inputs.fetch_from_database || true }}
-
-    - name: notify_pipeline_failed
-      if: ${{ failure() }}
-      run: ./vendored/notify-on-job-fail Ingest nextstrain/ncov-ingest
+            $CONFIG_OVERRIDES
```
tsibley, line 114: FWIW, this opens up a command-line injection via `inputs.trial_name` (though in practice it's probably fine given our usage).

.github/workflows/fetch-and-ingest-genbank-master.yml
```diff
@@ -60,11 +60,11 @@ jobs:
         name: Set config overrides
         env:
           TRIAL_NAME: ${{ inputs.trial_name }}
-          FETCH_FROM_DATABASE: ${{ inputs.fetch_from_database != '' && inputs.fetch_from_database || true }}
+          FETCH_FROM_DATABASE: ${{ github.event_name != 'workflow_dispatch' && true || inputs.fetch_from_database }}
         run: |
           config="--config"
 
-          if [[ "$FETCH_FROM_DATABASE" ]]; then
+          if "$FETCH_FROM_DATABASE"; then
```
tsibley, line 67: This relies on running the programs `true` and `false`... 🫣

.github/workflows/fetch-and-ingest-genbank-master.yml
tsibley: Oh, I see it gets fixed in a later commit, but also https://github.com/nextstrain/ncov-ingest/pull/440#discussion_r1571480694.

.github/workflows/fetch-and-ingest-genbank-master.yml
joverlee521: OH, I've repeated this pattern across other GH Action workflows 🙈 

What would be the safer thing to do here?

.github/workflows/fetch-and-ingest-genbank-master.yml
joverlee521: Thanks for flagging! Updating in https://github.com/nextstrain/ncov-ingest/pull/444

.github/workflows/fetch-and-ingest-genbank-master.yml
tsibley: The best thing is building up an array of strings, which can then be expanded with quoting to prevent subsequent word splitting. But with the config overrides being computed in a separate job, you can't pass an array, only a single long string. One way forward with that is to use shell-quoting semantics inside that single long string (i.e. nested quoting), but that gets tedious and is easy to mess up. Stepping back, why did the config override get moved into a separate job? Does it need to be?

Or, as an alternative to passing around a command-line snippet in a string/env var, we could produce another YAML file and pass that instead?

joverlee521: > Stepping back, why did the config override get moved into a separate job? Does it need to be?

It had to get moved to a separate job because the config override is not currently part of the `pathogen-repo-build` workflow. Since configs are pretty different across pathogen workflows, I didn't want to shoehorn it into the reusable workflow. 

> Or, as an alternative to passing around a command-line snippet in a string/env var, we could produce another YAML file and pass that instead?

Ah, so build a custom config YAML that is uploaded/download as an artifact of the GH Action workflow. I think this is doable with updating `pathogen-repo-build` workflow to take a artifact name as an input and download the artifact as part of the reusable workflow. Yeah, this seems like the way to go if we are unable fully standardize configs across pathogens.

tsibley: > It had to get moved to a separate job because the config override is not currently part of the `pathogen-repo-build` workflow. Since configs are pretty different across pathogen workflows, I didn't want to shoehorn it into the reusable workflow.

Can't it go in the `run` input of the `pathogen-repo-build` workflow call?

joverlee521: Hah, yes 🤦‍♀️

PR #445, opened 2024-06-05
Switch AWS credentials from a long-lived IAM user to short-lived role sessions

tsibley:
I'll also be removing the corresponding repository secrets.  Both that and this commit are required to effect the switch.

Related-to: <https://github.com/nextstrain/private/issues/110>

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -103,8 +103,6 @@ jobs:
           --no-download \
           --cpus 16 \
           --memory 68GiB \
-          --env AWS_ACCESS_KEY_ID \
-          --env AWS_SECRET_ACCESS_KEY \
           --env GITHUB_RUN_ID \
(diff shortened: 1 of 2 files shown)


victorlin:
I don't think this ever went into effect. The repository secrets still exist and were last updated just one week after this was merged.

The latest GenBank fetch and ingest run confirms that a long-lived IAM user is still being used (from [log](https://github.com/nextstrain/ncov-ingest/actions/runs/27070469340/job/79898667615#step:7:32)):

```
Proceeding with IAM user credentials
```

joverlee521:
> I don't think this ever went into effect. The repository secrets still exist and were last updated just one week after this was merged.

AH, I should have followed up here when I added the secret back as part of https://github.com/nextstrain/ncov-ingest/issues/446#issuecomment-2168687617

victorlin:
Ahh I knew I was missing something! In that case I'll leave things as-is for https://github.com/nextstrain/private/issues/246.

Issue #446, opened 2024-06-11
GISAID workflow hitting max_session_duration


joverlee521:
Our automated workflows use short lived AWS credentials for sessions which are limited by the [max_session_duration](https://github.com/nextstrain/infra/blob/814a7e4cd635b40c12fb0c4e83b1680080fa802b/env/production/aws-iam-role-GitHubActionsRoleNextstrainRepo%40.tf#L8) of 12 hours, which is the [maximum allowed by AWS](https://docs.aws.amazon.com/singlesignon/latest/userguide/howtosessionduration.html). 

The GISAID workflow this max yesterday and ran into [errors](https://github.com/nextstrain/ncov-ingest/actions/runs/9453424605/job/26059500954#step:5:2661):

```
upload failed: - to s3://nextstrain-ncov-private/gisaid.ndjson.xz An error occurred (ExpiredToken) when calling the UploadPart operation: The provided token has expired.
```

joverlee521:
It may be time to revisit https://github.com/nextstrain/ncov-ingest/issues/240

joverlee521:
My comment from related thread on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1718122450933829):

> In the absence of benchmark files, I'm just scanning the Snakemake log files from the workflows for some general timings:
> ~1.5h - downloading data from GISAID/S3
> ~4h - transform-gisiad
> ~1h - filter fasta for new sequences to run through Nextclade
> ~0.5h - joining metadata + nextclade
> = ~7h of data munging - this is about the same with/without new data
> 
> The rest of the workflow is just uploading files to S3!!
> 
> Without new data, it still takes ~2h to generate the hash for sequences.fasta for checking against S3.
> With new data, it takes ~4h to upload sequences.fasta to S3.

joverlee521:
Ah, this is also not considering the full run that gets triggered when there's a new Nextclade dataset released.

The last [full run on Apri 16, 2024](https://github.com/nextstrain/ncov-ingest/actions/runs/8710477088) ran for ~15h.

joverlee521:
Bumping the CPUs in #447 decreased the build time by about ~1h, which came from parallelizing the download of data at the beginning of the workflow. 

We will still run over the 12h limit for full Nextclade runs, so I'm going to work on https://github.com/nextstrain/ingest/issues/41

corneliusroemer:
Thanks @joverlee521 for the summary!

[...]

PR #447, opened 2024-06-11
Bump CPUs for fetch-and-ingest workflows

joverlee521:
I've been only bumping the memory but not the CPUs for the fetch-and-ingest workflows. Might as well use all the compute that we are paying for. GenBank should be using c5.9xlarge and GISAID should be using c5.12xlarge, so bumping CPUs to match the instances.¹

Maybe this will magically help https://github.com/nextstrain/ncov-ingest/issues/446?

¹ <https://aws.amazon.com/ec2/instance-types/c5/>

## Checklist

- [ ] Checks pass
- [x] [GISAID workflow](https://github.com/nextstrain/ncov-ingest/actions/runs/9474169604)

### DIFF
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -101,7 +101,7 @@ jobs:
           --aws-batch \
           --detach \
           --no-download \
-          --cpus 16 \
+          --cpus 36 \
           --memory 68GiB \
           --env GITHUB_RUN_ID \
           --env SLACK_TOKEN \
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -100,7 +100,7 @@ jobs:
           --aws-batch \
           --detach \
           --no-download \
-          --cpus 16 \
+          --cpus 48 \
           --memory 90GiB \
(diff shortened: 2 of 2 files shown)


joverlee521:
[Last nights run](https://github.com/nextstrain/ncov-ingest/actions/runs/9474169604) finished under 12h 🎉 
I'm going to dig into the logs a little bit, but at least this is an improvement so I'm merging this ahead of today's run.

corneliusroemer:
We could probably scrap those CPU limits altogether. All they do is make snakemake restrict the number of jobs run in parallel.

This has obvious downsides and only rare benefits.

The CPU scheduler figures out ways to give all jobs some share when oversubscribed. We don't really need snakemake to do pessimistic scheduling.

joverlee521:
> We could probably scrap those CPU limits altogether. 

For the aws batch runtime, the `--cpus` option is used to override the default [nextstrain-job definition](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#job-definition/ec2/detail/arn:aws:batch:us-east-1:827581582529:job-definition/nextstrain-job:25), which is only 4 cpus.

tsibley:
> The CPU scheduler figures out ways to give all jobs some share when oversubscribed.

Yes, progress will still be made with oversubscription, but the run time of the whole workflow will increase, sometimes substantially, depending on the kind of workload. It's still better not to oversubscribe when you can avoid it.

Issue #459, opened 2024-07-09
Define threads for Nextclade rules


joverlee521:
Prompted by https://github.com/nextstrain/ncov-ingest/issues/456#issuecomment-2218083580

From my reading of [Snakemake docs on threads](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#threads):

> Also, setting a threads: maximum is required to achieve parallelism in tools that (often implicitly and without the user knowing) rely on an environment variable for the maximum of cores to use. 

Makes me think we _need_ to define threads of the Nextclade Snakemake rules in order for Nextclade to parallelize. 

Proposing that we split the workflow cores between the two Nextclade rules:

```diff
diff --git a/workflow/snakemake_rules/nextclade.smk b/workflow/snakemake_rules/nextclade.smk
index 8d24c23..e97c8aa 100644
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -193,6 +193,8 @@ rule run_wuhan_nextclade:
             temp(f"data/{database}/nextclade.translation_{gene}.upd.fasta")
             for gene in GENE_LIST
         ],
+    threads:
+        workflow.cores * 0.5
     benchmark:
         f"benchmarks/run_wuhan_nextclade_{database}.txt"
     shell:
@@ -224,6 +226,8 @@ rule run_21L_nextclade:
         sequences=f"data/{database}/nextclade_21L.sequences.fasta",
     output:
         info=f"data/{database}/nextclade_21L_new_raw.tsv",
+    threads:
+        workflow.cores * 0.5
     benchmark:
         f"benchmarks/run_21L_nextclade_{database}.txt"
     shell:
```

corneliusroemer:
Good idea but I don't think we need to set these threads. All they do, in my experience, is _limit_ parallelism.

Have you checked that there's a need for this? I'm pretty sure things have always been running in parallel just fine.

Nextclade uses all cores by default, unless you tell it through `-j` otherwise. It doesn't read env variables AFAIK

corneliusroemer:
I still don't think this is necessary. I've never used nproc and it seems to report something that nextclade doesn't care about - at least on macOS.

See how these three rules run all in parallel, even though one doesn't mention threads at all.

They all run the same time, using as many jobs as there are cores (unless limited by `-j`). So the threads variable has no effect other than _limiting_ parallelness.

[...]

PR #460, opened 2024-07-09
Add threads to Nextclade runs

joverlee521:
Resolves https://github.com/nextstrain/ncov-ingest/issues/459

## Checklist

- [x] Checks pass


joverlee521:
Merging ahead of today's automated runs to see if it will speed things up.

corneliusroemer:
Hmm, did you see my comments? I don't think we should set those threads.

PR #461, opened 2024-07-10
Pass threads to Nextclade

joverlee521:

Follow up to https://github.com/nextstrain/ncov-ingest/pull/460
Related to #459 

## Checklist

- [ ] Checks pass


joverlee521:
As stated in https://github.com/nextstrain/ncov-ingest/issues/459#issuecomment-2218918061, merging this ahead of today's automated run to see how explicitly passing threads to Nextclade will affect runtimes. 

We will only see the effects when there are new sequences to run through Nextclade so I'll most likely circle back to this ~next week.

Issue #462, opened 2024-07-12
Add Snakemake retries to trigger jobs


joverlee521:
## Context

Prompted by the workflow errors from the automated runs on 2024-07-11 ([GISAID](https://github.com/nextstrain/ncov-ingest/actions/runs/9896410508), [GenBank](https://github.com/nextstrain/ncov-ingest/actions/runs/9896415179)). 

Both failed during trigger of downstream forecasts-ncov GH Action workflows:

```
[batch] [2024-07-12T02:30:27+00:00] rule trigger_counts_pipeline:
[batch] [2024-07-12T02:30:27+00:00]     input: data/genbank/upload.done
[batch] [2024-07-12T02:30:27+00:00]     output: data/genbank/trigger-counts.done
[batch] [2024-07-12T02:30:27+00:00]     jobid: 96
[batch] [2024-07-12T02:30:27+00:00]     benchmark: benchmarks/trigger_counts_pipeline_genbank.txt
[batch] [2024-07-12T02:30:27+00:00]     reason: Missing output files: data/genbank/trigger-counts.done; Input files updated by another job: data/genbank/upload.done
[batch] [2024-07-12T02:30:27+00:00]     resources: tmpdir=/tmp
[batch] [2024-07-12T02:30:27+00:00]         ./vendored/trigger nextstrain/forecasts-ncov genbank/clade-counts
[batch] [2024-07-12T02:30:27+00:00]         
[batch] [2024-07-12T02:30:27+00:00] curl: (22) The requested URL returned error: 404 
[batch] [2024-07-12T02:30:27+00:00] Request failed
```

The workflows were able to trigger the downstream ncov builds just fine, so I don't think there was anything wrong with the GitHub token permissions. The last update to the forecasts-ncov GH Action workflows was [a month ago](https://github.com/nextstrain/forecasts-ncov/commit/0a7ab89ec0cc71cb759c66e392823c8612e81fb2), so I suspect this was a transient error. 

We should add Snakemake `retries` to the trigger rules so that the workflow can automatically retry in case of transient errors like this.

PR #463, opened 2024-07-12
Add retries to trigger rules

joverlee521:

Add retires to trigger rules so that the workflow can automatically
retry them if it encounters transient network errors.

Picked 5 retries to match what is set for the various fetch rules.

## Related issue(s)

Resolves https://github.com/nextstrain/ncov-ingest/issues/462

## Checklist

- [x] Checks pass


joverlee521:
I plan to merge this on Monday.

PR #466, opened 2024-07-24
Ignore cache if Nextclade or dataset version is different

joverlee521:
## Description of proposed changes

Update workflow to ignore the Nextclade cache if the current Nextclade version or the Nextclade dataset version is different than the version in the cache. 

Currently checks Nextclade and dataset versions of the first row of the nextclade.tsv file and formats them as the proposed JSON in #458. Once the version JSON file is in place, it should be easy to swap out the check for the new file.

## Related issue(s)

Resolves #457 

## Checklist

- [x] Checks pass

PR #467, opened 2024-07-27
Surface Nextclade versions

### LINKED ISSUE #458: Surface Nextclade version + Nextclade dataset version in final metadata output
Prompted by discussion in [blab/forecasting project](https://docs.google.com/document/d/1uioKPOF1U34EiheboX6UVe6pCnfRkdZouUXjgQLLLa4/edit#heading=h.b6uq0bfygq0t)

Naively, we could include the Nextclade version and Nextclade dataset version in the [join-metadata-and-clades](https://github.com/nextstrain/ncov-ingest/blob/master/bin/join-metadata-and-clades). 

However, if https://github.com/nextstrain/ncov-ingest/issues/457 is implemented, then the metadata comes from a single Nextclade version/dataset version. Then these versions should be surfaced through the file name or file metadata.

joverlee521:

Creates one version JSON for each Nextclade TSV and one version JSON for the metadata TSV. Since the metadata just uses the Nextclade TSV columns directly, just add the `metadata_tsv_sha256sum` to the SARS-CoV-2 dataset version JSON. If we ever want to track data provenance by column, we will update the schema to include the 21L dataset version.

The two Nextclade version JSONs will be used to check whether the workflow should use the existing cache. The metadata version JSON will be used to surface the version info to downstream users of the data.

## Related issue(s)

Depends on #466 
Resolves #458 

## Checklist

- [x] Checks pass

### DIFF
--- a/workflow/snakemake_rules/upload.smk
+++ b/workflow/snakemake_rules/upload.smk
@@ -33,6 +33,9 @@ def compute_files_to_upload():
                         "aligned.fasta.zst":           f"data/{database}/aligned.fasta",
                         "nextclade_21L.tsv.zst":       f"data/{database}/nextclade_21L.tsv",
 
+                        "nextclade_version.json":          f"data/{database}/nextclade_version.json",
+                        "nextclade_21L_version.json":      f"data/{database}/nextclade_21L_version.json",
+                        "metadata_version.json":           f"data/{database}/metadata_version.json",
                     }
     files_to_upload = files_to_upload | {
         f"translation_{gene}.fasta.zst" : f"data/{database}/translation_{gene}.fasta"
--- a/bin/generate-nextclade-version-json
+++ b/bin/generate-nextclade-version-json
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+vendored="$(dirname "$0")"/../vendored
+
+
+nextclade="${1:?A path to the Nextclade executable is required as the first argument}"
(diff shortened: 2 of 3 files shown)


workflow/snakemake_rules/upload.smk
```diff
@@ -33,6 +33,9 @@ def compute_files_to_upload():
                         "aligned.fasta.zst":           f"data/{database}/aligned.fasta",
                         "nextclade_21L.tsv.zst":       f"data/{database}/nextclade_21L.tsv",
 
+                        "nextclade_version.json":          f"data/{database}/nextclade_version.json",
+                        "nextclade_21L_version.json":      f"data/{database}/nextclade_21L_version.json",
+                        "metadata_version.json":           f"data/{database}/metadata_version.json",
```
joverlee521, line 38: Open to different S3 file names for the version JSONs.

huddlej (approved):
Thank you, @joverlee521! I like this approach of building separate versions JSONs per file and merging the main Nextclade version JSON into the metadata version JSON.

I don't have strong feelings about the names, either. These names seem fine. Some day we'll have `nextclade_25A_version.json` or more than one additional file, but this approach is flexible enough to support that.

Assuming you tested locally, we could merge and see how it works on the next run?

joverlee521:
> Assuming you tested locally, we could merge and see how it works on the next run?

Yup! I'll plan to merge tomorrow morning and monitor the automated runs.

joverlee521:
The public metadata version file is available at https://data.nextstrain.org/files/ncov/open/metadata_version.json

PR #468, opened 2024-07-27
Remove version columns

joverlee521:

Updates the cache version check to use the version JSON added in https://github.com/nextstrain/ncov-ingest/pull/467 and removes the version columns from the Nextclade TSV. 

Removal of the version columns should be considered a breaking change for the Nextclade cache. Once merged, this will need to be run with the renew flags to manually force a full-rerun.

## Checklist

- [ ] Checks pass


joverlee521:
I plan on doing more thorough testing on this next week but wanted to get people's eyes on this to make sure we are okay with removing the extra Nextclade version columns. 

I don't think anyone is using the version columns for downstream work, but wanted to double check.

PR #473, opened 2024-08-22
fix: make legacy mapping use `{clade} (Omicron)` by default

### LINKED ISSUE #471: The `Nextstrain_clade` column does not contain `24C`, but `?` instead.
### Current Behavior

The `Nextstrain_clade` column does not contain `24C`, but `?` instead.

This was noticed by Emma who still uses this legacy column.

@emmahodcroft - I will let you know once this issue is confirmed fixed.

corneliusroemer:
Resolves #471

I forgot to update the clade-legacy-mapping.yml with the last clade update (24C). As a result, the Nextstrain_clade output for 24C was wrong (`?` instead of `24C (Omicron)`).

To reduce maintenance effort, we now use `{value} (Omicron)` as the default mapping. This means one less file to update when clades are updated. If there's ever a non-Omicron again, we'd have to updated `clade-legacy-mapping.yml` but WHO has stopped giving variant names and there's nothing left that's not Omicron. So this should work into the indefinite future (until Emma stops using `Nextstrain_clade` - she's possibly the only user of that backwards-compat column)

## Checklist

- [ ] Checks pass
- [ ] Ping Emma once this is fixed for good in the actual s3 files.


corneliusroemer:
Merging this as I have high confidence this change will work - if not we can revert and/or fix tomorrow

corneliusroemer:
Famous last words - of course there was an issue: I forgot the parens around `(Omicron)`, fixed in https://github.com/nextstrain/ncov-ingest/commit/8276d2206a21540d81595a5671ae0040e6ea3491

corneliusroemer:
Running ingest (without fetch):
- https://github.com/nextstrain/ncov-ingest/actions/runs/10513065908
- https://github.com/nextstrain/ncov-ingest/actions/runs/10513067722

Issue #475, opened 2024-09-12
Automated GenBank ingest failing 


joverlee521:
Automated GenBank ingest failures
- https://github.com/nextstrain/ncov-ingest/actions/runs/10817236034
- https://github.com/nextstrain/ncov-ingest/actions/runs/10836359393

Both workflows failed [during transform-rki-data-to-ndjson](https://github.com/nextstrain/ncov-ingest/actions/runs/10836359393/job/30069988588#step:5:704). This is due to upstream changes in the RKI data. RKI announced in https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland/issues/59 that some variables have been renamed. This particular error comes from `SEQUENCE.ID` being renamed to `igs_id`, but we use other columns as well.

The [available variables](https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland?tab=readme-ov-file#variablen-und-werte) are listed in the top level README

## TODOs
- [x] update `RKI_INDEX_COL` in [bin/transform-rki-data-to-ndjson](https://github.com/nextstrain/ncov-ingest/blob/a17a6932d7053282d07babed05e3379d87ed7dd5/bin/transform-rki-data-to-ndjson#L8)
- [x] update `COLUMN_MAP` in [bin/transform-rki](https://github.com/nextstrain/ncov-ingest/blob/a17a6932d7053282d07babed05e3379d87ed7dd5/bin/transform-rki#L31-L38)
- [x] add new date format `%Y-%m-%dT%H:%M:%S` to [expected `date_formats`](https://github.com/nextstrain/ncov-ingest/blob/a17a6932d7053282d07babed05e3379d87ed7dd5/lib/utils/transformpipeline/transforms.py#L302)
- [x] parse new JSON blob for pango_lineage
- [ ] diff with old version on S3 (s3://nextstrain-data/files/ncov/rki/metadata.tsv.zst)

joverlee521:
RKI announced in https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland/issues/59 that some variables have been renamed. This particular error comes from `SEQUENCE.ID` being renamed to `igs_id`, but we should verify other [columns that are used in transform-rki](https://github.com/nextstrain/ncov-ingest/blob/a17a6932d7053282d07babed05e3379d87ed7dd5/bin/transform-rki#L31-L38)

PR #476, opened 2024-09-13
Update RKI data processing

j23414:

Addressing: https://github.com/nextstrain/ncov-ingest/issues/475

## Related issue(s)

* Related to https://github.com/nextstrain/ncov-ingest/issues/475

## Checklist

- [ ] Checks pass

### DIFF
--- a/lib/utils/transformpipeline/transforms.py
+++ b/lib/utils/transformpipeline/transforms.py
@@ -1,6 +1,7 @@
 import csv
 import re
 import unicodedata
+import json
 from collections import defaultdict
 from typing import Any, Collection, List, MutableMapping, Sequence, Tuple , Dict , Union
 import pandas as pd
@@ -287,6 +288,8 @@ class StandardizeDataRki(Transformer):
     def transform_value(self, entry: dict) -> dict:
         entry['sequence'] = entry['sequence'].replace('\n', '')
         entry['length'] = len(entry['sequence'])
+        lineage_dict = json.loads(entry['pango_lineage'])
+        entry['pango_lineage'] = lineage_dict[0]['lineage']
 
         # Normalize all string data to Unicode Normalization Form C, for
         # consistent, predictable string comparisons.
--- a/lib/utils/transformpipeline/transforms.py
+++ b/lib/utils/transformpipeline/transforms.py
@@ -288,8 +288,13 @@ class StandardizeDataRki(Transformer):
     def transform_value(self, entry: dict) -> dict:
         entry['sequence'] = entry['sequence'].replace('\n', '')
         entry['length'] = len(entry['sequence'])
-        lineage_dict = json.loads(entry['pango_lineage'])
-        entry['pango_lineage'] = lineage_dict[0]['lineage']
+
+        # Pull out latest pango lineage from json blob
+        # Currently this pulls the first entry, but we've added an assert statement to see if there are ever more than one entry
+        # At that time, we can loop over the json blob to find the latest pango lineage assignment
+        lineage_json_blob = json.loads(entry['pango_lineage'])
+        entry['pango_lineage'] = lineage_json_blob[0]['lineage']
+        assert len(lineage_json_blob)==1, f"RKI pango_lineage unexpectedly had more than one entry. rki_accession: {entry['rki_accession']}"
 
         # Normalize all string data to Unicode Normalization Form C, for
         # consistent, predictable string comparisons.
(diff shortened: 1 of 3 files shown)


lib/utils/transformpipeline/transforms.py
```diff
@@ -287,6 +288,8 @@ def __init__(self):
     def transform_value(self, entry: dict) -> dict:
         entry['sequence'] = entry['sequence'].replace('\n', '')
         entry['length'] = len(entry['sequence'])
+        lineage_dict = json.loads(entry['pango_lineage'])
+        entry['pango_lineage'] = lineage_dict[0]['lineage']
```
jameshadfield, line 292: Out of interest, what does `entry['pango_lineage']` look like?

j23414: Sure thing! From [here](https://github.com/robert-koch-institut/SARS-CoV-2-Sequenzdaten_aus_Deutschland?tab=readme-ov-file#variablen-und-werte), the original format looks like

```
[{'method': 'PANGOLIN_LATEST', 'classification_version': 'PUSHER-v1.28.1', 'tool_version': '4.3', 'lineage': 'BA.2', '@qc_notes': 'Ambiguous_content:0.02', '@is_designated': False, '@qc_status': 'pass', '@conflict': 0.0, '@note': 'Usher placements: BA.2(1/1)'}]
```

Or a few examples from the data:

|pango_lineages|
|---|
|[{"classification_version": "PUSHER-v1.29", "lineage": "BA.2", "tool_version": "4.3", "method": "PANGOLIN_LATEST", "@is_designated": false, "@qc_status": "pass", "@conflict": 0.0, "@qc_notes": "Ambiguous_content:0.02", "@note": "Usher placements: BA.2(1/1)"}]|
|[{"classification_version": "PUSHER-v1.29", "lineage": "BA.2.9", "tool_version": "4.3", "method": "PANGOLIN_LATEST", "@is_designated": false, "@qc_status": "pass", "@conflict": 0.0, "@qc_notes": "Ambiguous_content:0.02", "@note": "Usher placements: BA.2.9(1/1)"}]|
|[{"classification_version": "PUSHER-v1.29", "lineage": "BA.2", "tool_version": "4.3", "method": "PANGOLIN_LATEST", "@is_designated": false, "@qc_status": "pass", "@conflict": 0.0, "@qc_notes": "Ambiguous_content:0.02", "@note": "Usher placements: BA.2(2/2)"}]

which, after the change, get corrected to:

|pango_lineages|
|---|
|"BA.2"|
|"BA.2.9"|
|"BA.2"|

jameshadfield: Thanks! My only suggestion would be to do something like
```py
assert len(lineage_dict)==1, f"RKI pango_lineage unexpectedly had more than one entry. Strain: {entry['name']}"
# (not sure if "name" is the correct key to use)
```
(also, perhaps `lineage_dict` isn't the best name if it's a list)

genehack (approved):

lib/utils/transformpipeline/transforms.py
j23414: I might rename `lineage_dict` to "lineage_json_blob" based on their own description (`JSON-Blob`).

Instead of erroring out on multiple entries, I would move to looping through the entries to find the one with `"method": "PANGOLIN_LATEST"`. We're not sure if they're setting this up to historical past pangolin lineage assignments, as lineage systems get updated.

j23414: Heh, in the end, I went with your assert statement to at least flag when multiple entries start showing up (https://github.com/nextstrain/ncov-ingest/pull/476/commits/e57f3c0ff53ba134e54b7dca0d9b759e1c653658). This will help us double-check the purpose of the multiple entries, if they do eventually occur.

I double-checked `entry.keys()`, and the `entry['rki_accession']` is the strain ID.

TheBready: Hi, I just stumbled upon this issue, and we intend always to use PANGOLIN_LATEST for the latest assignment. So, you can filter the JSON array this way. I am sorry for not being more specific in our README. I will change that in our documentation. Please feel free to write us an email if you have questions igs-developers(at)rki.de .

joverlee521: Thanks for chiming in @TheBready! 

Tracking in https://github.com/nextstrain/ncov-ingest/issues/478

Issue #484, opened 2024-12-13
Update links for GISAID and GenBank accessions


joverlee521:
_requested by @trvrb on [Slack](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1734046576334539)_

Link between GISAID and GenBank accessions currently come from https://github.com/nextstrain/ncov-ingest/blob/master/source-data/accessions.tsv.gz which was kindly updated by @AngieHinrichs during the pandemic 🙏 

We haven't picked up the task of updating the file, but the links are now maintained at https://hgwdev.gi.ucsc.edu/~angie/epiToPublicAndDate.latest

## TODOs

- [x] download the public accessions file
- [x] replace the manually curated accessions.tsv.gz with the downloaded file
- [x] remove accessions.tsv.gz

PR #485, opened 2024-12-17
Update links for GISAID and GenBank accessions

joverlee521:

Adds a rule to fetch GISAID and GenBank accession links from UCSC
and uses the output in the `transform_genbank` and `transform_gisaid`
rules.

The UCSC file is modified to keep the headers and format exactly the
same as the current `source-data/accessions.tsv.gz` file so that it can
be directly replaced.

## Related issue(s)

Resolves https://github.com/nextstrain/ncov-ingest/issues/484

## Checklist

- [x] Checks pass
- [x] [GISAID trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/12364597364)
- [x] [GenBank trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/12364600337)
- [x] [2nd GISAID trial](https://github.com/nextstrain/ncov-ingest/actions/runs/12383552360)
- [x] [2nd GenBank trial](https://github.com/nextstrain/ncov-ingest/actions/runs/12383553984)

### DIFF
--- a/bin/fetch-accession-links
+++ b/bin/fetch-accession-links
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+curl "https://hgwdev.gi.ucsc.edu/~angie/epiToPublicAndDate.latest" \
+    --fail --silent --show-error \
+    --header 'User-Agent: https://github.com/nextstrain/ncov-ingest (hello@nextstrain.org)' \
+    | csvtk -t add-header --names gisaid_epi_isl,genbank_accession,strain,date \
+    | csvtk -t cut --fields genbank_accession,gisaid_epi_isl  \
+    | gzip -c
--- a/bin/fetch-accession-links
+++ b/bin/fetch-accession-links
@@ -5,5 +5,4 @@ curl "https://hgwdev.gi.ucsc.edu/~angie/epiToPublicAndDate.latest" \
     --fail --silent --show-error \
     --header 'User-Agent: https://github.com/nextstrain/ncov-ingest (hello@nextstrain.org)' \
     | csvtk -t add-header --names gisaid_epi_isl,genbank_accession,strain,date \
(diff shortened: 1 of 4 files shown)


joverlee521:
Brief look at the trial run output metadata

1. I subset to the `gisaid_epi_isl` and `genbank_accession` columns, only including rows where both are not "?". 
```
zstdcat open-accession-metadata.tsv.zst \
  | csvtk -t cut -f gisaid_epi_isl,genbank_accession \
  | csvtk -t filter2 -f '$genbank_accession != ?' 
  | csvtk -t sort -k genbank_accession > open-accessions.tsv
```
2. I used [daff](https://paulfitz.github.io/daff/) to compare the trial data with the prod data to get a summary of changes:

GISAID

|@@   |count    |
|-----|--------:|
|\+\+\+|2141069  |
|\-\-\-|81421    |
|\-\> |755      |


GenBank
|@@   |count    |
|-----|--------:|
|\+\+\+|1248778  |
|\-\-\-|7964     |
|\-\> |2424     |

The number of removals (`---`) and changes (`->`) makes me think that we might want to keep the original `source-data/accessions.tsv.gz` and just append the UCSC file.

AngieHinrichs:
Glad you're using the latest now and sorry I fell behind with the manual updates!  Feel free to send me examples of any changes/omissions that look concerning.  When the same sequence has been submitted multiple times to GISAID or GenBank (and especially when it has been submitted multiple times to both), there can be some instability from one run to the next regarding which of multiple possible IDs get paired up, which would explain some changes.  Also, over time, some sequences have updates submitted, or are withdrawn from one repo or the other (or both).

trvrb:
I think it makes sense as you say to have the linkage be the best current attempt at linkage and it's okay if this is not fully stable.

joverlee521:
If there's no other feedback, I'll merge this tomorrow morning before the automated run.

genehack (approved):

Issue #489, opened 2025-01-29
GenBank/open failure during `transform_rki_data`


joverlee521:
Latest Genbank/open workflow failed [during `transform_rki_data`](https://github.com/nextstrain/ncov-ingest/actions/runs/13037666962/job/36372057750#step:5:844)

<details>
<summary> Full error </summary>

```
[batch] [2025-01-29T19:07:38+00:00] Traceback (most recent call last):
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/_base.py", line 61, in __next__
[batch] [2025-01-29T19:07:38+00:00]     return self.pipe_component.process(self.data_source_iterator)
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/_base.py", line 95, in process
[batch] [2025-01-29T19:07:38+00:00]     return self.transform_value(next(iterator))
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/transforms.py", line 296, in transform_value
[batch] [2025-01-29T19:07:38+00:00]     entry['pango_lineage'] = lineage_json_blob[0]['lineage']
[batch] [2025-01-29T19:07:38+00:00] IndexError: list index out of range
[batch] [2025-01-29T19:07:38+00:00] During handling of the above exception, another exception occurred:
[batch] [2025-01-29T19:07:38+00:00] Traceback (most recent call last):
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/./bin/transform-rki", line 152, in <module>
[batch] [2025-01-29T19:07:38+00:00]     sorted_metadata = sorted(
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/_base.py", line 61, in __next__
[batch] [2025-01-29T19:07:38+00:00]     return self.pipe_component.process(self.data_source_iterator)
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/_base.py", line 95, in process
[batch] [2025-01-29T19:07:38+00:00]     return self.transform_value(next(iterator))
[batch] [2025-01-29T19:07:38+00:00]   File "/nextstrain/build/lib/utils/transformpipeline/_base.py", line 61, in __next__
[...]

Issue #478, opened 2024-10-03
RKI: filter for `PANGOLIN_LATEST`


joverlee521:
As a follow up to https://github.com/nextstrain/ncov-ingest/pull/476#discussion_r1780889881, we should replace the current [assertion](https://github.com/nextstrain/ncov-ingest/blob/2a5ca550d5105b77e76496f6972ff8650775d11e/lib/utils/transformpipeline/transforms.py#L297) with a filter for the pango lineage labelled as `PANGOLIN_LATEST`.

PR #490, opened 2025-01-30
Fix RKI transform

joverlee521:

Fix and update the RKI transformation for pango_lineage.
- Default to '?' if the RKI data does not include lineages to avoid the IndexError.
- Filter to `PANGOLIN_LATEST` to get the latest lineage assignment.

## Related issue(s)

Resolves https://github.com/nextstrain/ncov-ingest/issues/489
Resolves https://github.com/nextstrain/ncov-ingest/issues/478

## Checklist

- [x] Checks pass
- [x] [Trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/13043702709)

### DIFF
--- a/lib/utils/transformpipeline/transforms.py
+++ b/lib/utils/transformpipeline/transforms.py
@@ -284,6 +284,7 @@ class StandardizeDataRki(Transformer):
 
     def __init__(self):
         self.line_count = 1
+        self.pango_method = "PANGOLIN_LATEST"
 
     def transform_value(self, entry: dict) -> dict:
         entry['sequence'] = entry['sequence'].replace('\n', '')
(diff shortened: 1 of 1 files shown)


joverlee521:
Will merge tomorrow before the automated workflow runs tomorrow if the [trial run](https://github.com/nextstrain/ncov-ingest/actions/runs/13043702709) completes successfully.

jameshadfield:
Nice one! I was going to say it's great that [the assertion we added proved useful](https://github.com/nextstrain/ncov-ingest/pull/476#discussion_r1759268741), but I actually think the error was an `IndexError` from the line immediatly before the assertion?

joverlee521:
> Nice one! I was going to say it's great that [the assertion we added proved useful](https://github.com/nextstrain/ncov-ingest/pull/476#discussion_r1759268741), but I actually think the error was an `IndexError` from the line immediatly before the assertion?

Ah, yeah. If the assertion was added to the line before `entry['pango_lineage'] = lineage_dict[0]['lineage']` then this would have been an `AssertionError` instead of the `IndexError`.

PR #495, opened 2025-03-24
Update GitHub Actions cron timing

trvrb:
Instead of checking for sequences every day, instead check for sequences every week, only on Saturdays. Using Saturday here empirically as the latest updates have been:

- Sat Mar 22
- Sat Mar 15
- Sat Mar 8
- Sat Mar 1
- Sat Feb 22
- Sat Feb 15
- Sat Feb 8
- Sat Feb 1

### DIFF
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -4,23 +4,16 @@ on:
   schedule:
     # Note times are in UTC, which is 1 or 2 hours behind CET depending on daylight savings.
     #
-    # Currently, we aim to trigger ingest every day at 18:07 UTC which is 19:07 CET (as of Mar 2022).
-    # Note the actual runs might be late. As of right now, the action starts around 20 past the hour.
-    # Numerous people were confused, about that, including me:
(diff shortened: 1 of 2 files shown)


genehack (approved):

.github/workflows/fetch-and-ingest-genbank-master.yml
```diff
@@ -4,23 +4,16 @@ on:
...
+    # Currently, we aim to trigger runs at 6pm UTC on Saturday
     #
     # Note, '*' is a special character in YAML, so you have to quote this string.
     #
     # Docs:
     #  - https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule
     #
-    # Tool that deciphers this particular format of crontab string:
-    #  - https://crontab.guru/
-    #
```
victorlin, line 20: I go to this website whenever I want to understand GitHub Actions schedule syntax, so the link is useful. However, I do see that it's linked in the GitHub doc link above, so I'm fine with the removal.

victorlin (approved):

trvrb:
Thanks for the feedback!

PR #515, opened 2026-01-07
Simplify GISAID data update

corneliusroemer:
This PR simplifies the process of adding new GISAID sequences.

It's no longer required to a) download multiple files, b) rename them, c) run a local Snakemake workflow. Instead one can simply download a single file (.tar containing both sequences and metadat) from GISAID and push it with a single `aws s3 cp` command: `aws s3 cp /Users/cr/Downloads/gisaid_auspice_input_hcov-19_2026_01_07_11.tar s3://nextstrain-ncov-private/gisaid-downloads/unprocessed/`

Rather than using checkpointing in the snakemake workflow, there's now a single Python script that downloads all tars and turns them into ndjson.

Test run: https://github.com/nextstrain/ncov-ingest/actions/runs/20780469229/job/59676511113


workflow/snakemake_rules/upload.smk
```diff
@@ -124,49 +124,45 @@ rule remove_rerun_touchfile:
...
-            --recursive \
-            --exclude "*" \
-            --include "{wildcards.gisaid_pair}-metadata.tsv.zst" \
-            --include "{wildcards.gisaid_pair}-sequences.fasta.zst"
+        # Only run if s3_src == s3_dst (not a trial run)
+        if [[ "{params.s3_src}" != "{params.s3_dst}" ]]; then
+            echo "Skipping move (trial run: s3_src != s3_dst)"
+            exit 0
+        fi
+
```
corneliusroemer:
```suggestion
```

README.md
```diff
@@ -28,21 +28,11 @@ Follow these instructions to run the ncov-ingest pipeline _without_ all the bell
...
-
-```sh
-nextstrain build \
-  --image nextstrain/ncov-ingest \
-  --env GISAID_API_ENDPOINT \
-  --env GISAID_USERNAME_AND_PASSWORD \
-  . \
-    --configfile config/local_gisaid.yaml
-```
+1. Download them from GISAID using the "Input for the Augur pipeline" option. This results in a file of the form `gisaid_auspice_input_hcov-19_2026_01_07_11.tar` (exact date will vary).
```
joverlee521: I'd push to rename the file during download to our own standardized name so that we are not at the mercy of the default name that GISAID produces, e.g.`2026-01-07-01.tar`. This ensures that the downloaded *.tar are sorted as expected for the deduplication process.

bin/process-unprocessed-tars
```diff
@@ -0,0 +1,144 @@
...
+def main():
+    source_path = sys.argv[1]  # e.g., s3://nextstrain-ncov-private/gisaid-downloads/unprocessed/ OR local path
+    output_ndjson = sys.argv[2]  # e.g., data/gisaid/unprocessed-combined.ndjson
+    manifest_file = sys.argv[3]  # e.g., data/gisaid/processed-manifest.txt
+
+    # Check if source is S3 or local path
+    if source_path.startswith("s3://"):
+        # Download all tar files from S3
+        tmp_dir = Path("tmp/unprocessed-tars")
+        tmp_dir.mkdir(parents=True, exist_ok=True)
```
joverlee521: _non-blocking_

Should this `tmp_dir` get cleaned up at the end of the script?

bin/process-unprocessed-tars
```diff
@@ -0,0 +1,144 @@
...
+
+    print(f"\nSuccessfully processed {len(processed_tars)}/{len(tar_files)} tar files", file=sys.stderr)
+    if processed_tars:
+        print("Processed tars:", file=sys.stderr)
+        for tar_name in processed_tars:
+            print(f"  ✓ {tar_name}", file=sys.stderr)
+
+    failed_count = len(tar_files) - len(processed_tars)
+    if failed_count > 0:
+        print(f"\nWarning: {failed_count} tar files failed to process", file=sys.stderr)
```
joverlee521: _non-blocking_

Could be nice to have a list of failed tar files for debugging.

workflow/snakemake_rules/fetch_sequences.smk
```diff
@@ -39,100 +37,42 @@ if config.get("s3_src"):
...
-    assert len(GISAID_PAIRS), "No GISAID metadata and sequences inputs were found"
-
-    return expand("data/gisaid/{gisaid_pair}.ndjson", gisaid_pair=GISAID_PAIRS)
-
-
-rule concatenate_gisaid_ndjsons:
     input:
-        ndjsons=aggregate_gisaid_ndjsons,
+        gisaid_cache="data/gisaid/gisaid_cache.ndjson" if config.get('s3_src') else [],
+        unprocessed="data/gisaid/unprocessed-combined.ndjson" if config.get('s3_src') else [],
```
joverlee521: If both of these inputs are only available when `s3_src` is set, then the `cat` command below will hang if `s3_src` is not set. This also means there's no way to run the workflow on a local set of tar files.

bin/process-unprocessed-tars
```diff
@@ -0,0 +1,144 @@
...
+    for tar in sorted(tar_files):
+        print(f"  - {tar.name}", file=sys.stderr)
+    print("", file=sys.stderr)
+
+    processed_tars = []
+    all_records = []
+
+    # Sort by filename to ensure consistent ordering
+    # Process oldest first so newer records overwrite during deduplication
+    for tar_path in sorted(tar_files):
```
joverlee521: The [dedup-by-gisaid-id script](https://github.com/nextstrain/ncov-ingest/blob/dd264b7cff54333b2f12c193c7e39d1e8e87d8ca/bin/dedup-by-gisaid-id#L3-L4) keeps the _first_ record, so this needs to be sorted in reverse order to keep the newer records.

workflow/snakemake_rules/fetch_sequences.smk
```diff
@@ -39,100 +37,42 @@ if config.get("s3_src"):
...
     output:
         ndjson=temp("data/gisaid.ndjson"),
     params:
         gisaid_id_field="covv_accession_id",
     log: "logs/concatenate_gisaid_ndjsons.txt"
     shell:
         r"""
-        (cat {input.ndjsons:q} \
+        (cat {input.gisaid_cache:q} {input.unprocessed:q} \
```
joverlee521: The unprocessed file needs to be _before_ the gisaid cache since the [dedup-by-gisaid-id script](https://github.com/nextstrain/ncov-ingest/blob/dd264b7cff54333b2f12c193c7e39d1e8e87d8ca/bin/dedup-by-gisaid-id#L3-L4) keeps the first record
```suggestion
        (cat {input.unprocessed:q} {input.gisaid_cache:q} \
```

joverlee521 (changes requested):
I'm fine with moving to the .tar files if that's easier to manual processing, but I'd still push to keep our own standard file naming to ensure sorting order and also to compress the .tar files before uploading to S3.

joverlee521:
While reviewing https://github.com/nextstrain/private/pull/247, I realized this branch was being used for manual updates for the last couple months. I'm merging this as-is since it's been in use and I want to bring the default branch up-to-date for #529.

PR #525, opened 2026-03-30
combine_alignments: remove `benchmark` directive

joverlee521:

With Nextstrain CLI v10.5.0, the `--benchmark-extended` flag is automatically added in `nextstrain build` commands. This causes a `FileNotFoundError` during benchmark generation for the `combine_alignments` rule because the shell block potentially renames the input files to output. Renaming inputs to outputs is an unusual practice in Snakemake workflows because it can mess up the DAG for partial runs, but we do so here to speed up the workflow by avoiding copying giant FASTA files. We do not want to slow down the workflow for benchmark generation, so just remove the `benchmark` directive for the affected rule.

## Related issue(s)

Resolves https://github.com/nextstrain/ncov-ingest/issues/524
Related to https://github.com/nextstrain/cli/pull/513

## Checklist

- [x] Checks pass
- [x] Confirmed locally that workflow no longer raises `FileNotFoundError`
- [ ] ~Update changelog~

### DIFF
--- a/workflow/snakemake_rules/nextclade.smk
+++ b/workflow/snakemake_rules/nextclade.smk
@@ -345,8 +345,10 @@ rule combine_alignments:
         new_alignment=f"data/{database}/nextclade.{{seqtype}}.upd.fasta",
     output:
         alignment=f"data/{database}/{{seqtype}}.fasta",
-    benchmark:
-        f"benchmarks/combine_alignments_{database}{{seqtype}}.txt"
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Thanks for catching this!

joverlee521:
Manually triggered the workflow: https://github.com/nextstrain/ncov-ingest/actions/runs/23808539092

PR #528, opened 2026-06-05
Stop uploading .gz/.xz files

### LINKED ISSUE #502: Stop producing *.xz and/or *.gz files?
(Per 1:1 conversation with @joverlee521…)

While reviewing S3 bucket contents as part of [the cleanup effort](https://github.com/nextstrain/private/issues/179), I noticed that we're generating multiple copies of the same file using different compression schemes (e.g., `aligned.fasta.xz` and `aligned.fasta.zst`, `flagged_metadata.txt.gz` and `flagged_metadata.txt.zst`, etc).

Is the ZST format mature enough that we can consolidate down to just generating those files? (Or, alternatively, if we think we're never going to be rid of gz and/or xz, can we just generate those? Multiple versions of things that differ only by the compression scheme feels like a waste of resources.)

joverlee521:

Updates the workflow to stop using and uploading .gz/.xz files since we've standardized around .zst files. Outside of changes in this repo, we should also manually delete the .gz/.xz objects on S3.

Verified that the Nextstrain ncov workflow starts from the .zst files ([open](https://github.com/nextstrain/ncov/blob/4ece1d4d660a423bd17f582b8caf509c8a319168/nextstrain_profiles/nextstrain-open/builds.yaml#L35-L36), [gisaid](https://github.com/nextstrain/ncov/blob/4ece1d4d660a423bd17f582b8caf509c8a319168/nextstrain_profiles/nextstrain-gisaid/builds.yaml#L33-L34)), but will need to update [docs on available files](https://github.com/nextstrain/ncov/blob/4ece1d4d660a423bd17f582b8caf509c8a319168/docs/src/reference/remote_inputs.rst?plain=1#L27-L29). Note the ncov workflow itself [uploads .xz build files](https://github.com/nextstrain/ncov/blob/4ece1d4d660a423bd17f582b8caf509c8a319168/workflow/snakemake_rules/common.smk#L312-L314). However, it's not uploading duplicate .zst files, so it's not as pressing to switch over there. 

Also verified that the forecasts-ncov workflow starts from the .zst files ([open](https://github.com/nextstrain/forecasts-ncov/blob/149318290917b8320807774f09496c273cd0c8dc/ingest/config/defaults.yaml#L53), [gisaid](https://github.com/nextstrain/forecasts-ncov/blob/149318290917b8320807774f09496c273cd0c8dc/ingest/config/defaults.yaml#L14)).

## Related issue(s)

Resolves https://github.com/nextstrain/ncov-ingest/issues/502

## Checklist

- [x] Checks pass
- [ ] ~Update changelog~


vendored/notify-on-record-change
```diff
@@ -24,7 +24,7 @@ dst_record_count="$(aws s3api head-object --bucket "$bucket" --key "$key" --quer
 if [[ -z "$dst_record_count" ]]; then
   # This object doesn't have the record count stored as metadata
   # We have to download it and count the lines locally
-  dst_record_count="$(wc -l < <(aws s3 cp --no-progress "$dst" - | xz -T0 -dcfq))"
+  dst_record_count="$(wc -l < <(aws s3 cp --no-progress "$dst" - | zstd -T0 -dcq))"
```
joverlee521: Oh, just realized in review that this was a vendored script. 

I'll drop this change in this PR and make the change in nextstrain/shared. This PR does not _have_ to rely on that change since this only runs if the object does not have a record count stored as metadata.

joverlee521: Removed change in [force-push](https://github.com/nextstrain/ncov-ingest/compare/54e21859836b5b51e6cfae743aa72239ddf8a554..47a239a7374b7bb67a2accf4aa485f0387efeba6) and updating in shared script in https://github.com/nextstrain/shared/pull/74.

victorlin (approved):
> we should also manually delete the .gz/.xz objects on S3

+1, and soon after merging so that users aren't unknowingly using stale data.

joverlee521:
Manually deleted the _current_ versions of .gz/.xz files ([gist](https://gist.github.com/joverlee521/096ed54308b9f68492e1caf0b12ff5fb))

```
$ /tmp/delete-s3-gz-xz-files
delete: s3://nextstrain-ncov-private/aligned.fasta.xz
delete: s3://nextstrain-ncov-private/metadata.tsv.gz
delete: s3://nextstrain-ncov-private/nextclade.tsv.gz
delete: s3://nextstrain-ncov-private/sequences.fasta.xz
delete: s3://nextstrain-ncov-private/additional_info.tsv.gz
delete: s3://nextstrain-ncov-private/flagged_metadata.txt.gz
delete: s3://nextstrain-ncov-private/gisaid.ndjson.xz
delete: s3://nextstrain-data/files/ncov/open/aligned.fasta.xz
delete: s3://nextstrain-data/files/ncov/open/metadata.tsv.gz
delete: s3://nextstrain-data/files/ncov/open/nextclade.tsv.gz
delete: s3://nextstrain-data/files/ncov/open/sequences.fasta.xz
delete: s3://nextstrain-data/files/ncov/open/biosample.ndjson.gz
delete: s3://nextstrain-data/files/ncov/open/biosample.tsv.gz
delete: s3://nextstrain-data/files/ncov/open/cog_uk_accessions.tsv.gz
delete: s3://nextstrain-data/files/ncov/open/cog_uk_metadata.csv.gz
delete: s3://nextstrain-data/files/ncov/open/duplicate_biosample.txt.gz
delete: s3://nextstrain-data/files/ncov/open/genbank.ndjson.xz
delete: s3://nextstrain-data/files/ncov/open/mutation-summary.tsv.xz
```

Older versions should get automatically cleaned up with lifecycle rules to be added for https://github.com/nextstrain/private/issues/180 and https://github.com/nextstrain/private/issues/245.

PR #529, opened 2026-06-09
Upload to archive bucket

victorlin:
Adding for use with one file, but the code can be adapted to work with any file.

## Checklist

- [x] Tested new archive script using AWS CLI v1

	<details>
	
	First run uploads successfully:
			
	```console
	$ ./bin/archive-to-s3 \
	   victorlin-test.txt \
	   s3://nextstrain-archive/test \
	   victorlin-test.txt.zst
	Uploading victorlin-test.txt → s3://nextstrain-archive/test/victorlin-test_2026-06-22T18:26:46.txt.zst
	Quiet mode. No Slack notification sent.
	```
		
	Second run skips upload:
	
	```console
	$ ./bin/archive-to-s3 \
	   victorlin-test.txt \
	   s3://nextstrain-archive/test \
	   victorlin-test.txt.zst
	Found existing archive for victorlin-test.txt.zst this month, skipping upload
	```

	</details>

- [x] Checks pass
- [ ] Post-merge: next manual upload is successful

### DIFF
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -52,6 +52,8 @@ jobs:
 
           if [[ "$TRIAL_NAME" ]]; then
             config+=" s3_dst=s3://nextstrain-ncov-private/trial/${TRIAL_NAME}"
+          else
+            config+=" s3_archive_dst=s3://nextstrain-archive/ncov"
           fi
 
           echo "config=$config" >> "$GITHUB_OUTPUT"
--- a/bin/archive-to-s3
+++ b/bin/archive-to-s3
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+main() {
+    local src="${1:?A source file is required as the first argument.}"
+    local s3_prefix="${2:?A destination s3:// prefix is required as the second argument.}"
+    local remote_filename="${3:?A remote filename is required as the third argument.}"
+    local filename archive_filename dst
+
+    s3_prefix="${s3_prefix%/}"
+    filename="$(basename "$remote_filename")"
+    archive_filename="$(make_archive_filename "$filename")"
+
+    dst="$s3_prefix/$archive_filename"
+    if archived_this_month "$s3_prefix" "$filename"; then
+        echo "Found existing archive for $filename this month, skipping upload"
+        return
+    fi
+
+    # TODO: Upload with an if-none-match header to prevent accidental overriding
+    # in the non-versioned bucket. This would require using lower-level aws
+    # s3api calls and handling of multipart uploads. It is blocked by Nextstrain
+    # Docker image being stuck on AWS CLI v1, but the chance of collision is
+    # slim with the timestamped filenames.
+    # <https://github.com/nextstrain/docker-base/issues/216>
+    "$(dirname "$0")/../vendored/upload-to-s3" --quiet "$src" "$dst"
+}
+
+make_archive_filename() {
+    # Return the timestamped filename.
+
+    local filename="$1"
+    local timestamp basename extension
+
+    timestamp="$(date -u +"%Y-%m-%dT%H:%M:%S")"
+
+    if [[ "$filename" == *.* ]]; then
+        basename="${filename%%.*}"
+        extension="${filename#*.}"
+        echo "${basename}_${timestamp}.${extension}"
+    else
+        echo "${filename}_${timestamp}"
+    fi
+}
+
+archived_this_month() {
+    # Return true if the S3 prefix already has an archive for the given file in
+    # the current month.
+
+    local s3_prefix="$1"
+    local filename="$2"
+    local timestamp month basename s3path bucket prefix_path prefix existing_key
+
+    timestamp="$(date -u +"%Y-%m-%dT%H:%M:%S")"
+    month="${timestamp:0:7}"
+    basename="${filename%%.*}"
+
+    s3path="${s3_prefix#s3://}"
+    if [[ "$s3path" == */* ]]; then
+        bucket="${s3path%%/*}"
+        prefix_path="${s3path#*/}"
+        prefix="$prefix_path/${basename}_${month}"
+    else
+        bucket="$s3path"
+        prefix="${basename}_${month}"
+    fi
+    if ! existing_key="$(
+        # NOTE: The ListObjectsV2 action does *not* move objects back to the
+        # Frequent Access tier.
+        # <https://docs.aws.amazon.com/AmazonS3/latest/userguide/intelligent-tiering-overview.html>
+        aws s3api list-objects-v2 \
+            --bucket "$bucket" \
+            --prefix "$prefix" \
+            --max-items 1 \
+            --query 'Contents[0].Key' \
+            --output text
+    )"; then
+        exit 2
+    fi
+
+    [[ "$existing_key" != "None" ]]
+}
+
+main "$@"
--- a/workflow/snakemake_rules/upload.smk
+++ b/workflow/snakemake_rules/upload.smk
@@ -79,6 +79,19 @@ def compute_files_to_upload():
 
 files_to_upload = compute_files_to_upload()
 
+def archive_targets(wildcards):
+    if (
+        config.get("fetch_from_database", False) and
+        config.get("s3_archive_dst")
+    ):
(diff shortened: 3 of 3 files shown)


.github/workflows/fetch-and-ingest-gisaid-master.yml
```diff
@@ -52,6 +52,8 @@ jobs:
 
           if [[ "$TRIAL_NAME" ]]; then
             config+=" s3_dst=s3://nextstrain-ncov-private/trial/${TRIAL_NAME}"
+          else
+            config+=" s3_archive_dst=s3://nextstrain-archive"
```
joverlee521: To match the previously archived files, this should also include the `ncov` prefix

```suggestion
            config+=" s3_archive_dst=s3://nextstrain-archive/ncov"
```

workflow/snakemake_rules/upload.smk
```diff
@@ -105,6 +118,24 @@ rule upload_single:
             {params.cloudfront_domain} 2>&1 | tee {output}
         """
 
+rule archive:
+    input:
+        file_to_upload=lambda w: files_to_upload[w.remote_filename],
+        upload_single_touchfile="data/{database}/{remote_filename}.upload",
```
joverlee521, line 124: Hmm, I think we'd only want to archive the file if it has any changes. We could technically check the contents of this `upload_single_touchfile` to verify that the file was actually uploaded because it had new content.

workflow/snakemake_rules/upload.smk
```diff
@@ -105,6 +118,24 @@ rule upload_single:
...
+    benchmark:
+        "benchmarks/archive_{database}_{remote_filename}.txt"
+    params:
+        s3_bucket=config.get("s3_archive_dst", ""),
+    shell:
+        """
+        ./vendored/upload-to-s3 \
+            --quiet \
+            {input.file_to_upload:q} \
+            {params.s3_bucket:q}/{wildcards.remote_filename:q} 2>&1 | tee {output}
```
joverlee521: Should the uploaded object have a timestamp in the name similar to the existing archived objects?

joverlee521: From https://github.com/nextstrain/infra/pull/72#discussion_r3383577018, this would need to be updated to support `if-none-match` header. 

We could update `shared/upload-to-s3` to support `if-none-match` (and bypassing the [hash comparison](https://github.com/nextstrain/shared/blob/bfa9de236fc0dae1d42a4e3a36b95ff586c51438/scripts/upload-to-s3#L27-L31)) or just directly use `aws s3api put-object --if-none-match '*'`.

I'm in favor of just using `aws` directly since this is a single use-case.

joverlee521 (commented):
Left comments as I was prompted by https://github.com/nextstrain/infra/pull/72#discussion_r3383577018.

.github/workflows/fetch-and-ingest-gisaid-master.yml
victorlin: Done:

https://github.com/nextstrain/ncov-ingest/blob/42968f4b568ee77987148e03059209aa82da9db8/.github/workflows/fetch-and-ingest-gisaid-master.yml#L56

workflow/snakemake_rules/upload.smk
victorlin: Done (untested for now):

https://github.com/nextstrain/ncov-ingest/blob/42968f4b568ee77987148e03059209aa82da9db8/workflow/snakemake_rules/upload.smk#L133-L139

workflow/snakemake_rules/upload.smk
victorlin: `aws s3api put-object` only works for files [up to 5 GB](https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html). We'll need to use different commands to implement multipart upload. It should be straightforward – I'll wrap this up in a new `scripts/archive-to-s3` which can also take care of timestamping filenames.

victorlin: Yes, I'll plan to do this in a new `scripts/archive-to-s3` as mentioned in https://github.com/nextstrain/ncov-ingest/pull/529#discussion_r3384323002

victorlin: Done:

https://github.com/nextstrain/ncov-ingest/blob/aafc8cd6c29a4ce28b4f267e3b4c8cfda313ff1c/bin/archive-to-s3#L23

victorlin: Done (untested for now):

https://github.com/nextstrain/ncov-ingest/blob/aafc8cd6c29a4ce28b4f267e3b4c8cfda313ff1c/bin/archive-to-s3#L148-L153

victorlin:
Updated the PR description with testing notes. I've manually tested the new archiving script, but didn't test archiving the files from this workflow as that might want to wait for #515.

bin/archive-to-s3
```diff
@@ -0,0 +1,189 @@
...
+        return
+    fi
+
+    local src_hash payload payload_size
+    src_hash="$("$bin/../vendored/sha256sum" < "$src")"
+    tmpdir="$(mktemp -d)"
+    payload="$tmpdir/payload"
+    trap cleanup EXIT
+
+    cp "$src" "$payload"
```
joverlee521: Why make a copy of `src` here? Oh, should this be compressing the file based on the extension similar to upload-to-s3?

bin/archive-to-s3
```diff
@@ -0,0 +1,189 @@
...
+        bucket="$s3path"
+        prefix="${basename}_${month}"
+    fi
+    if ! existing_key="$(
+        aws s3api list-objects-v2 \
+            --bucket "$bucket" \
+            --prefix "$prefix" \
+            --max-items 1 \
+            --query 'Contents[0].Key' \
+            --output text
```
joverlee521, line 76: Just in case others are curious, I checked the [Intelligent-tiering docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intelligent-tiering-overview.html) to verify that listing objects should **not** move objects out of archive:

> Other actions don't constitute access that automatically moves objects from the Infrequent Access tier or the Archive Instant Access tier back to the Frequent Access tier. The following is a sample, not a definitive list, of such actions:
> 
> Invoking [HeadObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html), [GetObjectTagging](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectTagging.html), [PutObjectTagging](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectTagging.html), [ListObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html), [ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html), [ListObjectVersions](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions), and [UpdateObjectEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UpdateObjectEncryption.html).

bin/archive-to-s3
victorlin: Oh yeah it should be compressing. The streamed compression done in upload-to-s3 doesn't work well with multipart uploads, but I think we can remove the multipart upload handling entirely now that we don't need `if-none-match` and can use `aws s3 cp` directly.

victorlin: Now calling `upload-to-s3` within this script.

https://github.com/nextstrain/ncov-ingest/blob/4a714833584ec81cb989b9fe5feef828b3ded3e6/bin/archive-to-s3#L20-L26

bin/archive-to-s3
victorlin: Noted inline.

https://github.com/nextstrain/ncov-ingest/blob/4a714833584ec81cb989b9fe5feef828b3ded3e6/bin/archive-to-s3#L68-L71

bin/archive-to-s3
victorlin: Compression works:

```console
 $ ./bin/archive-to-s3 \
    victorlin-test.txt \
    s3://nextstrain-archive/test \
    victorlin-test.txt.zst
 Uploading victorlin-test.txt → s3://nextstrain-archive/test/victorlin-test_2026-06-22T18:26:46.txt.zst
 Quiet mode. No Slack notification sent.
```

joverlee521 (approved):

PR #530, opened 2026-06-15
Fix GenBank ingest

victorlin:
Follow-up to #515 which caused an InputFunctionException in the GenBank path.

This rule only applies to GISAID, but it was being run whenever s3_src == s3_dst (true for GenBank runs).

Make the GISAID condition explicit and name the rule/touchfile accordingly.

## Checklist

- [x] `nextstrain build . --configfile config/genbank.yaml --dry-run --quiet` works
- [x] Checks pass
- ~Update changelog~

### DIFF
--- a/workflow/snakemake_rules/upload.smk
+++ b/workflow/snakemake_rules/upload.smk
@@ -107,7 +107,16 @@ rule remove_rerun_touchfile:
         """
 
 
-rule mv_all_processed_tars:
+def should_move_processed_gisaid_tars():
+    return (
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Ah, thanks for fixing @victorlin!

joverlee521:
The [GenBank workflow](https://github.com/nextstrain/ncov-ingest/actions/runs/27654586231/attempts/1) successfully completed after the fix.

PR #534, opened 2026-06-26
Fix GISAID ingest

victorlin:
2 follow-up commits after actually running changes in https://github.com/nextstrain/ncov-ingest/actions/runs/28135995624.

## Checklist

- [x] `nextstrain build . --configfile config/gisaid.yaml --dry-run --quiet` works
- [x] Checks pass


victorlin:
This is primarily meant to fix the archive rule added in #529. Merging so that we can see whether it works in the next manual run.

PR #535, opened 2026-07-01
Remove dead code and unused deps; fix use-nextclade-cache fallback

trvrb:
_Via Claude_

First pass of an ncov-ingest cleanup. **All changes are output-neutral** — the subsampled open/GenBank transform + merge outputs (`metadata_transformed.tsv` + `sequences.fasta`) are byte-identical before and after (md5 of the column set, sorted rows, and sorted FASTA records all unchanged).

## Changes
- **Delete 9 unreferenced `bin/` scripts** (~840 LoC): `check-gisaid-geoRules`, `detect-generalizable-geoRules`, `filter-gisaid-annot-using-transform-log`, `check-annotations`, `split-fasta`, `gisaid-get-all-clades`, `join-rows`, `metadata-additions`, `launch-build`. Each verified to have zero references across the repo (rules, other scripts, `.github/`, docs, config).
- **`requirements.txt`:** drop unused `envdir`. (Kept `regex` — used with a `regex.V1`-specific split — and `csv-diff`, still used by `notify-on-additional-info-change`.)
- **`Dockerfile`:** drop unused apt packages `aria2`, `lbzip2`, `pigz`, `pixz` (none referenced anywhere in the workflow).
- **Stale env docs:** remove `GISAID_API_ENDPOINT` / `GISAID_USERNAME_AND_PASSWORD` from the Snakefile `onstart` list + README — the GISAID path ingests via S3 tars now, not the API (their only code user was `gisaid-get-all-clades`, deleted here).
- **Bug fix — `bin/use-nextclade-cache`:** the cache-version fallback invoked a non-existent `bin/cache-version`; now calls `fetch-cache-version` on the src file (the intended behavior).
- **Comment fix — `notify_genbank`:** resolve the "which rule produces `problem_data.tsv`?" TODO (`transform-genbank` writes it via its `--problem-data` default).

## Verification
- **Output-equivalence:** subsampled open GenBank run — `metadata_transformed.tsv` / `sequences.fasta` md5s unchanged before/after.
- **DAG intact:** `snakemake -n` job list unchanged; no rule references a deleted script.
- **Not locally exercised:** the `use-nextclade-cache` fix (S3 nextclade-cache path) and the image rebuild — justified by inspection; neither affects transform/merge outputs.

Note: the `Dockerfile` / `requirements.txt` changes take effect once `update-image.yml` rebuilds the image on merge.


trvrb:

PR #536, opened 2026-07-01
Stream FASTA in merge-open and filter-fasta (drop Biopython SeqIO)

trvrb:
_Via Claude_

Second cleanup PR. Replaces the slow Biopython `SeqIO` FASTA passes in `merge-open` and `filter-fasta` with a small shared line-streaming reader (`lib/utils/fasta.py`). **Output-neutral.**

## Why
Both scripts read the whole-corpus FASTA every run via `SeqIO.parse` (a `SeqRecord` per record); `filter-fasta` also rewrites it via `SeqIO.write`. Line-streaming is faster and lower-memory. The streaming `id` is the first whitespace token of the header, matching `SeqRecord.id`, so id-membership checks are identical.

## Output-equivalence (verified on the open GenBank subsample)
- **merge-open:** `sequences.fasta` + `metadata_transformed.tsv` byte-identical to before (md5 of column set, sorted rows, and sorted FASTA records all unchanged).
- **filter-fasta:** empty-cache output byte-identical to the input; a partial-cache run emits exactly the records whose id is absent from the cache TSV.

## Performance (20k-record / 597 MB fixture)
| step | before (Biopython) | after (streaming) | speedup |
|---|---|---|---|
| merge-open FASTA read | 0.41 s | 0.15 s | ~2.7× |
| filter-fasta read+write | 1.59 s | 0.31 s | ~5.2× |

Modest at this size but grows with the corpus and removes per-record `SeqRecord` allocation. `filter-fasta`'s output is a Nextclade-input intermediate, so dropping `SeqIO.write`'s 60-col wrap is harmless.

Testing production performance via GitHub Action https://github.com/nextstrain/ncov-ingest/actions/runs/28487102285. Full reruns are generally about 7.5 hours.

## Scope note
The `transform-genbank` double NDJSON parse (another redundant full-corpus pass) is deliberately left for a later PR — it's a memory-for-time tradeoff entangled with the whole-corpus in-memory sort, best reworked together with the streaming-sort rearchitecture.

### DIFF
--- a/bin/filter-fasta
+++ b/bin/filter-fasta
@@ -1,8 +1,12 @@
 #!/usr/bin/env python3
 import argparse
+import sys
+from pathlib import Path
 
 import pandas as pd
-from Bio import SeqIO
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
+from utils.fasta import stream_fasta
 
 
 def parse_args():
@@ -33,11 +37,10 @@ def main():
     tsv = pd.read_csv(args.input_tsv, sep="\t", usecols=["seqName"], dtype=str)
     tsv_ids = set(tsv['seqName'])
 
-    with open(args.input_fasta) as f_input:
-        with open(args.output_fasta, "w") as f_output:
-            for seq in SeqIO.parse(f_input, "fasta"):
-                if seq.id not in tsv_ids:
-                    SeqIO.write(seq, f_output, "fasta")
+    with open(args.input_fasta) as f_input, open(args.output_fasta, "w") as f_output:
+        for seq_id, sequence in stream_fasta(f_input):
+            if seq_id not in tsv_ids:
+                f_output.write(f">{seq_id}\n{sequence}\n")
 
 
 if __name__ == '__main__':
--- a/bin/merge-open
+++ b/bin/merge-open
@@ -14,9 +14,14 @@ def main(
     output_sequences: str = typer.Option(...),
     output_metadata: str = typer.Option(...),
 ):
+    import sys
+    from pathlib import Path
+
     import pandas as pd
-    from Bio import SeqIO
     from xopen import xopen
+
+    sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
+    from utils.fasta import stream_fasta
 
     # Load metadata files
     with xopen(input_rki_metadata, "r") as fin:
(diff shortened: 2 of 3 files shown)


trvrb:
_Via Claude_

## Production benchmarking (per-rule wall + RSS)

Comparing the Snakemake `benchmark:` files (col 1 = wall s, col 3 = `max_rss` MB, col 10 = cpu s) for the two rules this PR touches — `merge_open_data` (merge-open) and `get_sequences_without_nextclade_annotations` ×2 (filter-fasta).

### Jobs
| | AWS Batch job ID | GH Action | Nextclade cache |
|---|---|---|---|
| **baseline** (master) | `f1504818-4e36-447b-9474-d6b1b5c1c1d4` | — | **warm** |
| **this PR (#536)** | `e30add64-9572-463b-b5b6-f36f3162efb8` | run 28487102285 (10h 50m) | **cold** |

### Results
| rule | baseline wall | baseline max_rss | PR wall | PR max_rss | note |
|---|---|---|---|---|---|
| `merge_open_data` | 4792 s (80 min) | 11.4 GB | 4738 s (79 min) | 11.2 GB | unchanged — valid comparison |
| `get_sequences…genbank` | 3918 s | 1.16 GB | 2295 s | 0.0 GB | not comparable (see below) |
| `get_sequences…genbank_21L` | 3918 s | 1.16 GB | 2295 s | 0.0 GB | not comparable (see below) |

**`merge_open_data`: no meaningful change** (4738 s vs 4792 s; 11.2 vs 11.4 GB). Exactly as predicted — the rule is I/O-bound (cpu 477 s of 4738 s wall), so streaming the FASTA doesn't move the wall, and the 11 GB RSS is the pandas metadata load (a later PR / Phase 5c).

**`get_sequences…` comparison is invalid this round.** The PR job ran with a **cold Nextclade cache**, so `get_sequences` took the `ln` branch (RSS **0.0 GB** — filter-fasta never ran) rather than the filter-fasta path the baseline exercised (RSS 1.16 GB, warm cache). filter-fasta's production wall/RSS still needs a **warm-cache** PR run to measure.

## Why this run took 10h 50m (it's not this PR)

The run is dominated by a **full, from-scratch Nextclade run over the entire corpus**:

| rule | PR wall | PR cpu | master (warm) |
|---|---|---|---|
| `run_wuhan_nextclade_genbank` | **4.3 h** | 247,441 s | ~1 s |
| `run_21L_nextclade_genbank` | **3.6 h** | 167,729 s | ~1 s |

~8 h of the ~11 h run. Cause (from the job log):

```
[INFO] Current Nextclade version (nextclade 3.21.2) is different from cache version ()
```

The cache version is **empty**: this is a **trial run** whose `s3_dst` is the branch prefix `s3://nextstrain-data/files/ncov/open/trial/cleanup-fasta-streaming`, which has no prior `nextclade.tsv.zst`, so `use-nextclade-cache` got no version and fell through to a full rerun. On a warm cache (e.g. master's `f1504818`) `run_*_nextclade` is ~1 s and the whole pipeline is ~3 h.

**So the slowness is the cold trial cache, not the FASTA-streaming changes.** Sanity check: `transform_genbank_data` (untouched by this PR) was 4954 s / 36.2 GB here vs 4897 s / 36.2 GB on master — same environment.

Follow-up worth filing (separate from this PR): trial runs don't fall back to the **production** Nextclade cache (`s3_src`), so every first-run-of-a-branch pays a full ~8 h Nextclade rerun. Fixing that would make branch CI both faster and much cheaper. Also note the two `run_21L` rows disappear once #538 (21L removal) lands.

trvrb:
So... Claude's hypothesis that `merge-open` and `fasta-filter` were good targets for optimization were based on working with a small local subsample. Running the full build results in IO bound rules and the CPU performance improvements fail to materialize. I'm just going to close this PR.

But I do note that the warm vs cold Nextclade cache behavior on `master` vs `trial` builds observed above should get fixed separately. It will make all this performance monitoring much more apples-to-apples.

PR #538, opened 2026-07-01
Remove the 21L Nextclade run

trvrb:
_Via Claude_

## Summary

ncov-ingest runs Nextclade **twice** on every sequence — once against the primary `sars-cov-2` (Wuhan) dataset and once against `sars-cov-2-21L` (BA.2), parameterized by a `{reference}` wildcard — solely to produce the `immune_escape` and `ace2_binding` metadata columns.

Those two columns are no longer consumed by anyone:
- the dedicated `ncov/gisaid/21L/global/6m` build was retired last fall;
- the main ncov build references neither column (only a historical `change_log.md` mention);
- `nextclade_21L.tsv.zst` is **not** an advertised remote input, and the two columns live only inside the advertised `metadata.tsv.zst`, which nothing reads for them.

So the 21L run is dead weight — a full redundant Nextclade pass (per the master benchmarks, e.g. `get_sequences…gisaid_21L` ~3635s, `nextclade_info_gisaid_21L` ~671s, plus the `_21L` uploads) and a second big TSV that the memory-heavy `generate_metadata` join has to load.

## What this does
- **`nextclade.smk`**: delete `run_21L_nextclade`, strip the `{reference}` wildcard from the seven shared rules so Nextclade runs exactly once, and make the dataset paths static (`sars-cov-2.zip`). Drop the 21L input from `generate_metadata`.
- **`upload.smk`**: drop the `nextclade_21L.tsv.zst` + `nextclade_21L_version.json` uploads and the renew-touchfile entry.
- **`join-metadata-and-clades`**: remove the `immune_escape`/`ace2_binding` columns, the `--nextclade-21L-tsv` input, and the associated read/merge/round/null-out.
- Delete the now-dead `scripts/developer_scripts/rerun-nextclade-21L.sh` and its README section.

Net: **−98 / +27 lines**, one file deleted. Also drops one input from `generate_metadata`, which helps the planned Phase 5b memory work on that join.

## Correctness
The published `metadata.tsv` loses **exactly two columns** (`immune_escape`, `ace2_binding`); everything else is byte-identical.

- **Join output-equivalence:** ran `join-metadata-and-clades` old-vs-new on a 4000-record slice of the real open `nextclade.tsv`. The new output is **byte-for-byte equal to the old output with those two columns dropped** — same columns (in order), same values, and the two removed columns were populated in the old output (so the merge + `< 21L` null-out path was exercised).
[...]

PR #537, opened 2026-07-01
Replace in-memory sort with an external sort in all three transforms

trvrb:
_Via Claude_

## Summary

The GenBank / GISAID / RKI transforms deduplicate records by strain, keeping the longest/earliest per strain. To do that each held **every parsed record in memory** (~9M dicts at production scale for GenBank) to run one in-memory `sorted()` by `(strain, -length, <id>, line_no)`. That list is the dominant driver of each rule's peak RSS — **~37 GB** for `transform_genbank_data` (job `f1504818`) — and the reason the AWS Batch boxes are pinned so high (open 60 GiB, GISAID 90 GiB).

This PR replaces the in-memory sort with a disk-backed external sort:
1. stream the pipeline to a temp file, tagging each record with its sort keys as leading TSV columns + the full record as a trailing JSON blob;
2. `LC_ALL=C sort` the file (spills to disk);
3. stream the sorted records once to dedup + write outputs.

**Peak memory for the sort now stays flat regardless of corpus size.**

The logic is factored into a shared helper `lib/utils/transformpipeline/externalsort.py` (`spill_to_sorted_tempfile` + `read_sorted_records`) used by all three transforms — each just passes its own dedup id column (`genbank_accession` / `gisaid_epi_isl` / `rki_accession`). This also collapses each transform's separate dedup and metadata-write loops into a single streaming pass, and sets up the Phase 4 driver-unification work.

### Commits
1. `transform-genbank`: inline external sort (the headline win).
2. Extract the shared helper; apply to `transform-gisaid` and `transform-rki`.

## Correctness (the hard gate for this repo)

Output is **byte-for-byte identical** to `master`, verified old-vs-new in the same runtime:
- **genbank** — metadata + sequences match the pre-change golden md5s on the frozen review subsample; also identical on a 60k-record synthetic corpus.
- **gisaid** — metadata + sequences + `additional_info.tsv` + flagged-annotations all identical on a 5000-record slice of the private ndjson (**dedup exercised: 5000 → 4972 records**).
- **rki** — metadata + sequences identical on a 5000-record slice.

Why the ordering is preserved:
- `LC_ALL=C` byte order == Python's code-point comparison over the UTF-8 `strain`/id fields (UTF-8 sorts bytewise the same as by code point).
- `-k2,2nr` reproduces `-length` (numeric, descending); `-k4,4n` reproduces `line_no` ascending.
- `(strain, length, id, line_no)` is a **total order** (line number is unique per record), so `sort` stability is irrelevant.

[...]

PR #539, opened 2026-07-01
Remove the GISAID flag_metadata rule

trvrb:
_Via Claude_

## Summary

`flag_metadata` is a GISAID-only rule that scans the full joined `metadata.tsv` to emit `flagged_metadata.txt` — a list of suspect strains (future collection dates, pre-March-2020 divergence, clock-deviation outliers) — and fires a Slack notification about newly-flagged sequences. It's a leftover from the manual-curation era; that QC now happens at **build** time, so the list is dead weight. It's also the **third-largest memory rule on the GISAID box** (~33 GB peak per the master benchmarks) and has no GenBank counterpart.

This removes the rule and everything wired to it:
- delete `bin/flag-metadata` and `bin/notify-on-flagged-metadata-change` (each used only here);
- delete the `flag_metadata` rule (`curate.smk`);
- drop the `flagged_metadata` input + the `notify-on-flagged-metadata-change` call from `notify_gisaid` (`slack_notifications.smk`), leaving the rest of that rule intact;
- drop the `flagged_metadata.txt.zst` upload (`upload.smk`).

Net: **−103 lines**, two files deleted, pure removal.

## No data change, no consumers

`flag_metadata` only *reads* `metadata.tsv` to produce a **separate** side-file — it never writes the published metadata/sequences, so **nothing published changes**. What goes away is the `flagged_metadata.txt` artifact, its Slack notification, and its S3 upload.

Nothing else in ncov-ingest reads `flagged_metadata.txt`, and the main ncov build has zero consumers (only a docs listing). A companion one-line ncov docs PR de-advertises the now-unpublished `flagged_metadata.txt.zst`.

## Verification

No output-equivalence gate needed (no data output changes). Structural:
- `snakemake -n` for **both** `config/gisaid.yaml` and `config/genbank.yaml` resolves cleanly — zero `flag_metadata` jobs, **zero `MissingInputException`**.
- Re-ran the gisaid dry-run with `SLACK_TOKEN`/`SLACK_CHANNELS` set (forcing the notify rules into the DAG): `notify_gisaid` and `upload` resolve without the removed `flagged_metadata` input.
- Repo grep clean of `flag_metadata` / `flagged_metadata` / `flag-metadata` / `notify-on-flagged-metadata-change`.

## Notes
- The stale `flagged_metadata.txt.zst` already on the private GISAID bucket is orphaned but harmless; deleting it from the bucket is an optional manual follow-up.
- Separate from the Phase 5 series (#537) and the 21L removal (#538).

## Test plan
- [x] `snakemake -n` clean for gisaid + genbank, zero `flag_metadata` jobs, no `MissingInputException`
- [x] `notify_gisaid` + `upload` resolve with notifications forced on
- [x] Repo grep clean
- [x] Companion ncov docs PR merged (de-advertise `flagged_metadata.txt.zst`)


trvrb:

PR #540, opened 2026-07-01
Stream generate_metadata join and split clock_deviation into its own rule

trvrb:
**Draft / WIP** until rebased onto `master` (after #538, ideally #537) and benchmarked in production.

Makes `generate_metadata` low-memory. It was the top remaining hog (~26 GB GenBank / ~49 GB GISAID) because `bin/join-metadata-and-clades` held the full metadata × Nextclade merge (~64 columns × ~9M rows) in pandas. **The published `metadata.tsv` is unchanged** (content-equivalent; see the row-order note) and nothing downstream changes.

## What it does
1. **Stream the join.** `bin/join-metadata-and-clades` is rewritten as a flat-memory external merge-join (`LC_ALL=C`-sort both inputs by the join key → two-pointer merge, one row at a time). Drops pandas. Peak → a small constant.
2. **Split `clock_deviation` into its own rule.** `clock_deviation` was the *only* reason the join needed the whole merge in memory (its per-clade offset needs all rows) — but it only needs 3 columns (`date`, `divergence`, `Nextstrain_clade`). A new `compute_clock_deviation` rule (`bin/compute-clock-deviation`) reads just those, runs the **verbatim** per-clade numpy from the old join, and appends `clock_deviation` as the last column → the final `metadata.tsv`. ~160 MB at 100k rows.

`generate_metadata` now emits a temp `metadata_without_clock_deviation.tsv`; the new rule produces `metadata.tsv`. The four consumers (`all_targets`, `metadata_version_json` sha256, GISAID `flag_metadata`, `upload`) reference the same path, so Snakemake inserts the new rule with **no other edits**.

## Why this shape (not removal / not clade-agnostic)
Earlier drafts removed `clock_deviation` (relying on ncov to recompute it) — but for `/global/6m`-style builds ncov would recompute on a *subsample* where old clades have <100 members (offset → default) . Keeping the **exact per-clade, full-corpus** value avoids that entirely: `metadata.tsv` keeps a byte-identical `clock_deviation`, ncov's join short-circuit keeps trusting it, and **the ncov side needs no change**. It's also just cleaner — a QC metric belongs in its own step, not the join.

## Verification
- **Byte-identical (strict):** on strain-sorted input, the old integrated join vs the new pipeline (streaming join → `compute-clock-deviation`) are byte-for-byte identical — confirms faithful extraction + exact join/append.
- **End-to-end:** on shuffled input, content-identical after sorting both by `strain`; `clock_deviation` matches exactly (unmatched → `?`, values like `5.0`).
- **Memory:** `compute-clock-deviation` 159 MB at 100k rows (below the transform floor). Join is architecturally flat.
- `snakemake -n` (genbank + gisaid) resolves with the new rule inserted; no `MissingInputException`.

**Output-order note:** the streaming join re-sorts rows to global strain order. GISAID's `metadata_transformed.tsv` is already strain-sorted (unchanged); the open pipeline's RKI rows interleave into strain order (same rows/values, reordered).

## Test plan
- [x] Streaming join content-equivalent (sort-by-strain); flat memory
- [x] `compute_clock_deviation` byte-identical to old (strain-sorted) + content-identical (shuffled); ~160 MB/100k
- [x] `snakemake -n` clean, new rule inserted, no consumer edits
- [x] Production Batch benchmark: `generate_metadata` 26/49 GB → flat; `compute_clock_deviation` ~1–2 GB; `metadata.tsv` content-equivalent
- [x] Rebase onto `master` once #538 (+ #537) merge

### DIFF
--- a/bin/compute-clock-deviation
+++ b/bin/compute-clock-deviation
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+"""
+Append a `clock_deviation` column to the joined metadata.
+
+`clock_deviation` is a per-clade molecular-clock QC metric: each sequence's
+divergence minus its clade's clock-expected divergence for its collection date.
+It only needs three columns -- `date`, `divergence`, `Nextstrain_clade` -- so
+computing it in its own step keeps the (much larger) metadata/Nextclade join a
+flat-memory streaming operation. The calculation here is the one that used to
+live in `bin/join-metadata-and-clades`, unchanged.
+"""
+import argparse
+import sys
+from datetime import datetime
+
+import numpy as np
+import pandas as pd
+
+CLADE_COLUMN = "Nextstrain_clade"
+CLOCK_DEVIATION_COLUMN = "clock_deviation"
+VALUE_MISSING_DATA = '?'
+
+rate_per_day = 0.0007 * 29903 / 365
+reference_day = datetime(2020, 1, 1).toordinal()
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Append a clock_deviation column to the joined metadata TSV.",
+    )
+    parser.add_argument("--metadata", required=True,
+                        help="Joined metadata TSV (output of join-metadata-and-clades)")
+    parser.add_argument("-o", default=sys.stdout,
+                        help="Output metadata TSV with clock_deviation appended")
+    return parser.parse_args()
+
+
+def datestr_to_ordinal(x):
+    try:
+        return datetime.strptime(x, "%Y-%m-%d").toordinal()
+    except:
+        return np.nan
+
+
+def isfloat(value):
+    try:
+        float(value)
+        return True
+    except ValueError:
+        return False
+
+
+def main():
+    args = parse_args()
+
+    # Pass 1: read only the three columns needed and compute clock_deviation
+    # (verbatim from the former join-metadata-and-clades logic).
+    result = pd.read_csv(args.metadata, sep='\t',
+                         usecols=["date", "divergence", CLADE_COLUMN],
+                         dtype="object", na_filter=False)
+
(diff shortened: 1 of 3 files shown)


trvrb:
_Via Claude_

## Benchmarks

Ran full-corpus trial builds on both databases and pulled the Snakemake `benchmark:` files. `max_rss` = peak RSS, `s` = wall.

### Memory — `generate_metadata` is now flat

| rule | GenBank | GISAID |
|---|---|---|
| `generate_metadata` (master) | 26,610 MB | ~49,000 MB |
| `generate_metadata` (this PR) | **21.8 MB** | **22.1 MB** |
| `compute_clock_deviation` (new rule) | 1,090 MB | 2,446 MB |

The streaming external merge-join holds ~one row at a time, so `generate_metadata` drops from ~26 GB / ~49 GB to **~22 MB**. `clock_deviation`, now its own rule, peaks at **1.1 GB / 2.4 GB** — comfortably under the `transform` floor. Net peak for the metadata stage: **~26 GB → 1.1 GB** (GenBank), **~49 GB → 2.4 GB** (GISAID).

### Wall time — modest regression from the external sort

| | GenBank | GISAID |
|---|---|---|
| `generate_metadata` (master) | 930s | — |
| `generate_metadata` (this PR) | 1,556s | 1,422s |
| `compute_clock_deviation` | 247s | 222s |
| **metadata stage total** | **1,803s** | **1,644s** |

GenBank's metadata stage goes 930s → ~1,803s (+~875s). CPU actually *drops* (635s → 411s); the extra wall is I/O wait as the external sort spills to disk and re-reads — the trade that buys the flat memory. Small against the full pipeline (`transform` alone runs 80–160 min).

trvrb:

PR #541, opened 2026-07-01
Fall back to source Nextclade cache when destination has none

trvrb:
_Via Claude_

## Summary

Trial (branch) ingest runs set `s3_dst` to a per-branch prefix (`s3://…/open/trial/<branch>`) while `s3_src` stays production. `bin/use-nextclade-cache` reads the cached Nextclade/dataset version from `s3_dst`, **falling back to `s3_src`** — except that fallback was dead:

- `bin/fetch-cache-version` **exits 0 with empty output** on a missing S3 object (no `pipefail`; the trailing `jq --raw-input` emits nothing and exits 0), so `fetch-cache-version dst || cache-version src` never triggered the `||`; and
- `bin/cache-version` **doesn't exist** in the repo anyway.

So a **first-of-a-branch trial run** read an empty version, treated the cache as stale, and reran Nextclade **from scratch over the entire corpus** — ~8h on the last #536 trial run (`run_wuhan_nextclade` 4.3h + `run_21L_nextclade` 3.6h), vs **~1s each** on a warm production run. Production runs were never affected because there `s3_dst == s3_src`, so the first read hits the real cache.

## Fix

One function, `get-cache-version-info`: fall back to `s3_src` when `s3_dst` yields no version, using the working `fetch-cache-version` for both (dropping the dead `cache-version` reference). A missing object prints empty, so a plain `[[ -z "$info" ]]` check is exactly the failure mode. The download side already falls back `dst → src` (`download_nextclade_tsv_from_s3`), so nothing else changes; a second run of the same branch still prefers its own now-populated trial cache.

## Verification

With S3 creds:
- `./bin/fetch-cache-version s3://…/open/trial/__nonexistent__/nextclade.tsv.zst` → prints **nothing** (the empty case the fallback keys on).
- `./bin/fetch-cache-version s3://nextstrain-data/files/ncov/open/nextclade.tsv.zst` → `{"nextclade_version":"nextclade 3.21.2","nextclade_dataset_version":"2026-06-16--14-30-45Z"}`.

`nextclade 3.21.2` is exactly the version the cold trial run reported as "different from cache version ()", so with this fix a trial run matches and reuses the cache. Definitive check: the next trial branch run should show `run_wuhan_nextclade` back at seconds, not hours.

## GitHub Action

Running at https://github.com/nextstrain/ncov-ingest/actions/runs/28526775213 and inspectable with
```
nextstrain build --aws-batch --attach c96ef07d-fb24-4666-9b4f-064eb775930e .
```

## Test plan
- [x] `fetch-cache-version` returns empty on a missing trial path, version JSON on production
- [x] Fixed `get-cache-version-info` falls back to `s3_src` for an empty `s3_dst`
- [x] shellcheck CI green
- [x] Next trial run: `run_wuhan_nextclade` ≈ seconds (warm), not hours

### DIFF
--- a/bin/use-nextclade-cache
+++ b/bin/use-nextclade-cache
@@ -53,7 +53,19 @@ get-cache-version-info() {
     local dst_version_file="$s3_dst/$version_file"
     local src_version_file="$s3_src/$version_file"
 
-    "$bin"/fetch-cache-version "$dst_version_file" || "$bin"/cache-version "$src_version_file"
+    # Prefer a cache already at the destination (e.g. a previous trial run of this
+    # branch); fall back to the source (production) cache when the destination has
+    # none. Without this fallback a first-time trial run — whose s3_dst is an empty
+    # per-branch prefix — gets an empty version, treats the cache as stale, and
+    # reruns Nextclade from scratch over the whole corpus. (The old `|| cache-version`
+    # fallback never fired: fetch-cache-version exits 0 on a missing object and
+    # bin/cache-version does not exist.)
+    local info
+    info="$("$bin"/fetch-cache-version "$dst_version_file")"
+    if [[ -z "$info" ]]; then
+        info="$("$bin"/fetch-cache-version "$src_version_file")"
+    fi
+    echo "$info"
 }
 
 main "$@"


trvrb:
_Via Claude_

### Confirmed on a trial build 

Trial GenBank run [28526775213](https://github.com/nextstrain/ncov-ingest/actions/runs/28526775213) (job `c96ef07d-fb24-4666-9b4f-064eb775930e`) off this branch. The fix does exactly what it should — a first-of-a-branch trial run now reuses the **production** Nextclade cache instead of rerunning from scratch:

| rule | prior cold trial | this run (warm) |
|---|---|---|
| `run_wuhan_nextclade_genbank` | 4.7h (16,816s) | **1s** |
| `run_21L_nextclade_genbank` | 3.8h (13,832s) | **0s** |

The job log shows the mechanism directly:
```
[INFO] Downloading cached nextclade.tsv.zst
[ INFO] Number of sequences to run Nextclade on: 62
```
i.e. it found and reused the production cache (via the new `s3_src` fallback) and Nextcladed only the **62 genuinely-new sequences** rather than the whole ~9M-sequence corpus. `get_sequences` also ran at 1.1 GB RSS (the warm filter-fasta path) instead of 0.0 GB (the cold `ln`-the-whole-corpus path).

**Total run time: ~11h → 7.6h**

trvrb:
This was verified to work as intended in the trial run and is a small targeted change. With this on `master` I can do better apples-to-apples performance comparisons between standard builds and trial builds. I'll going to merge this.

Issue #532, opened 2026-06-24
GISAID ingest fails with OOM


corneliusroemer:
Most recent GISAID ingest failed due to OOM at step transform-gisaid:

```
[batch] [2026-06-23T19:53:57+00:00] /usr/bin/bash: line 2:   423 Killed                  ./bin/transform-gisaid data/gisaid.ndjson --accessions data/all_accessions.tsv.gz --output-metadata data/gisaid/metadata_transformed.tsv --output-fasta data/gisaid/sequences.fasta --output-additional-info data/gisaid/additional_info.tsv --output-unix-newline > data/gisaid/flagged-annotations

Job FAILED after 263.3 minutes (OutOfMemoryError: Container killed due to memory usage, exited 1)
```

https://github.com/nextstrain/ncov-ingest/actions/runs/28037143489/job/82993626925

Likely due to reductions in memory here: https://github.com/nextstrain/ncov-ingest/commit/b5cf23fab4499c4812972855b2c06c7ca66f9414

```
Reduce AWS Batch CPU and memory to fit c7a.8xlarge cap
Our AWS Batch Compute Environment now uses c7a instances capped at
c7a.8xlarge (32 vCPU, 64 GiB). Both ingest workflows previously
requested more CPUs than this limit allows.

- fetch-and-ingest-genbank: 36 CPUs → 32, 68 GiB → 60 GiB.
- fetch-and-ingest-gisaid: 48 CPUs → 32, 90 GiB → 60 GiB.

Both target c7a.8xlarge. The main parallel step is nextclade, which
uses workflow.cores * 0.5 threads per invocation (two nextclade rules
can run concurrently). With 32 CPUs, each nextclade run gets 16
threads, which should provide good throughput for the millions of
SARS-CoV-2 sequences processed.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
```

victorlin:
The next larger instance is c7a.12xlarge at 48 vCPUs and 96 GiB memory for  $2.46/hr.

We can enable that in the compute environment, bump the GISAID ingest invocation back to 90 GiB, and see if it works.

victorlin:
@corneliusroemer I've started a new run: https://github.com/nextstrain/ncov-ingest/actions/runs/28135995624

corneliusroemer:
Thanks a lot @victorlin! Seems to have worked! Seems to take around 15hr now even if it reruns Nextclade on everything due to dataset update.

[...]

PR #543, opened 2026-07-12
Scale GenBank and GISAID AWS Batch requests to 16 cpus / 28 GiB

trvrb:
Scale both master ingest builds down from their current AWS Batch sizing to **--cpus 16 --memory 28GiB** (c7a.4xlarge).

## Rationale — perf baseline
The two integration/perf-baseline runs (2026-07-11) showed large memory headroom (peak per-rule `max_rss`):

| Build | Was | Peak rule (RSS) | New |
|-------|-----|-----------------|-----|
| GenBank | 32 cpu / 60 GiB (c7a.8xlarge) | `transform_genbank_data` — 19.9 GB | 16 cpu / 28 GiB (c7a.4xlarge) |
| GISAID | 48 cpu / 90 GiB (c7a.12xlarge) | `transform_gisaid_data` — 14.6 GB | 16 cpu / 28 GiB (c7a.4xlarge) |

28 GiB still leaves >1.5× over the hungriest rule on each side. `--cpus` and `--memory` scale together, so both move as a pair; c7a.4xlarge is the next convenient step down (nothing sits between it and c7a.8xlarge).

## Changes
- Both workflows: `--cpus 16 --memory 28GiB`.
- GISAID: removed the explicit `--aws-batch-queue nextstrain-job-queue-c7a-12xlarge` pin so it uses the default queue and the instance is sized by the resource request.

### DIFF
--- a/.github/workflows/fetch-and-ingest-gisaid-master.yml
+++ b/.github/workflows/fetch-and-ingest-gisaid-master.yml
@@ -76,11 +76,10 @@ jobs:
       run: |
         nextstrain build \
           --aws-batch \
-          --aws-batch-queue nextstrain-job-queue-c7a-12xlarge \
           --detach \
           --no-download \
-          --cpus 48 \
-          --memory 90GiB \
+          --cpus 16 \
+          --memory 28GiB \
           --env GITHUB_RUN_ID \
           --env SLACK_TOKEN \
           --env SLACK_CHANNELS \
--- a/.github/workflows/fetch-and-ingest-genbank-master.yml
+++ b/.github/workflows/fetch-and-ingest-genbank-master.yml
@@ -94,8 +94,8 @@ jobs:
           --aws-batch \
           --detach \
           --no-download \
-          --cpus 32 \
-          --memory 60GiB \
+          --cpus 16 \
+          --memory 28GiB \
           --env GITHUB_RUN_ID \
           --env SLACK_TOKEN \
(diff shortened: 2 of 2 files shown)


.github/workflows/fetch-and-ingest-gisaid-master.yml
```diff
@@ -76,11 +76,10 @@ jobs:
       run: |
         nextstrain build \
           --aws-batch \
-          --aws-batch-queue nextstrain-job-queue-c7a-12xlarge \
```
victorlin, line 79: Re: PR description

> GISAID: removed the explicit `--aws-batch-queue nextstrain-job-queue-c7a-12xlarge` pin so it uses the default queue and the instance is sized by the resource request.

For clarity, `nextstrain-job-queue-c7a-12xlarge` doesn't force c7a-12xlarge; the instance is always sized by the resource request. Rather, it allows the larger instance that's not allowed by the default queue. I added this to address memory issues (#532).

I assume this PR is a follow-up to #537 which significantly reduced memory usage.

~Once this is merged, we can remove the job queue (revert https://github.com/nextstrain/infra/pull/74).~ It's useful to keep this available [when needed](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1784752580040159?thread_ts=1784752482.432639&cid=C01LCTT7JNN).

victorlin (approved):

